PR #7, opened 2019-05-08
Make tests run on Gitlab CI

vgrichina:


.gitlab-ci.yml
```diff
@@ -2,26 +2,30 @@ image: node:10
...
 
 cache:
   paths:
+    - nearcore/
     - node_modules/
 
 install_dependencies:
   stage: build
   script:
+    - rm -rf nearcore/
```
janedegtiareva: could you leave a comment why this is needed? It's definitely not intuitive.

vgrichina: probably makes more sense to remove `nearcore` from cache, then this won't be needed

vgrichina: added comment

PR #9, opened 2019-05-13
Wallet integration: add a utility function to reconstruct key from ur…

janedegtiareva:
…lparams


janedegtiareva:
tests pass locally

janedegtiareva:
we don't need this based on latest discussion

PR #11, opened 2019-05-15
added more crypto functions

vans163:
Lets get this started https://github.com/nearprotocol/near-wallet/issues/32.

I dont like how the base58 is used atm, but not sure what the protocol would be. Return uint8array, return b58 or return b64?

Up for first review, this is not final.


potatodepaulo:
Hey wanted to clear this up to clear out older issues. This has been addressed in different pushes.

PR #13, opened 2019-05-16
Access key integration initial take (a few known issues exist)

janedegtiareva:
known issue: first rpcs after coming back to app from wallet won't work.
page needs to be reloaded


vgrichina:
@janedegtiareva is nearlib still going to work with old wallet after these changes? If not I suggest we create separate branch for access keys release.

PR #15, opened 2019-05-16
Changes to wallet-account flow that completes authorization to work with access keys

janedegtiareva:


janedegtiareva:
This will require a small change to templates (main.js wallet integration code)

PR #23, opened 2019-05-22
Revert access keys changes temporarily to allow a nearlib version

janedegtiareva:
which fixeses a few nearmint side effects but does not require new wallet
yet


vgrichina:
Turned out access keys already work well enough after fixing infinite redirect bug http://wallet.nearprotocol.com

PR #24, opened 2019-05-23
Expose method to send tokens

vgrichina:
Extract common transaction encoding/signing/submission code into separate method

### DIFF
--- a/near.js
+++ b/near.js
@@ -97,32 +97,19 @@ class Near {
             functionCall.amount = amount;
         }
 
-        return this.submitTransaction(originator, functionCall);
+        return this.nearClient.signAndSubmitTransaction(originator, functionCall);
     }
 
-    async submitTransaction(originator, transaction) {
-        const protoClass = transaction.constructor;
-        const className = protoClass.name;
-        const propertyName = className[0].toLowerCase() + className.replace(/Transaction$/, '').substring(1);
-        console.log('submitTransaction', className, propertyName, originator);
-
-        const buffer = protoClass.encode(transaction).finish();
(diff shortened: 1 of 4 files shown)


nearclient.js
```diff
@@ -52,6 +52,7 @@ class NearClient {
     async getTransactionStatus(transactionHash) {
         const encodedHash = _arrayBufferToBase64(transactionHash);
         const response = await this.jsonRpcRequest('tx', [encodedHash, false]);
+        //console.log('response', response);
```
bowenwang1996: Should we remove this?

nearclient.js
```diff
@@ -74,6 +75,7 @@ class NearClient {
             id: Date.now().toString(),
         };
         const response = await this.nearConnection.request('', request);
+        console.log('response', response);
```
bowenwang1996: Also this

near.js
```diff
@@ -97,18 +97,37 @@ class Near {
...
+
+        const buffer = protoClass.encode(transaction).finish();
         const signatureAndPublicKey = await this.nearClient.signer.signBuffer(
             buffer,
             originator,
         );
 
         const signedTransaction = SignedTransaction.create({
-            functionCall,
+            [propertyName]: transaction,
```
bowenwang1996: What is this syntax `[propertyName]` as a field name?

vgrichina: it uses value of `propertyName` as a field name

nearclient.js
janedegtiareva: I would actually like to have some sort of debug flag so that you can more easily have debug output from nearlib. (Can be a different PR)

PR #28, opened 2019-05-29
update dev account funding

potatodepaulo:
Just adding dev account funding with default param set.

### DIFF
--- a/dev.js
+++ b/dev.js
@@ -89,10 +89,10 @@ module.exports = {
     /**
      * Function to create an account on local node. This will not work on non-dev environments.
      */
-    createAccountWithLocalNodeConnection: async function (newAccountName, newAccountPublicKey) {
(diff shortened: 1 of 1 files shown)


dev.js
```diff
@@ -89,10 +89,10 @@ module.exports = {
     /**
      * Function to create an account on local node. This will not work on non-dev environments.
      */
-    createAccountWithLocalNodeConnection: async function (newAccountName, newAccountPublicKey) {
+    createAccountWithLocalNodeConnection: async function (newAccountName, newAccountPublicKey, amount = 1000000000000) {
```
vgrichina: If we add additional parameter here it also needs to be introduced for `createAccountWithContractHelper` which would require plumbing in contract helper. Probably easier to use constant, one always can use lower level API for more control.

potatodepaulo: done

PR #29, opened 2019-05-30
RPC to view account details (to show authorized apps)

janedegtiareva:


dist/nearlib.js
```diff
@@ -153,6 +153,26 @@ class Account {
...
+     */
+    async getAccountDetails(accountId) {
+        const response = await this.nearClient.jsonRpcRequest('abci_query', [`access_key/${accountId}`, '', '0', false]);
+        const decodedResponse = this.nearClient.decodeResponseValue(response.response.value);
+        const result = {
+            authorizedApps : [],
+            transactions: []
+        };
+        Object.keys(decodedResponse).forEach(function(key) {
+            result.authorizedApps.push({
```
evgenykuzyakov: We probably want to store public key for the access key here as well. In case user decides to deauthorize an app.

janedegtiareva: Agree, I was going to do this when I work on the deauthorize flow so that it's more clear what the format should be.

PR #33, opened 2019-05-31
new RPC to remove access key 

janedegtiareva:


test/integration.test.js
```diff
@@ -245,8 +245,44 @@ describe('with access key', function () {
...
+        const contract = await nearjs.loadContract(contractId, {
+            viewMethods: ['getValue'],
+            changeMethods: ['setValue'],
+            sender: newAccountId,
+        });
+        await account.removeAccessKey(
+            newAccountId,
+            (await keyStore.getKey(newAccountId)).getPublicKey()
+        );
+        // Replacing public key for the account with the access key
```
vgrichina: > test that load contract works and we can make calls afterwards
and
> Replacing public key for the account with the access key

should be separate tests with common setup

test/integration.test.js
```diff
@@ -245,8 +245,44 @@ describe('with access key', function () {
...
+        await keyStore.setKey(newAccountId, keyForAccessKey);
+        try {
+            const setCallValue2 = await generateUniqueString('setCallPrefix');
+            await contract.setValue({ value: setCallValue2 });
+            fail('Transaction signed by the access key should not have been accepted');
+        } catch (_e) {
```
vgrichina: Jest has specific way for asserting rejected promise:

`await expect(contract.triggerAssert()).rejects.toThrow(/Transaction .+ failed.+expected to fail/);`

test/integration.test.js
janedegtiareva: this is not actually testing that we can make calls. Copy paste artifact. I updated the comment.

test/integration.test.js
janedegtiareva: done

PR #35, opened 2019-05-31
Add public key to getaccountdetails rpc so that it can be used to rem…

janedegtiareva:
…ove key

### DIFF
--- a/test/integration.test.js
+++ b/test/integration.test.js
@@ -178,6 +178,7 @@ describe('with access key', function () {
     let contractId = 'test_contract_' + Date.now();
     let newAccountId;
     let newAccountKeyPair;
+    let keyForAccessKey = KeyPair.fromRandomSeed();
 
     jasmine.DEFAULT_TIMEOUT_INTERVAL = 200000;
 
@@ -212,6 +213,16 @@ describe('with access key', function () {
             aliceAccountName);
         await nearjs.waitForTransactionResult(createAccountResponse);
         await keyStore.setKey(newAccountId, newAccountKeyPair);
+
+        const addAccessKeyResponse = await account.addAccessKey(
+            newAccountId,
+            keyForAccessKey.getPublicKey(),
+            contractId,
+            '',  // methodName
+            '',  // fundingOwner
+            1000000000,  // fundingAmount
+        );
+        await nearjs.waitForTransactionResult(addAccessKeyResponse);
     });
 
     afterEach(async () => {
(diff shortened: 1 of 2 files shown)


test/integration.test.js
```diff
@@ -258,19 +258,29 @@ describe('with access key', function () {
             4000000000,  // fundingAmount
         );
         await nearjs.waitForTransactionResult(addAccessKeyResponse);
+
```
vgrichina: let's avoid code duplication in tests. E.g.:

```
        // Adding access key
        const keyForAccessKey = KeyPair.fromRandomSeed();
        const addAccessKeyResponse = await account.addAccessKey(
            newAccountId,
            keyForAccessKey.getPublicKey(),
            contractId,
            '',  // methodName
            '',  // fundingOwner
	            4000000000,  // fundingAmount
	        );
```

this can easily be in `beforeEach` for whole set of tests. This will also help differentiate what is setup for a test case and what is actually tested.

```
        // test that we can't make calls with deleted key
	        const contract = await nearjs.loadContract(contractId, {
	            viewMethods: ['getValue'],
	            changeMethods: ['setValue'],
	            sender: newAccountId,
	        });
```

this can be a helper function

janedegtiareva: this is a function call. I don't want to make a helper function for a single function call.

vgrichina: first snippet is 10 line of code for repeated over at least 3 tests, can be replaced by zero additional lines for each test. There is hardcoded stuff like funding amount which very likely needs to be updated for all tests at once vs individually. 

and second snippet is  5 lines of code used also about 3 times. Stuff like `viewMethods: ['getValue'],
	            changeMethods: ['setValue'],` also is boilerplate which likely will need to get replaced for all places at once and not individually, as it's specific to .wasm file deployed (and we don't have different files for different testS).

test/integration.test.js
```diff
@@ -212,24 +213,23 @@ describe('with access key', function () {
...
-    });
-
-    afterEach(async () => {
-        console.log = oldLog;
-    });
 
-    test('make function calls using access key', async () => {
-        // Adding access key
-        const keyForAccessKey = KeyPair.fromRandomSeed();
         const addAccessKeyResponse = await account.addAccessKey(
```
vgrichina: Does this need to happen for literally every test in the file? 

Generally similar tests should be grouped using `describe`, then you can have common init code for tests in the group isolated from other tests. This can be nested multiple times. That's whole idea of RSpec-like test frameworks.

Like you nest stuff something like this:
- with node connection
    - with deployed contract
         - with access key
              - make function calls
              - view account details
              - add second key
              - remove access key

See e.g. https://github.com/mochajs/mocha/blob/master/test/unit/runnable.spec.js
Or array example here https://rspec.info/documentation/3.8/rspec-core/#A_Word_on_Scope

janedegtiareva: I prefer the way they are grouped now, because I am not testing local node connection but access keys. The top level thing should describe the broad category you are testing, not setup steps. I think we should not block PRs over a stylistic disagreements over how to organize things right before release IMO.

vgrichina: right now access keys setup happens for literally every test in the file. Is this what you intended? This is not just stylistic, it's also test performance issue.

vgrichina: never mind, I misread the diff and thought it got move to top-level, keeping within `describe('with access key'` is fine

PR #38, opened 2019-06-03
Add/remove account keys (not-specific to contract)

vgrichina:


account.js
```diff
@@ -112,11 +112,43 @@ class Account {
...
+        publicKey = bs58.decode(publicKey);
+        return this.nearClient.signAndSubmitTransaction(ownersAccountId, AddKeyTransaction.create({
+            nonce,
+            originator: ownersAccountId,
+            newKey: publicKey,
+        }));
+    }
+
+    /**
+     * Removes given unrestricted key from the owners account.
```
evgenykuzyakov: This will also remove an accessKey with the given public key. `removeAccessKey` does the same.

account.js
```diff
@@ -112,11 +112,43 @@ class Account {
...
+     * @param {string} ownersAccountId id of the owner's account.
+     * @param {string} publicKey public key encoded in bs58
+     */
+    async removeAccountKey(ownersAccountId, publicKey) {
+        const nonce = await this.nearClient.getNonce(ownersAccountId);
+        publicKey = bs58.decode(publicKey);
+        return this.nearClient.signAndSubmitTransaction(ownersAccountId, DeleteKeyTransaction.create({
+            nonce,
+            originator: ownersAccountId,
+            curKey: publicKey,
```
evgenykuzyakov: Either `cur_key` in `removeAccessKey` is incorrect or `curKey` here is incorrect. Since you have a test, it probably the case that `removeAccessKey` is not correctly working.

account.js
vgrichina: See here https://github.com/nearprotocol/nearlib/pull/37

account.js
vgrichina: See here https://github.com/nearprotocol/nearlib/pull/37

account.js
```diff
@@ -49,7 +49,7 @@ class Account {
      * @param {string} ownersAccountId id of the owner's account.
      * @param {string} newPublicKey public key for the access key.
      * @param {string} contractId if the given contractId is not empty, then this access key will only be able to call
-     *      the given contractId.
+     *      the given contractId. Otherwise his method is going to add unrestricted key.
```
evgenykuzyakov: typo

Issue #39, opened 2019-06-04
Refactor Nearlib


ilblackdragon:
There are few things we want to improve:

Configuration search:
 - [ ] For Node.js use cases, search for `.near/` folder with config and keys. Main uses cases:
   - local to project (so go up to folder with `node_modules` / `package.json` and check for `.near`). This can be official TestNet and custom network.
   - running node connected to official TestNet (`~/.near/` for configs). Want to use it, but need to use contract helper for account creation.
   - running custom network with local node (`~/.near/` for configs). Want to use local key pair for account creation.
 - [ ] Create various types of transactions functions that support creating and signing transactions (without having connection).
 - [ ] `ConnectProvider` interface that various providers like RPC, RPCv2, Network, WebSockets can implement (e.g. current NearClient is RPC provider). Provides `sendTransaction(txBytes, [timeout]) -> Promise<FinalTransactionResult>` and `query(path, data) -> Promise<QueryResult>` interfaces.
 - [ ] `AccountCreatorProvider` - interface, that either uses helper or given keys to create accounts.
 - [ ] `connect` that is configured / or searches for configuration and uses given `ConnectProvider`, `AccountCreatorProvider` and `KeyStore`.
 - [ ] `Account` abstraction, that provides accessors to account state (balance, etc), access keys (with permissions per key), etc. Allows to initialize `Contract` with it to send Tx from it.
 - [ ] Wallet API is defined via 2 URLs: "Authorize App" API and "Sign Tx" API.
 - [ ] `Contract` handles itself sending to Wallet if currently signed Tx is not allowed by access key or access key is missing.
 - [ ] `KeyStore` should have some form of `Wallet` to provide what are current accounts available for given network, and allow to change accounts and networks. Also this should use Wallet API to "login" to retrieve current account id if unavailable.

vans163:
setItem for localstorage tokens/secretket is kinda inbetween nearwallet and nearlib.  The secretket is set in nearlib and the account id and tokens are set in nearwallet. Should these be all set in nearlib, or shipping to wallet?

vgrichina:
@vans163 it's perfectly valid use case not not use wallet at all. From our POV wallet is basically a special type of `Signer`.

vans163:
Yea I am thinking that thin API in wallet repo should be moved into nearlib

DanielRX:
[...]

PR #41, opened 2019-06-05
Nightshade Typescript refactoring

### LINKED ISSUE #5: NEARLIB: no explicit error if trying to sign transaction, but no key is found.

### LINKED ISSUE #12: bs58 should be smarter
```
    /**
     * Encode a buffer as string using bs58
     * @param {Buffer} buffer 
     * @example
     * KeyPair.encodeBufferInBs58(key.publicKey)
     */
    static encodeBufferInBs58(buffer) {
        const bytes = Buffer.from(buffer);
        const encodedValue = bs58.encode(bytes);
        return encodedValue;
    }
```
Stuff like this should be inside the bs58.encode() function instead of outside, bs58 should auto determine the type is Buffer or Uint8Array.  Maybe use a forked bs58 with this update?

Or get a base58 lib that supports uint8array buffers.

### LINKED ISSUE #18: Better error reporting
Hard to interpret error messages like:
> Error: [-32603] Internal error: Tx (00) not found

### LINKED ISSUE #27: Clean-up 'make function calls' test in integration.test.js
@vgrichina commented on [Fri Feb 15 2019](https://github.com/nearprotocol/nearcore/issues/579)

Test is currently whole wall of code with few unrelated tests jammed together:

```js
    test('make function calls', async () => {
        const args = {
            'name': 'trex'
        };
        const viewFunctionResult = await nearjs.callViewFunction(
            contractName,
            'hello', // this is the function defined in hello.wasm file that we are calling
            args);
        expect(viewFunctionResult).toEqual('hello trex');

        var setCallValue = await generateUniqueString('setCallPrefix');
        const setArgs = {
            'value': setCallValue
        };
        const scheduleResult = await nearjs.scheduleFunctionCall(
            0,
            aliceAccountName,
            contractName,
            'setValue', // this is the function defined in hello.wasm file that we are calling
            setArgs);
        expect(scheduleResult.hash).not.toBeFalsy();
        await nearjs.waitForTransactionResult(scheduleResult);
        const getValueResult = await nearjs.callViewFunction(
            contractName,
            'getValue', // this is the function defined in hello.wasm file that we are calling
            {});
        expect(getValueResult).toEqual(setCallValue);

        // test that load contract works and we can make calls afterwards
        const options = {
            viewMethods: ['hello', 'getValue'],
            changeMethods: ['setValue'],
            sender: aliceAccountName
        };
        const contract = await nearjs.loadContract(contractName, options);
        const helloResult = await contract.hello(args);
        expect(helloResult).toEqual('hello trex');
        var setCallValue2 = await generateUniqueString('setCallPrefix');
        const setArgs2 = {
            'value': setCallValue2
        };
        const setValueResult = await contract.setValue(setArgs2);
        expect(setValueResult.status).toBe('Completed');
        const getValueResult2 = await contract.getValue();
        expect(getValueResult2).toEqual(setCallValue2);
    });
```

compare to other tests, e.g.:

```js
    test('can get logs from method result', async () => {
        await contract.generateLogs();
        expect(logs).toEqual([`[${contractName}]: LOG: log1`, `[${contractName}]: LOG: log2`]);
    });
```

---

@ilblackdragon commented on [Tue Mar 26 2019](https://github.com/nearprotocol/nearcore/issues/579#issuecomment-476840403)

@janedegtiareva any update here?

ilblackdragon:
This follows ideas for #39 + TypeScript for better typed APIs.

Fixes #5 
Fixes #12 
Ref #13 
Fixes #18 
Fixes #27


test/account.test.js
```diff
@@ -28,12 +34,29 @@ test('create account and then view account returns the created account', async (
...
+    expect(state.amount).toEqual(INITIAL_BALANCE + BigInt(10000));
+});
+
 // Generate some unique string with a given prefix using the alice nonce.
 const generateUniqueString = async (prefix) => {
     let status = await masterAccount.state();
     return prefix + status.nonce;
 };
+
+async function testCreateAccount() {
```
vgrichina: 1) `s/testCreateAccount/createAccount/`
2) I think there should be ability to override default values (like amount) to make it more flexible, i.e. similar to this https://github.com/nearprotocol/nearlib/blob/master/test/integration.test.js#L175

src.ts/transaction.ts
```diff
@@ -1,3 +1,6 @@
+'use strict';
+
+import { BN } from "bn.js";
```
vgrichina: I think we shouldn't use `bn.js`, but use https://github.com/GoogleChromeLabs/jsbi which is forward-compatible with native JS BigInt.

test/account.test.js
vgrichina: Might also be worth looking into some of the test object factory modules:
https://www.npmtrends.com/factory-girl-vs-factory-js-vs-fixture-factory-vs-rosie

src.ts/transaction.ts
ilblackdragon: added comment explaining why we use it.

test/account.test.js
ilblackdragon: Hm, not sure about the factory stuff - if it's test fixture stuff can do separately.

PR #44, opened 2019-07-10
Exposed `status` and `block` RPC

frol:


frol:
@ilblackdragon @vgrichina please, review.

frol:
1. Sure
2. Yes, the snake case style is from the Nearcore side, so it cannot be changed without manual renaming.

frol:
@vgrichina I have added the tests and rebased (there has been a merge conflict). Please, review.

test/providers.test.js
```diff
@@ -1,6 +1,20 @@
 
 const nearlib = require('../lib/index');
 
+test('json rpc fetch node status', async () => {
+    const config = Object.assign(require('./config')(process.env.NODE_ENV || 'test'));
```
vgrichina: looks like there is common setup (`config`, `provider`) that you want to move to `beforeEach`

PR #47, opened 2019-07-22
 Add parameter to allow passing amount to `functionCall`

vgrichina:
Required by https://github.com/nearprotocol/near-shell/pull/72


src.ts/account.ts
```diff
@@ -143,13 +143,13 @@ export class Account {
         return this.signAndSendTransaction(deployContract(this._state.nonce, this.accountId, data));
     }
 
-    async functionCall(contractId: string, methodName: string, args: any): Promise<FinalTransactionResult> {
```
janedegtiareva: do you mind adding a comment explaining what amount does? (might as well document the other ones too, although they are more self-explanatory)

vgrichina: what would you write about amount?

PR #50, opened 2019-07-29
[Shell integration] add a separate param for shell url suffix

janedegtiareva:
The purpose is to move away from the harcoded url suffix in code


vgrichina:
Why is this change needed? If it's for the context of authenticating through lightweight wallet in nearcore – why not just support `/login` URL there?

Ideally I would like to avoid configuring 2 different string values to point to a given wallet.

janedegtiareva:
> Why is this change needed? If it's for the context of authenticating through lightweight wallet in nearcore – why not just support `/login` URL there?
> 
> Ideally I would like to avoid configuring 2 different string values to point to a given wallet.
Couple of things here: 1) not sure it's possible to support login/?title=Who+was from a simple html page. 
2) I don't like having harcoded piece of path in the code. 

We can have the full path in config, but that would make a backwards incompatible change (unless we do some sort of "walletUrlV2" and stop using the old one)

vgrichina:
> Couple of things here: 1) not sure it's possible to support login/?title=Who+was from a simple html page.

not sure what do you mean. How is this different from existing wallet which is only using static pages?

> 2) I don't like having harcoded piece of path in the code.

I don't think this justifies having to update config in every example. Also this isn't consistent with how other stuff works, e.g. we don't let user to pass a path for every nearcore API endpoint.

janedegtiareva:
> > Couple of things here: 1) not sure it's possible to support login/?title=Who+was from a simple html page.
> 
> not sure what do you mean. How is this different from existing wallet which is only using static pages?
> 
> > 1. I don't like having harcoded piece of path in the code.
> 
> I don't think this justifies having to update config in every example. Also this isn't consistent with how other stuff works, e.g. we don't let user to specify a path for every nearcore API endpoint.



#1: OK detailed problem: react is doing parsing urls and routing, which lets you have a query string after a /  ->>>> " login/?title=Who+was"
We don't want a fancy react setup for the fake wallet, and as far as I know there is no way to do this with plain old html + javascript. This definitely by itself justifies #2.

janedegtiareva:
> > > Couple of things here: 1) not sure it's possible to support login/?title=Who+was from a simple html page.
> > 
> > 
> > not sure what do you mean. How is this different from existing wallet which is only using static pages?
> > > 1. I don't like having harcoded piece of path in the code.
> > 
> > 
> > I don't think this justifies having to update config in every example. Also this isn't consistent with how other stuff works, e.g. we don't let user to specify a path for every nearcore API endpoint.
> 
> #1: OK detailed problem: react is doing parsing urls and routing, which lets you have a query string after a / ->>>> " login/?title=Who+was"
> We don't want a fancy react setup for the fake wallet, and as far as I know there is no way to do this with plain old html + javascript. This definitely by itself justifies #2.

Another thing to consider is that we have a chance to fix a suboptimal config decision before mainnet, so it's better to do this now.

janedegtiareva:
Do we need a NEP for this since it turned out to be a non-trivial discussion?

ilblackdragon:
Here are my 2cents:
```
#1: OK detailed problem: react is doing parsing urls and routing, which lets you have a query string after a / ->>>> " login/?title=Who+was"
We don't want a fancy react setup for the fake wallet, and as far as I know there is no way to do this with plain old html + javascript.
```

Generally, React is not required for urls & routing in this case -> you can just create login.html page and parse arguments from `(new URL(document.location)).searchParams.get("title")`.

But generally, I would want us to have the "Wallet URL API" which defines the end points and parameters for Wallets, that other wallets can implement and it is swappable from our wallet to any other wallet (including for mobile just something like url near://wallet/)

My expectation it would be either:
   - login/?title=...
   - sign/?tx=<base64 of tx>
   - authorize?account_id=...&public_key=...

_Note, title here is not really correct because frontend can pass anything, title should be loaded from the contract_ 
  
or what probably is better to not require many different end points, as it is really just a signing transaction and single UI:
  -  `/?type=login&title=abc` 
  - `/?type=sign&transaction=...` 
  - `/?type=authorize?account_id=...`

I think this ending up as NEP would be good, if nothing else but for documenting purposes for other wallets to use (e.g. formats of end points and parameters).

janedegtiareva:
> Here are my 2cents:
> 
> ```
> #1: OK detailed problem: react is doing parsing urls and routing, which lets you have a query string after a / ->>>> " login/?title=Who+was"
> We don't want a fancy react setup for the fake wallet, and as far as I know there is no way to do this with plain old html + javascript.
> ```
> 
> Generally, React is not required for urls & routing in this case -> you can just create login.html page and parse arguments from `(new URL(document.location)).searchParams.get("title")`.
> 
> But generally, I would want us to have the "Wallet URL API" which defines the end points and parameters for Wallets, that other wallets can implement and it is swappable from our wallet to any other wallet (including for mobile just something like url near://wallet/)
> 
> My expectation it would be either:
> 
> * login/?title=...
> * sign/?tx=
> * authorize?account_id=...&public_key=...
> 
> _Note, title here is not really correct because frontend can pass anything, title should be loaded from the contract_
> 
> or what probably is better to not require many different end points, as it is really just a signing transaction and single UI:
> 
> * `/?type=login&title=abc`
> * `/?type=sign&transaction=...`
> * `/?type=authorize?account_id=...`
> 
> I think this ending up as NEP would be good, if nothing else but for documenting purposes for other wallets to use (e.g. formats of end points and parameters).

It's a known issue re title url param -- once we have this in metadata, we can stop passing it as url params. It's there now as a stop gap to have a usable wallet before we have metadata. (I just filed an actual github issue for this so that we don't forget to remove it later https://github.com/nearprotocol/near-wallet/issues/118) 

I don't think we should change the API to ?type=login style at this point since it requires changes to the wallet. 
So we basically just need to decide if we feel like we need to enforce the /login/ part of {configurableloginurl}/login/?public_key=...  as part of API. 
PROs: ensure readable/standard urls.
CONS: More constraints (which are not strictly necessary) on 3rd party wallet devs. Eg. they have an existing url structure and want to be consistent with it.

janedegtiareva:
I am going to start a NEP for wallet API. We can revisit implementation after NEP gets a chance to be reviewed.

PR #51, opened 2019-08-10
add result.body to responses from called functions

potatodepaulo:
All other stuff is generated. Just look at account.ts and provider.ts. This works; tested this with local project. TODO: Tests seem to be generally broken across the board in nearlib.


vgrichina:
I think it's best if we just unwrap result in contract call sugar methods:
https://github.com/nearprotocol/nearlib/pull/54

These changes aren't really conflicting though.

dist/nearlib.js
```diff
@@ -16,6 +16,7 @@ const bn_js_1 = __importDefault(require("bn.js"));
 const transaction_1 = require("./transaction");
 const provider_1 = require("./providers/provider");
 const serialize_1 = require("./utils/serialize");
+// import { restElement } from '@babel/types';
```
janedegtiareva: nit: clean up commented out code please :)

potatodepaulo:
That works for me. Closing for #54

PR #56, opened 2019-08-15
Changes to serialization + NEPs 5 & 8

ilblackdragon:
Also changes required for
 - NEP5: https://github.com/nearprotocol/NEPs/pull/5
 - NEP8: https://github.com/nearprotocol/NEPs/pull/8


src.ts/transaction.ts
```diff
@@ -3,91 +3,199 @@
...
-    [CreateAccountTransaction, 'createAccount'],
-    [DeployContractTransaction, 'deployContract'],
-    [FunctionCallTransaction, 'functionCall'],
-    [SendMoneyTransaction, 'sendMoney'],
-    [StakeTransaction, 'stake'],
-    [SwapKeyTransaction, 'swapKey'],
-    [AddKeyTransaction, 'addKey'],
-    [DeleteKeyTransaction, 'deleteKey'],
-]);
+class Enum {
```
janedegtiareva: it's strange to have Enum under transaction. is there some other name that would work, or at least a comment on what this means?

src.ts/utils/serialize.ts
```diff
@@ -12,3 +13,216 @@ export function base_encode(value: Uint8Array | string): string {
 export function base_decode(value: string): Uint8Array {
     return bs58.decode(value);
 }
+
+const INITIAL_LENGTH = 1024;
```
janedegtiareva: It would be helpful to have a comment here what this means.

test/serialize.test.js
```diff
@@ -0,0 +1,41 @@
+
```
janedegtiareva: Are there deserialize tests too?

janedegtiareva:
The pipeline is failing due to lint errors (most likely easy fix by npm run fix) and also tests are failing. test/serialize.test.js is actually failing -- is that because we don't have a node with the nearcore changes yet?

src.ts/transaction.ts
```diff
@@ -3,91 +3,199 @@
...
 
-export function signedTransaction(transaction: AllTransactions, signature: Signature): SignedTransaction {
-    const fieldName = TRANSACTION_FIELD_MAP.get(transaction.constructor);
-    return new SignedTransaction({
-        signature: signature.signature,
-        publicKey: google.protobuf.BytesValue.create({ value: base_decode(signature.publicKey) }),
-        [fieldName]: transaction,
-    });
+const SCHEMA = {
+    'Signature': {kind: 'struct', fields: [['keyType', 'u8'], ['data', [32]]]},
```
janedegtiareva: It would be very useful to document this.

PR #61, opened 2019-09-01
Switch to use external Borsh JS library

ilblackdragon:


janedegtiareva:
Should we close this PR since it's really stale?

PR #64, opened 2019-09-04
Verify that access keys with no contract name get added with full access

janedegtiareva:


test/account.access_key.test.js
```diff
@@ -69,10 +69,17 @@ test('view account details after adding access keys', async() => {
...
+    // wallet calls this with an empty string for contract id and method
+    await workingAccount.addKey(keyPair.getPublicKey(), '', '');
 
     let accessKeys = await workingAccount.getAccessKeys();
 
     expect(accessKeys.length).toBe(2);
     expect(accessKeys.map((item) => item.public_key).includes(keyPair.getPublicKey().toString())).toBe(true);
+
+    for (let i = 0; i < accessKeys.length; i++) {
+        if (accessKeys[i].public_key == keyPair.getPublicKey()) {
```
vgrichina: @janedegtiareva I think this also should need `.toString()` for public key. 

also I think it's better to combine this with above check by doing something like:

```
const addedKey = accessKeys.find((item) => item.public_key.includes(keyPair.getPublicKey().toString()));
```

then you can assert both that key is non-null and that it has `FullAccess`

vgrichina: (this avoids the case where test would still pass if `getPublicKey` doesn't return string (just because there won't be any `expect` call in that case)

test/account.access_key.test.js
```diff
@@ -69,10 +69,17 @@ test('view account details after adding access keys', async() => {
 
 test('loading account after adding a full key', async() => {
     const keyPair = nearlib.utils.KeyPair.fromRandom('ed25519');
-    await workingAccount.addKey(keyPair.getPublicKey());
+    // wallet calls this with an empty string for contract id and method
+    await workingAccount.addKey(keyPair.getPublicKey(), '', '');
```
bowenwang1996: Why do we do this when we've already made contractId optional?

janedegtiareva: This is how wallet calls the API. It's easy to make that mistake when dealing with UX components that may do a '' instead of null, so we should make sure it works on nearlib level

PR #67, opened 2019-09-05
Better error handling for missing keys and better support for key files

ilblackdragon:
Support `secret_key` and `private_key` field in the key files because we use different format in nearcore and nearlib.

### DIFF
--- a/src.ts/account.ts
+++ b/src.ts/account.ts
@@ -51,19 +51,16 @@ export class Account {
     }
 
     async fetchState(): Promise<void> {
+        this._accessKey = null;
         this._state = await this.connection.provider.query(`account/${this.accountId}`, '');
-        try {
-            const publicKey = await this.connection.signer.getPublicKey(this.accountId, this.connection.networkId);
-            if (publicKey === null) {
-                console.warn(`Missing public key for ${this.accountId} in ${this.connection.networkId}`);
-            }
-            this._accessKey = await this.connection.provider.query(`access_key/${this.accountId}/${publicKey.toString()}`, '');
-            if (this._accessKey === null) {
-                throw new Error(`Failed to fetch access key for '${this.accountId}' with public key ${publicKey.toString()}`);
-            }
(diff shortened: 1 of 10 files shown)


vgrichina:
@ilblackdragon LGTM, but doesn't pass tests https://gitlab.com/near-protocol/nearlib/-/jobs/288922291

src.ts/account.ts
```diff
@@ -51,15 +51,16 @@ export class Account {
...
-        try {
-            const publicKey = (await this.connection.signer.getPublicKey(this.accountId, this.connection.networkId)).toString();
-            this._accessKey = await this.connection.provider.query(`access_key/${this.accountId}/${publicKey}`, '');
-            if (this._accessKey === null) {
-                throw new Error(`Failed to fetch access key for '${this.accountId}' with public key ${publicKey}`);
-            }
-        } catch {
-            this._accessKey = null;
+        const publicKey = await this.connection.signer.getPublicKey(this.accountId, this.connection.networkId);
+        if (publicKey === null) {
```
vgrichina: @ilblackdragon I'd prefer we use `!publicKey` here and in other cases when checking for absence of value (which cannot be a zero number). Otherwise even `undefined` won't pass this check.

src.ts/key_stores/unencrypted_file_system_keystore.ts
```diff
@@ -43,6 +43,16 @@ async function ensureDir(path: string): Promise<void> {
     }
 }
 
+export async function readKeyFile(path: string): Promise<[string, KeyPair]> {
+    const accountInfo = await loadJsonFile(path);
+    // The private key might be in private_key or secret_key field.
+    let privateKey = accountInfo.private_key;
+    if (privateKey === undefined && accountInfo.secret_key !== undefined) {
```
vgrichina: I'd just use `!` (as above instead of `=== null`). You don't really care whether it is empty string, `undefined` or `null`.

PR #68, opened 2019-09-10
Fix for crashes in view function when return value is null

janedegtiareva:


src.ts/account.ts
```diff
@@ -178,7 +178,8 @@ export class Account {
         if (result.logs) {
             this.printLogs(contractId, result.logs);
         }
-        return JSON.parse(Buffer.from(result.result).toString());
+        return result.result && result.result.length > 0 ?
+            JSON.parse(Buffer.from(result.result).toString()) :  null;
```
vgrichina: nit: there's extra space before `null` also idiomatic way would be:

```
return result.result && result.result.length > 0 && JSON.parse(Buffer.from(result.result).toString());
```

PR #71, opened 2019-09-16
Add nearlib implementation and test for deleteAccount

janedegtiareva:


bowenwang1996:
Looks like tests fail

PR #72, opened 2019-09-17
Add signing methods

DanielRX:
This PR adds methods to enable signing transactions, while preserving the existing interfaces, this would enable us to have users signed function calls.


DanielRX:
As we were unable to run `npm test`, we can't verify the changes yet. Can we have a solution for running the tests?

ilblackdragon:
@DanielRX what do you mean you can't run `npm test`? 
If you have a nearcore node running via `./scripts/start_unittest.py`, then running tests should be running.

vgrichina:
@DanielRX looks like this should be mostly covered by https://github.com/nearprotocol/nearlib/pull/184, feel free to re-open if not the case.

PR #75, opened 2019-09-25
Stop loading validator key on nearlib connection since it's not neede…

janedegtiareva:
…d for stake


bowenwang1996:
I think @vgrichina said something about this key being used as the master key if you run local nodes. Is that true?

vgrichina:
@bowenwang1996 that's really question for blockchain team. I have no idea what is your folks plan on this.

janedegtiareva:
> @bowenwang1996 that's really question for blockchain team. I have no idea what is your folks plan on this.
The original bug that I was trying to address with this was fixed yesterday by Vlad's change, and this is actually still needed for running localnet.

PR #78, opened 2019-09-25
Switching to new FinalExecutionOutcome. NEP#17

evgenykuzyakov:
Might need to change how I handle mixed enums. 

PR in nearcore: https://github.com/nearprotocol/nearcore/pull/1350
NEP#17: https://github.com/nearprotocol/NEPs/pull/17


test/account.test.js
```diff
@@ -48,7 +48,7 @@ test('delete account', async() => {
         await reloaded.state();
         fail('Getting state of deleted account was sucessful');
     } catch (_) {
-        // Expected to throw
+        // Expected to throwroviders.test.js
```
vgrichina: looks accidental

.gitlab-ci.yml
```diff
@@ -12,7 +12,7 @@ test:
   tags:
     - gcloud
   script:
-    - git clone --depth 1 https://github.com/nearprotocol/nearcore.git nearcore
+    - git clone --depth 1 --branch staging https://github.com/nearprotocol/nearcore.git nearcore
```
vgrichina: I think it's good idea to do both, similar to how @janedegtiareva did for `near-shell`:
https://github.com/nearprotocol/near-shell/pull/138

test/account.test.js
evgenykuzyakov: thanks

.gitlab-ci.yml
evgenykuzyakov: I think master is not going to work, since this staging change is breaking the results.

PR #79, opened 2019-09-26
docs dev deps upgraded

potatodepaulo:
Upgrading the docs dependencies to work with "hideBreadcrumbs" so it's not blocking or breaking stuff.


evgenykuzyakov:
Seems @vgrichina already pushed it

PR #83, opened 2019-10-04
Keep-alive (reuse) connections running in NodeJS environment to reduce the number of new connections

frol:
Ref: https://github.com/nearprotocol/nearcore/issues/1316

### DIFF
--- a/src.ts/utils/web.ts
+++ b/src.ts/utils/web.ts
@@ -11,7 +11,29 @@ export interface ConnectionInfo {
     headers?: { [key: string]: string | number };
 }
 
-const fetch = (typeof window === 'undefined' || window.name === 'nodejs') ? require('node-fetch') : window.fetch;
+let fetch;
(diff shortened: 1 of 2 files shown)


src.ts/utils/web.ts
```diff
@@ -11,7 +11,29 @@ export interface ConnectionInfo {
     headers?: { [key: string]: string | number };
 }
 
-const fetch = (typeof window === 'undefined' || window.name === 'nodejs') ? require('node-fetch') : window.fetch;
```
vgrichina: Feel like makes sense to extract into separate package. Like e.g. in wallet I use this package:
https://github.com/nearprotocol/send-json/blob/master/index.js

We should just extend it with other goodies.

frol:
@vgrichina Could you do that? (It will require releases on npmjs, anyway.)

vgrichina:
ok

PR #84, opened 2019-10-07
Use new staking transaction format

bowenwang1996:
Nearcore changed the format of staking transaction and this PR matches the schema change. However, the change is not backward compatible.


vgrichina:
Need to add basic backwards compatibility support for previous version of nearcore (crash on staking attempt).

src.ts/account.ts
```diff
@@ -6,7 +6,7 @@ import { Action, transfer, createAccount, signTransaction, deployContract,
 import { FinalExecutionOutcome, FinalExecutionStatusBasic } from './providers/provider';
 import { Connection } from './connection';
 import {base_decode, base_encode} from './utils/serialize';
-import { PublicKey } from './utils/key_pair';
+import {BlsPublicKey, PublicKey} from './utils/key_pair';
```
janedegtiareva: Do we have docs somewhere on what's the diff between BlsPublicKey and regular PublicKey?

vgrichina:
Needs to add tests for staking with BLS key as well (there are no tests for staking whatsoever)

src.ts/account.ts
bowenwang1996: It's not really used in nearlib. Rather it just holds data without actually doing anything.

src.ts/account.ts
```diff
@@ -173,8 +173,12 @@ export class Account {
         return this.signAndSendTransaction(this.accountId, [deleteKey(PublicKey.from(publicKey))]);
     }
 
-    async stake(publicKey: string | PublicKey, amount: BN): Promise<FinalExecutionOutcome> {
-        return this.signAndSendTransaction(this.accountId, [stake(amount, PublicKey.from(publicKey))]);
+    async stake(publicKey: string | BlsPublicKey, amount: BN): Promise<FinalExecutionOutcome> {
+        const status = await this.connection.provider.status();
+        if (status.version.version < '0.4.0') {
```
vgrichina: this isn't very robust way to compare versions, makes sense to use something like https://github.com/omichelsen/compare-versions

bowenwang1996:
We removed bls in nearcore so this is no longer necessary.

PR #85, opened 2019-10-07
Update error handling done by staging

evgenykuzyakov:
Failure status now provides the error message and error type.
See nearcore change: https://github.com/nearprotocol/nearcore/pull/1408


evgenykuzyakov:
It passes all tests on new nearcore change as well.

evgenykuzyakov:
I actually not sure, we'd remove any legacy stuff. And when should we remove it. Will add a TODO for `adapt...` function.

PR #87, opened 2019-10-09
Adding support for typed errors

evgenykuzyakov:


src.ts/account.ts
```diff
@@ -88,13 +88,13 @@ export class Account {
             waitTime *= TX_STATUS_RETRY_WAIT_BACKOFF;
             i++;
         }
-        throw new Error(`Exceeded ${TX_STATUS_RETRY_NUMBER} status check attempts for transaction ${base_encode(txHash)}.`);
+        throw new TypedError(`Exceeded ${TX_STATUS_RETRY_NUMBER} status check attempts for transaction ${base_encode(txHash)}.`, 'RetiesExceeded');
```
vgrichina: RetriesExceeded

src.ts/account.ts
```diff
@@ -88,13 +88,13 @@ export class Account {
...
     }
 
     private async signAndSendTransaction(receiverId: string, actions: Action[]): Promise<FinalExecutionOutcome> {
         await this.ready;
         if (!this._accessKey) {
-            throw new Error(`Can not sign transactions, initialize account with available public key in Signer.`);
+            throw new TypedError(`Can not sign transactions, initialize account with available public key in Signer.`, 'CantSignError');
```
vgrichina: KeyNotFound

Error suffix generally feels redundant as well

src.ts/account.ts
```diff
@@ -107,7 +107,7 @@ export class Account {
         try {
             result = await this.connection.provider.sendTransaction(signedTx);
         } catch (error) {
-            if (error.message === '[-32000] Server error: send_tx_commit has timed out.') {
```
vgrichina: looks like should have both checks for compatibility with testnet

evgenykuzyakov: I moved this check to provider to generate TimeoutError there

PR #88, opened 2019-10-22
Exposed `chunk` RPC API (nearcore 0.4.2+)

frol:
Ref: https://github.com/nearprotocol/nearcore/pull/1524

### DIFF
--- a/test/providers.test.js
+++ b/test/providers.test.js
@@ -1,38 +1,36 @@
 
 const nearlib = require('../lib/index');
 
-test('json rpc fetch node status', async () => {
+const withProvider = (fn) => {
     const config = Object.assign(require('./config')(process.env.NODE_ENV || 'test'));
     const provider = new nearlib.providers.JsonRpcProvider(config.nodeUrl);
+    return () => fn(provider);
+};
+
+test('json rpc fetch node status', withProvider(async (provider) => {
     let response = await provider.status();
     expect(response.chain_id).toContain('test-chain');
-});
+}));
 
-test('json rpc fetch block info', async () => {
-    const config = Object.assign(require('./config')(process.env.NODE_ENV || 'test'));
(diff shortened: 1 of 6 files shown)


test/providers.test.js
```diff
@@ -13,6 +13,18 @@ test('json rpc fetch block info', async () => {
     const provider = new nearlib.providers.JsonRpcProvider(config.nodeUrl);
     let response = await provider.block(1);
     expect(response.header.height).toEqual(1);
+    let sameBlock = await provider.block(response.header.hash);
+    expect(sameBlock.header.height).toEqual(1);
+});
+
+test('json rpc fetch chunk info', async () => {
+    const config = Object.assign(require('./config')(process.env.NODE_ENV || 'test'));
```
vgrichina: can we extract this into helper or `before`? Seems like every test is loading config and creating provider with it.

frol:
@vgrichina I expect it will pass once nearcore is deployed with the PR, so in theory, we only need to run CI once more once that happens. I am just proactively creating PRs :)

test/providers.test.js
frol: Sure, I will fix it.

vgrichina:
> @vgrichina I expect it will pass once nearcore is deployed with the PR, so in theory, we only need to run CI once more once that happens. I am just proactively creating PRs :)

Do you plan to wait until it is deployed on master testnet or on staging? If you only want to wait for staging deploy (or not wait at all), then it's necessary to gate this test by nearcore version.

frol:
I plan to wait for the API hit `master`. I hope it happens soon since without this API Explorer is useless (I cannot get a transactions list from nearcore otherwise).

frol:
@vgrichina The tests pass on the current (already deployed) master. Staging should work fine as well, though it is down at the moment. The CI deployed nodes needs to be updated to include the `chunk` API.

PR #106, opened 2019-11-13
Prototype WalletAccount.requestSignTransactions

vgrichina:
NOTE: This leaves out error handling so far. Works for apps which retrieve current state after redirect (any CRUD app basically).

The plan for error/status handling is following:
- user rejecting to sign has to be explicit error passed with callback
- every sign request will have random ID to avoid other apps faking callbacks
- "pending" transactions and hashes persisted before redirect
- when redirected back to app – monitor tx status for all persisted hashes
- allow app to subscribe to status of "pending" transactions when setting up `WalletConnection`


src.ts/wallet-account.ts
```diff
@@ -112,4 +140,97 @@ export class WalletAccount {
...
+    async accessKeyMatchesTransaction(accessKey, receiverId: string, actions: Action[]): Promise<boolean> {
+        const { access_key: { permission } } = accessKey;
+        if (permission === 'FullAccess') {
+            return true;
+        }
+
+        if (permission.FunctionCall) {
+            const { receiver_id: allowedReceiverId, method_names: allowedMethods } = permission.FunctionCall;
+            if (allowedReceiverId === receiverId) {
+                return actions.every(({ functionCall }) => functionCall &&
```
evgenykuzyakov: We only allow one `functionCall` action.

PR #108, opened 2019-11-15
Remove LOG: from logging

nearmax:
We don't need to add `LOG: ` to logs.


vgrichina:
hmm, where is the fix itself?

nearmax:
> hmm, where is the fix itself?

Here https://github.com/nearprotocol/nearcore/pull/1487

nearmax:
We should deal with it later: https://github.com/nearprotocol/nearcore/issues/1703

PR #112, opened 2019-11-19
Remove confusing log message about missing public key

vgrichina:
Fixes https://github.com/nearprotocol/nearlib/issues/111


vgrichina:
> this works to clear up the previous error but what if i don't know what a "matching key pair" looks like or where to get one? is there a clear path to resolve the issue at this point? a deep link into the docs might go stale unfortunately.

if you have better message, please edit directly )

> zooming out, nearlib shouldn't be logging anything to the console ever, right? i think there are a few other instances of this, let me look

there is convenience feature where logs from smart contract are piped back. We should make possible to override logger though to disable.

> also out of scope but what about attaching an error code that can be looked up somewhere that holds all the other useful references like docs, examples, etc.

that's work in progress with `TypedError` I think

PR #113, opened 2019-11-19
Remove LOG: prefix from log message in test

fckt:
Tests are executing against `nearcore` `staging` branch and we are failing with. But in the same time, we can't merge https://github.com/nearprotocol/nearcore/pull/1715 into `staging` cause CI fails to run test of nearlib (it uses master). How to resolve this ?

### DIFF
--- a/lib/providers/provider.d.ts
+++ b/lib/providers/provider.d.ts
@@ -7,11 +7,16 @@ export interface SyncInfo {
     latest_state_root: string;
     syncing: boolean;
 }
+interface Version {
+    version: string;
+    build: string;
+}
 export interface NodeStatusResult {
     chain_id: string;
     rpc_addr: string;
     sync_info: SyncInfo;
     validators: string[];
+    version: Version;
 }
 export declare type BlockHash = string;
 export declare type BlockHeight = number;
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "localstorage-memory": "^1.0.3",
     "near-hello": "^0.3.1",
     "rimraf": "^3.0.0",
+    "semver": "^6.3.0",
     "tslint": "^5.18.0",
     "typedoc": "^0.15.0",
     "typedoc-plugin-markdown": "^2.2.6",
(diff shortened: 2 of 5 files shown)


vgrichina:
> Tests are executing against `nearcore` `staging` branch and we are failing with. But in the same time, we can't merge [nearprotocol/nearcore#1715](https://github.com/nearprotocol/nearcore/pull/1715) into `staging` cause CI fails to run test of nearlib (it uses master). How to resolve this ?

Need to guard given tests by `nearcore` version

nearmax:
> > Tests are executing against `nearcore` `staging` branch and we are failing with. But in the same time, we can't merge [nearprotocol/nearcore#1715](https://github.com/nearprotocol/nearcore/pull/1715) into `staging` cause CI fails to run test of nearlib (it uses master). How to resolve this ?
> 
> Need to guard given tests by `nearcore` version

How do you do that?

vgrichina:
> > Need to guard given tests by `nearcore` version
> 
> How do you do that?

Query version from server using `/status` endpoint, compare to expected version using https://www.npmjs.com/package/semver, and then execute different tests based on result.

vgrichina:
@fckt in addition to guarding test also needs to revert whitespace changes. Formatting should match `.eslintrc` rules.

fckt:
Done.
But the problem is that we don't know in witch version the original PR going to be enabled. A solution could be just use a regex, not a strict compare. What do think @vgrichina?

vgrichina:
> But the problem is that we don't know from witch version this will going to be enabled. 

Let's merge this when we know which version then? Not sure how regex would help.

fckt:
> Not sure how regex would help.

I mean just check for a substring instead of the strict compression. 

> Let's merge this when we know which version then?

ok

nearmax:
@vgrichina Could you please review it?

vgrichina:
@nearmax @fckt  merged it for now, we can just update the version later if breaks CI

bowenwang1996:
looks like the PR in nearcore is not merged https://github.com/nearprotocol/nearcore/pull/1715, which blocks the PR https://github.com/nearprotocol/nearcore/pull/1871 that updates testnet

nearmax:
Correct, @janedegtiareva decided when to merge.

PR #126, opened 2019-12-03
Add util to convert units of account balance to near

janedegtiareva:


src.ts/utils/format.ts
```diff
@@ -0,0 +1,37 @@
+const BN = require ('bn.js');
+
+// Exponent for calculating how many units of account balance are in one near.
+const exp = 24;
```
vgrichina: nit: constants ideally should be ALL_CAPS.  So makes sense to use something like `NEAR_NOMINATION_EXP`.

src.ts/utils/format.ts
```diff
@@ -0,0 +1,37 @@
...
+export function convertAccountBalanceAmountToNear(balance: string): string {
+    const amtBN = new BN(balance, 10);
+    if (amtBN.lte(unitsInOneNear)) {
+        return `0.${balance.padStart(exp, '0')}`;
+    }
+    return `${amtBN.div(unitsInOneNear).toString(10, 0)}.${amtBN.mod(unitsInOneNear).toString(10, 0)}`;
+}
+
+/**
+ *
```
vgrichina: nit: missing doc

src.ts/utils/format.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param amt
+ */
+export function nearAmountToAccountBalanceAmount(amt) {
```
vgrichina: missing type for `amt` and return type

src.ts/utils/format.ts
```diff
@@ -0,0 +1,37 @@
+const BN = require ('bn.js');
+
+// Exponent for calculating how many units of account balance are in one near.
+const exp = 24;
+// actual number of units of account balance in one near.
+const unitsInOneNear = new BN('10', 10).pow(new BN(exp, 10));
+
+/**
+ * Convert account balance to near.
```
vgrichina: it's good idea to explain conversion here (e.g. refer to constants used to convert balance). "Convert account balance to near" has no additional info compared to method name.

src.ts/utils/format.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param balance
+ */
+export function convertAccountBalanceAmountToNear(balance: string): string {
```
vgrichina: `balance` argument should better support `BN` as well

src.ts/utils/format.ts
vgrichina: `convertAccountBalanceAmountToNear` and `nearAmountToAccountBalanceAmount` look confusing to me. Why one of them is `convert` and another isn't? Feels less confusing if we stick to `format` for one direction and `parse` for another. E.g. `parseBalance` and `formatBalance`.

src.ts/utils/format.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export function nearAmountToAccountBalanceAmount(amt) {
+    if (!amt) { return amt; }
+    amt = amt.trim();
+    const split = amt.split('.');
+    if (split.length === 1) {
+        return `${amt.padEnd(exp + 1, '0')}`;
+    }
+    if (split.length > 2 || split[1].length > exp) {
+        throw new Error('Invalid input format');
```
vgrichina: it's best to include input value into error, e.g. 

```js
throw new Error(`Cannot parse '${amt}' as NEAR amount`);
```

src.ts/utils/format.ts
janedegtiareva: done

src.ts/utils/format.ts
janedegtiareva: added

src.ts/utils/format.ts
janedegtiareva: reworded to be more informative

src.ts/utils/format.ts
janedegtiareva: fixed

src.ts/utils/format.ts
janedegtiareva: done

src.ts/utils/format.ts
janedegtiareva: discussed offline.

src.ts/utils/format.ts
janedegtiareva: Balance might not be the best option (e.g. for amount to transfer from one acct to another). Maybe parseNearAmount and formatNearAmount?

frol:
I would like to propose a `NearAmount` / `NearToken` type with the following interface:

```ts
class NearToken {
    yoctoNear: BN;

    static fromYoctoNear(value: string): NearToken;
    static fromAttoNear(value: string): NearToken;
    static fromNear(value: string): NearToken;
    
    formatYoctoNear(): string;
    formatAttoNear(): string;
    formatNear(): string;
}
```

This way we can use this class everywhere in a backward-compatible way. It would be even better if we can automatically parse the RPC responses replacing the amount string values with `NearToken` type.

janedegtiareva:
> I would like to propose a `NearAmount` / `NearToken` type with the following interface:
> 
> ```ts
> class NearToken {
>     yoctoNear: BN;
> 
>     static fromYoctoNear(value: string): NearToken;
>     static fromAttoNear(value: string): NearToken;
>     static fromNear(value: string): NearToken;
>     
>     formatYoctoNear(): string;
>     formatAttoNear(): string;
>     formatNear(): string;
> }
> ```
> 
> This way we can use this class everywhere in a backward-compatible way. It would be even better if we can automatically parse the RPC responses replacing the amount string values with `NearToken` type.

Should we be exposing yocto/atto as part of the actual function names? It seems better that the callers don't have to know/care so much about the internal representation of the balance. In an ideal world, the change attoNEAR->yoctoNEAR should produce no code changes in clients other than updating nearlib version.

frol:
@janedegtiareva I don't care much about the atto-/yocto- prefixes, but I prefer working with specific types instead of just a big number or a string.

PR #145, opened 2019-12-14
Tokens formatting -- add a param for significant digits + tests 

janedegtiareva:


src.ts/utils/format.ts
```diff
@@ -9,13 +9,17 @@ const NEAR_NOMINATION = new BN('10', 10).pow(new BN(NEAR_NOMINATION_EXP, 10));
  * Convert account balance value from internal units (currently yoctoNEAR) to NEAR.
  * @param balance
  */
-export function formatNearAmount(balance: string): string {
+export function formatNearAmount(balance: string, digits?: number): string {
+    balance = trimLeadingZeroes(balance);
+    console.log(balance);
```
icerove: leftover?

dist/nearlib.js
```diff
@@ -1252,6 +1256,25 @@ function parseNearAmount(amt) {
...
 exports.parseNearAmount = parseNearAmount;
+/**
+ * Truncates the fraction part of a number to a given number of digits and then trims extra zeroes.
+ * @param value
+ * @param digits
+ */
+function truncateFractionPart(value, digits) {
+    if (digits && value.length > digits) {
+        if (value[digits] >= '5') {
+            return value.substring(0, digits - 1) + String.fromCharCode(value[digits - 1].charCodeAt(0) + 1);
```
vgrichina: Looks like test with say `0.999999999999999` is missing.

dist/nearlib.js
```diff
@@ -1252,6 +1256,25 @@ function parseNearAmount(amt) {
     return `${wholePart.add(fractionPart).toString(10, 0)}`;
 }
 exports.parseNearAmount = parseNearAmount;
+/**
+ * Truncates the fraction part of a number to a given number of digits and then trims extra zeroes.
+ * @param value
```
vgrichina: why add @param tags with no docs?

janedegtiareva: I removed the fn so it's no longer relevant

dist/nearlib.js
janedegtiareva: added + fixed a bug

src.ts/utils/format.ts
janedegtiareva: fixed

PR #152, opened 2019-12-18
Add checks for number of arguments in the transaction

ilblackdragon:
Errors otherwise are unintelligible.

Is there a better way to do this?

### DIFF
--- a/src.ts/transaction.ts
+++ b/src.ts/transaction.ts
@@ -27,6 +27,9 @@ export class AccessKey extends Assignable {
 }
 
 export function fullAccessKey(): AccessKey {
+    if (arguments.length != 0) {
+        throw new Error(`Wrong number of arguments ${arguments.length}: expected none`);
+    }
     return new AccessKey({ nonce: 0, permission: new AccessKeyPermission({fullAccess: new FullAccessPermission({})}) });
 }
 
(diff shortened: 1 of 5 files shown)


src.ts/transaction.ts
```diff
@@ -27,6 +27,9 @@ export class AccessKey extends Assignable {
 }
 
 export function fullAccessKey(): AccessKey {
+    if (arguments.length != 0) {
+        throw new Error(`Wrong number of arguments ${arguments.length}: expected none`);
```
vgrichina: Let's create specific error here https://github.com/nearprotocol/nearlib/blob/master/src.ts/utils/errors.ts

vgrichina: I think it's also best to include arguments into error message instead of just length (faster to debug in most cases)

vgrichina:
I wonder if underlying issue is that we don't have proper errors in Borsh serializer. We know all the field types there, so can verify them generally and report field path with wrong type.

vgrichina:
also make sure this doesn't fail in `ci` env https://travis-ci.com/nearprotocol/nearlib/jobs/268369102

ilblackdragon:
You might be right, the errors in borsh should be improved as well.
It's sad that TS actually doesn't have a way to automatically enforce types when used from JS.

I think you have a better idea how to do this, so I'm going to close this.

PR #155, opened 2019-12-18
Update FinalExecutionOutCome

bowenwang1996:
Needed for https://github.com/nearprotocol/nearcore/pull/1865, which includes the transaction itself in `FinalExecutionOutcome` of a transaction. @frol How do we plan to use the transaction in the return value? I currently put `any` in the interface. @vgrichina I think this is a breaking change which is not backward compatible. I guess I don't have to rename the fields but then not sure what the field name for transaction should be.


vgrichina:
@bowenwang1996 you need to make this backwards compatible, i.e. support both response formats

bowenwang1996:
@vgrichina It's now backward compatible. @frol I took the easier route of patching txResult instead of writing a new function because I don't want to also update near-shell.

PR #163, opened 2019-12-30
Typed rpc errors implementation

fckt:


test/utils/rpc-errors.test.js
```diff
@@ -0,0 +1,17 @@
...
+    let rpc_error = {
+        TxExecutionError: {
+            Action: {
+                index: 1,
+                kind: { AccountAlreadyExists: { account_id: 'bob.near' } }
+            }
+        }
+    };
+    let error = parseRpcError(rpc_error);
+    console.log(error);
```
vgrichina: seems like debug leftover

test/utils/rpc-errors.test.js
```diff
@@ -0,0 +1,17 @@
+const nearlib = require('../../lib/index');
+const { parseRpcError, AccountAlreadyExists, TxExecutionError } = nearlib.utils.rpc_errors;
+
+test('test parseRpcError', async () => {
+    let rpc_error = {
+        TxExecutionError: {
+            Action: {
+                index: 1,
```
vgrichina: what if there are errors in multiple actions? Generally I think test should check more examples ideally as parsing code is non-trivial.

src.ts/generated/rpc_errors_types.ts
```diff
@@ -0,0 +1,156 @@
...
+    public account_id;
+    public predecessor_id;
+}
+
+export class DeleteAccountHasRent extends Action {
+    public balance;
+    public account_id;
+}
+
+export const CLASSMAP = {
```
vgrichina: Not sure whether this is needed. Why not just use module itself as a map?

```
import * as CLASSMAP from 'rpc_errors_types'
```

vgrichina:
Given that errors don't have messages now – we need some system to generate messages.

Ideally we should be able to re-use message definitions among implementations in different languages (TypeScript, Swift and C# at the moment). I think https://mustache.github.io is decent fit for this, let me know if you have other suggestions.

test/utils/rpc-errors.test.js
fckt: > what if there are errors in multiple actions?

That's interesting.. Right now we consider 1 result = 1 error. Maybe we would want to change it in the future, but it would require quite a bit of changes of how node is works.

> Generally I think test should check more examples ideally as parsing code is non-trivial.

Yeah, sure

vgrichina: @fckt so is it returning error on the first failed action? Does it even try to execute other ones or stops there? If it stops there, I think it's fine, it is simple enough model.

fckt: > Does it even try to execute other ones or stops there?

yeah, it fails on the first error

src.ts/generated/rpc_errors_types.ts
fckt: Addressed! Thanks!

test/utils/rpc-errors.test.js
fckt: thanks!

fckt:
> Given that errors don't have messages now – we need some system to generate messages.
Ideally we should be able to re-use message definitions among implementations in different languages (TypeScript, Swift and C# at the moment). I think https://mustache.github.io is decent fit for this, let me know if you have other suggestions.

The messages must be generated on the client side. `nearlib` throws an error (not formatted) and the near-shell or other client (web wallet). The error formatting could differ from client to client (if I understand the goal of the whole idea correctly).

vgrichina:
> The messages must be generated on the client side. nearlib throws an error (not formatted) and the near-shell or other client (web wallet) catch it and do displays. 

I agree that client-side there is usually most context to provide best message to user. However developers also are generally lazy and we cannot really expect every client app to cover all possible errors. There must be reasonable defaults readily available which can then be overridden by given client app.

> The error formatting could differ from client to client (if I understand the goal of the whole idea correctly).

It can differ, sure. But it's multiplied maintenance work. Baseline error messages can be reused between different client libraries as long as they use cross-platform format (e.g. JSON or YAML with embedded Mustache templates).

test/account.test.js
```diff
@@ -168,13 +176,21 @@ describe('with deploy contract', () => {
     });
 
     test('can get assert message from method result', async () => {
-        await expect(contract.triggerAssert()).rejects.toThrow(/Transaction .+ failed.+expected to fail.+/);
-        if (afterVersion('0.4.10')) {
```
vgrichina: We can remove tests for before 0.4.11

vgrichina:
this needs to pass CI

fckt:
What should I do with the "bundlewatch"?

src.ts/account.ts
```diff
@@ -118,9 +119,19 @@ export class Account {
         this.printLogs(signedTx.transaction.receiverId, flatLogs);
 
         if (typeof result.status === 'object' && typeof result.status.Failure === 'object') {
-            throw new TypedError(
-                `Transaction ${result.transaction_outcome.id} failed. ${result.status.Failure.error_message}`,
-                result.status.Failure.error_type);
+            // if error data has error_message and error_type properties, we consider that node returned an error in the old format
```
vgrichina: let's add a TODO with GH issue link to track removal of temporary stuff (not a blocker for merge)

src.ts/providers/json-rpc-provider.ts
```diff
@@ -76,7 +72,15 @@ export class JsonRpcProvider extends Provider {
         const response = await fetchJson(this.connection, JSON.stringify(request));
         if (response.error) {
             if (typeof response.error.data === 'object') {
-                throw new TypedError(response.error.data.error_message, response.error.data.error_type);
+                if (typeof response.error.data.error_message === 'string' && typeof response.error.data.error_type === 'string') {
```
vgrichina: let's add a TODO with GH issue link to track removal of temporary stuff (not a blocker for merge)

src.ts/providers/json-rpc-provider.ts
```diff
@@ -76,7 +72,15 @@ export class JsonRpcProvider extends Provider {
...
+                    // if error data has error_message and error_type properties, we consider that node returned an error in the old format
+                    throw new TypedError(response.error.data.error_message, response.error.data.error_type);
+                } else {
+                    // overwise we need to transform the new format into the old one for a transition period
+                    throw parseIntoOldTypedError(response.error.data);
```
vgrichina: I think we actually want to have new errors but with messages masquerading for old errors. I'll do the change myself before merging.

fckt: Sure, we can just add a new field to `TypedError` which will hold an error in a new format.

src.ts/providers/json-rpc-provider.ts
fckt: https://github.com/nearprotocol/nearlib/issues/173

PR #168, opened 2020-01-07
Parse NEAR amount with commas

vgrichina:


vgrichina:
@nearmax do you think we should just remove commas when parsing or actually validate?

vgrichina:
@nearmax friendly ping

PR #180, opened 2020-01-15
fix yarn audit warnings

janedegtiareva:


vgrichina:
`yarn.lock` doesn't impact library users BTW, only the ones who develop nearlib.

PR #182, opened 2020-01-16
run yarn audit on build

janedegtiareva:


vgrichina:
As I wrote before, generally I think this is redundant to what http://dependabot.com/ does (in more convenient way as it automatically creates PRs with fixes).

But if you think it's useful to have on CI – I think it makes sense to have it as the very last step (so that tests are executed even if it failed).

janedegtiareva:
> As I wrote before, generally I think this is redundant to what http://dependabot.com/ does (in more convenient way as it automatically creates PRs with fixes).
> 
> But if you think it's useful to have on CI – I think it makes sense to have it as the very last step (so that tests are executed even if it failed).

It would make me feel more warm and fuzzy inside about security of our dependencies, so I'd rather have it (it's pretty quick so not a large cost). Moved to last step of the build.

vgrichina:
> it's pretty quick so not a large cost

Not worried about that cost at all. My only worry is that dependabot is already given us a lot of churn with mostly irrelevant security stuff, so I feel like CI can be blocked too often for false positives. But we can see how it goes on practice.

vgrichina:
Another concern is that `yarn.lock` updates generally only give false sense of security for libraries. As anyone who installs library through package manager would get dependencies resolved completely ignoring the `yarn.lock` in this repo.

PR #189, opened 2020-01-21
Change the way access key is fetched by Account

vgrichina:
This enables using Account in read-only mode.


vgrichina:
Merging as it's blocker for https://github.com/nearprotocol/near-contract-helper/pull/69

PR #191, opened 2020-01-22
Increase default gas amount for function calls

vgrichina:
Set to 1/1000 of current `REACT_APP_ACCESS_KEY_FUNDING_AMOUNT` on testnet. 
So app can do 1000 transactions with default access key allowance.

Fixes https://github.com/nearprotocol/nearlib/issues/176

### DIFF
--- a/src.ts/account.ts
+++ b/src.ts/account.ts
@@ -15,7 +15,7 @@ import { parseRpcError } from './utils/rpc_errors';
 // the originator.
 // Default value is set to equal to max_prepaid_gas as discussed here:
 // https://github.com/nearprotocol/nearlib/pull/191#discussion_r369671912
-const DEFAULT_FUNC_CALL_GAS = new BN("10000000000000000");
+const DEFAULT_FUNC_CALL_GAS = new BN('10000000000000000');
 
 // Default number of retries before giving up on a transactioin.
 const TX_STATUS_RETRY_NUMBER = 10;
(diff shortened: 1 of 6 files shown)


vgrichina:
@bowenwang1996 is gas measured in yoctoNEAR or smth else?

bowenwang1996:
@vgrichina gas is not measure in anything related to NEAR, but the account will have to have gas_amount * gas_price for the transaction to be processed.

src.ts/account.ts
```diff
@@ -10,10 +10,10 @@ import { PublicKey } from './utils/key_pair';
 import { PositionalArgsError } from './utils/errors';
 import { parseRpcError } from './utils/rpc_errors';
 
-// Default amount of tokens to be send with the function calls. Used to pay for the fees
+// Default amount of gas to be sent with the function calls. Used to pay for the fees
 // incurred while running the contract execution. The unused amount will be refunded back to
 // the originator.
-const DEFAULT_FUNC_CALL_AMOUNT = 2000000 * 1000000;
+const DEFAULT_FUNC_CALL_GAS = "100000000000000000";
```
bowenwang1996: This is more than `max_gas_burnt` (2 * 10^14) so probably not a good idea.

vgrichina: So what is gas measured in? How much NEAR is max-gas-burnt?

vgrichina: So what is gas measured in? How much NEAR is max-gas-burnt?

bowenwang1996: Actually I `max_gas_burnt` is not the most relevant thing here since a call can generate receipts that span multiple blocks. So I think we should use `max_prepaid_gas` which is 10^16

vgrichina: @nearmax any concerns re setting this to `max_prepaid_gas`? Seems like with current gas price it's around 0.00001 NEAR.

PR #238, opened 2020-02-25
ReceiverMismatch error message #234

### LINKED ISSUE #234: ReceiverMismatch error description is confusing and non actionable
Generally errors should be actionable (e.g. what do I change to make error go away).

Specifically it took me 10min to figure out `ReceiverMismatch` error by tracing into nearcore:
`"Transaction receiver_id {{tx_receiver}} doesn't match the access key receiver_id {{ak_receiver}}"`

Something more actionable would be:
`"Transaction is sent to receiver_id={{tx_receiver}}, but is signed with function call access key that restricted to only use with receiver_id={{ak_receiver}}. Either switch to use FullAccessKey or change receiver_id in your transaction."`

fckt:
fixes #234

### DIFF
--- a/lib/res/error_messages.json
+++ b/lib/res/error_messages.json
@@ -37,7 +37,7 @@
     "InvalidPublicKey": "VM Logic provided an invalid public key",
     "ActorNoPermission": "Actor {{actor_id}} doesn't have permission to account {{account_id}} to complete the action",
     "RentUnpaid": "The account {{account_id}} wouldn't have enough balance to pay required rent {{amount}}",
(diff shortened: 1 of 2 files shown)


fckt:
@vgrichina the is a travis CI error which is not corresponds to the change.. How to fix?

vgrichina:
@fckt looks like Travis changed default behavior for yarn and now enforces that lock file is up to date.

Should be possible to merge after rebasing on top of https://github.com/nearprotocol/nearlib/commit/80cbd568afb275dbe1c043f8f1bbda0545b33217

I'll do more research to see if we want to keep the behavior though, as seems like it conflicts with dependabot-based approach to updating dependencies.

PR #245, opened 2020-02-28
add gas_price to provider

### LINKED ISSUE #242: Expose gas price
@icerove it is quite straight-forward:

1. Add `gasPrice` method to the Provider interface: https://github.com/nearprotocol/nearlib/blob/d935c9a7796d0f835a3b232315f7960fcf21a255/src.ts/providers/provider.ts#L138-L147
2. Implement it for [JsonRpcProvider](https://github.com/nearprotocol/nearlib/blob/d935c9a7796d0f835a3b232315f7960fcf21a255/src.ts/providers/json-rpc-provider.ts)

Here is the relevant API: https://docs.nearprotocol.com/docs/interaction/rpc#gas-price

icerove:
fix #242


src.ts/providers/json-rpc-provider.ts
```diff
@@ -60,6 +60,10 @@ export class JsonRpcProvider extends Provider {
         return this.sendJsonRpc('chunk', [chunkId]);
     }
 
+    async gasPrice(blockId: BlockId|null): Promise<GasPrice> {
+        return this.sendJsonRpc('Gas Price', [blockId])
```
bowenwang1996: well, if you add a test you will see that this is not the way to call this rpc. I believe that it should be 'gas_price', but please do add a test to make sure that this works properly.

icerove:
I don't know whether the gas price is all the same in our chain and write ugly test since I only know this gas price from rpc doc.

test/providers.test.js
```diff
@@ -21,6 +21,16 @@ test('json rpc fetch block info', withProvider(async (provider) => {
     expect(sameBlock.header.height).toEqual(height);
 }));
 
+test('json rpc fetch gas price', withProvider(async (provider) => {
+    const testHash = '8ehA3NYL5uSF8zefbnqnz66twYJ45rfst6SrqBNv7oka';
```
vgrichina: Looks like this will break on the next hard-fork (already doesn't see to pass CI). Shouldn't hardcode a specific hash in tests.

src.ts/providers/json-rpc-provider.ts
```diff
@@ -60,6 +60,10 @@ export class JsonRpcProvider extends Provider {
         return this.sendJsonRpc('chunk', [chunkId]);
     }
 
+    async gasPrice(blockId: BlockId|null): Promise<GasPrice> {
+        return this.sendJsonRpc('gas_price', [blockId])
+    }
+
     private async sendJsonRpc(method: string, params: any[]): Promise<any> {
```
vgrichina: might be good time to make this method public, so that we don't have to expose every RPC

src.ts/providers/json-rpc-provider.ts
```diff
@@ -60,6 +60,10 @@ export class JsonRpcProvider extends Provider {
         return this.sendJsonRpc('chunk', [chunkId]);
     }
 
+    async gasPrice(blockId: BlockId|null): Promise<GasPrice> {
```
vgrichina: given that `null` is allowed for `blockId` – tests should also check if method works with `null`

icerove: I know that but do not know the exact price for latest block. We should not suppose it is all the same for '5000', right? so I really want to know how to write that test. just to see whether to get the result? 
I know we have gas price in block header from rpc, but we do not have that in nearlib block header. So really bother me how to write test

src.ts/providers/json-rpc-provider.ts
frol: Well, given we want to expose all the APIs, it does not seem right to make it public, but I'll let you make the decision. Anyway, it should not be part of this PR.

src.ts/providers/json-rpc-provider.ts
frol: I think it should be enough to check that the value is there, is numeric and is greater than 0.

test/providers.test.js
frol: Let's use the status API to fetch the latest block hash and block height from it.

src.ts/providers/json-rpc-provider.ts
icerove: let's wait to see whether others is okay with this

test/providers.test.js
```diff
@@ -21,6 +21,19 @@ test('json rpc fetch block info', withProvider(async (provider) => {
...
+    let stat = await provider.status();
+    let height = stat.sync_info.latest_block_height - 1;
+    let block = await provider.block(height);
+    let hash = block.header.hash;
+    let heightPrice = await provider.gasPrice(height);
+    let hashPrice = await provider.gasPrice(hash);
+    let lastestPrice = await provider.gasPrice(null);
+    expect(typeof heightPrice).toBe('object');
+    expect(typeof hashPrice).toBe('object');
+    expect(typeof lastestPrice).toBe('object');
```
ailisp: expect here to be a string that can be convert to an number, and that number > 0

src.ts/providers/json-rpc-provider.ts
bowenwang1996: Since @frol's change that exposes config is already merged, we can check that this value is at least minimum gas price.

vgrichina: @bowenwang1996 note that we also need this change in both `ci` and `ci-staging` nets, not only merged

src.ts/providers/json-rpc-provider.ts
vgrichina: @frol there is really no practical benefit to not make it public IMO. There can be various reasons why somebody has to stay on older version of library, but still call newest RPC. This seems as practicality vs theoretical correctness trade-off.

src.ts/providers/json-rpc-provider.ts
icerove: what is at least minimum gas price? @bowenwang1996

src.ts/providers/json-rpc-provider.ts
frol: @vgrichina Double negative is hard to reason about in English. I am neutral about promoting it to the public interface, but I want to make it separately from this PR :man_shrugging:

src.ts/providers/json-rpc-provider.ts
frol: @bowenwang1996 That one is not even implemented in nearlib yet, and only available on the master branch, so it seems like a chicken and an egg problem here.

test/providers.test.js
```diff
@@ -21,6 +21,17 @@ test('json rpc fetch block info', withProvider(async (provider) => {
...
 
+test('json rpc fetch gas price', withProvider(async (provider) => {
+    let stat = await provider.status();
+    let height = stat.sync_info.latest_block_height - 1;
+    let block = await provider.block(height);
+    let hash = block.header.hash;
+    let heightPrice, hashPrice, latestPrice = await Promise.all([provider.gasPrice(height), provider.gasPrice(hash), provider.gasPrice(null)]);
+    expect(heightPrice.gas_price).toBeGreaterThan(0);
+    expect(hashPrice.gas_price).toBeGreaterThan(0);
+    expect(latestPrice.gas_price).toBeGreaterThan(0);
```
frol:
```js
const status = await provider.status();
for (gasPricePresponse in await Promise.all([
    provider.gasPrice(stat.sync_info.latest_block_height),
    provider.gasPrice(stat.sync_info.latest_block_hash),
    provider.gasPrice(null)
])) {
    expect(gasPriceResponse.gas_price).toBeGreaterThan(0);
}
```

src.ts/providers/json-rpc-provider.ts
vgrichina: >  but I want to make it separately from this PR 🤷‍♂

seems reasonable

src.ts/providers/json-rpc-provider.ts
bowenwang1996: Okay we can add that later.

test/providers.test.js
icerove: not working
image

vgrichina:
@ailisp @bowenwang1996 looks like this is blocked by given RPC being absent from current CI network:
https://travis-ci.com/nearprotocol/nearlib/jobs/293109758

bowenwang1996:
@ailisp can you take a look at CI?

frol:
@icerove Please, rebase it and let's see if CI gets fixed.

src.ts/providers/json-rpc-provider.ts
frol: Just for the reference: https://github.com/near/near-api-js/pull/272

Issue #243, opened 2020-02-26
Support for creating and funding linkdrop links


kcole16:
1.  1.  Story

As an app developer, I want to fund the creation of my users` accounts, so they can login and use my app.

1.  1.  Flow

1\. Generate new key pair (pub key/private key)
 2\. Call <span class="error">[Linkdrop]</span>([https://github.com/nearprotocol/near-linkdrop](https://github.com/nearprotocol/near-linkdrop)) contract's `send` function with `public_key` as the argument, and an attached NEAR balance
 3\. Redirect user to the wallet with "fundingKey" argument

1.  1.  1.  1.  Unhappy Path

*   [ ] Surface error if call to Linkdrop contract fails

potatodepaulo:
@kcole16 was this resolved?

kcole16:
Not yet (that I’m aware of). cc @vgrichina

potatodepaulo:
@kcole16, this solution just needs to be generalized to near-api-js correct?

kcole16:
Check this PR for more context: https://github.com/near/near-api-js/pull/256

vgrichina:
@potatodepaulo do we have specific customer who wants to integrate this? If so – I think it will help to do demo for them first and then refactor into library once we have it working end to end.

volovyk-s:
@mikedotexe @kcole16 do we still need this functionality? PR related to this issue was closed: https://github.com/near/near-api-js/pull/256
Should we close this issue?

vgrichina:
I think on client side it's relatively easy, more interesting is to create smart contract SDK which would allow to send more customised linkdrops. Like allow to avoid attaching 1 N for gas fees, allow to transfer NFTs on claim, etc. Basically make it easy to send any kind of stuff on NEAR as a linkdrop.

vgrichina:
> I think on client side it's relatively easy

I mean in the sense that you just create new `KeyPair` and then make regular smart contract `functionCall` transaction to whatever linkdrop contract you want to use

vikinatora:
Closing as issue isn't relevant anymore.

PR #256, opened 2020-03-10
Add fundingKey (linkdrop) support to WalletAccount

kcole16:
For issue #243 

Added support for an optional `fundingKey` parameter when calling `requestSignIn`, so devs can use the Linkdrop functionality.

See PR in wallet: https://github.com/nearprotocol/near-wallet/pull/440

### DIFF
--- a/src.ts/wallet-account.ts
+++ b/src.ts/wallet-account.ts
@@ -71,7 +71,7 @@ export class WalletConnection {
      *     onSuccessHref,
      *     onFailureHref);
      */
-    async requestSignIn(contractId: string, title: string, successUrl: string, failureUrl: string) {
+    async requestSignIn(contractId: string, title: string, successUrl: string, failureUrl: string, fundingKey: string) {
         if (this.getAccountId() || await this._keyStore.getKey(this._networkId, this.getAccountId())) {
             return Promise.resolve();
         }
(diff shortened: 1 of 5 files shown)


src.ts/wallet-account.ts
```diff
@@ -71,7 +71,7 @@ export class WalletConnection {
      *     onSuccessHref,
      *     onFailureHref);
      */
-    async requestSignIn(contractId: string, title: string, successUrl: string, failureUrl: string) {
+    async requestSignIn(contractId: string, title: string, successUrl: string, failureUrl: string, fundingKey: string) {
```
vgrichina: I think I'll do alternate PR to do this, because this method also needs other changes. We need to remove `title` and `app_url` as they can be spoofed by caller. I also want to consolidate callbacks into single `callbackUrl` like in `requestSignTransactions`. 

Also we'll need to do it in backward compatible way and I am not sure if it would work well to always require `fundingKey`. Like if I just started doing stuff on NEAR (generated project from template) – what do I use as `fundingKey`? It seems like it immediately requires me to deploy another contract to just get stuff going. I'd expect this value also not to change between different calls – so maybe we should just put it into config (similar to `walletUrl`) and provide reasonable default (our free faucet)?

kcole16: Right, might be worth making `fundingKey` an optional arg. 

We're actually using a single contract currently [linkdrop-test-1](https://explorer.nearprotocol.com/accounts/linkdrop-test-1), which is defined in the wallet. Though it might make sense to move this to config here, in case an app wanted to use their own.

vgrichina: > Though it might make sense to move this to config here, in case an app wanted to use their own.

I think the whole system only makes sense if app uses it's own contract. Otherwise how it would generate `fundingKey` at all?

kcole16: Thought the same, but after speaking with @ilblackdragon, can work with a single contract.

1. App generates a keypair
2. App calls [Linkdrop contract's](https://github.com/nearprotocol/near-linkdrop) `send` with attached NEAR and public key from **1.** as the argument 
   - Public key is added as a limited access key to Linkdrop, and mapped to the deposit amount
3.  Private key from **1.** is returned to the user as `fundingKey`, then used in the wallet to call `create_account_and_claim` in the Linkdrop contract
    - `fundingKey` signs the TX, with Linkdrop as the accountID
    - New account is created, amount deposited is added to the account, limited access key is removed from Linkdrop

kcole16:
@vgrichina Should we close this PR?

PR #276, opened 2020-03-31
Keypair from seed

ailisp:
Borrow from rust code, and verified by using the `near` seed account to create accounts in devnet

### DIFF
--- a/src.ts/utils/key_pair.ts
+++ b/src.ts/utils/key_pair.ts
@@ -88,16 +88,16 @@ export abstract class KeyPair {
     }
     
     static fromSeed(seedStr: string): KeyPair {
-        let seed = new Uint8Array(32);
+        const seed = new Uint8Array(32);
         for (let i = 0; i < 32; i++) {
             if (i < seedStr.length) {
                 seed[i] = seedStr.charCodeAt(i);
             } else {
-                seed[i] = " ".charCodeAt(0);
+                seed[i] = ' '.charCodeAt(0);
             }
         }
-        let keyPair = nacl.sign.keyPair.fromSeed(seed);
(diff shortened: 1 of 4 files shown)


src.ts/utils/key_pair.ts
```diff
@@ -86,6 +86,20 @@ export abstract class KeyPair {
             throw new Error('Invalid encoded key format, must be <curve>:<encoded key>');
         }
     }
+    
+    static fromSeed(seedStr: string): KeyPair {
```
vgrichina, line 90: What is the purpose for this, do we need to support this specific format of seed phrases?

Note that we already have library with support for conforming BIP39 seed phrases (compatible with other wallets):
https://github.com/nearprotocol/near-seed-phrase/blob/master/index.js

src.ts/utils/key_pair.ts
```diff
@@ -86,6 +86,20 @@ export abstract class KeyPair {
...
+        const seed = new Uint8Array(32);
+        for (let i = 0; i < 32; i++) {
```
vgrichina, line 92: it seems like using only 32 characters from a phrase which supposedly contains common words doesn't provide nearly as much entropy as needed for strong 256-bit key. It's basically only enough for about 6 word seed phrase. I suggest we abandon this scheme altogether (i.e. not use it in Rust either).

bowenwang1996: I think we use this in rust mostly for testing purposes.

src.ts/utils/key_pair.ts
ailisp: Good point! seems this change isn't very helpful for a general api-js user and they should seek for near-seed-phrase

PR #280, opened 2020-04-02
Fix timeout message so retry works

ailisp:
accounts.ts signAndSendTransaction check type of error is TimeoutError to retry txn status. However jsonrpc response of timeout changes so the retry wasn't executed.

### DIFF
--- a/src.ts/providers/json-rpc-provider.ts
+++ b/src.ts/providers/json-rpc-provider.ts
@@ -80,6 +80,8 @@ export class JsonRpcProvider extends Provider {
                 const errorMessage = `[${response.error.code}] ${response.error.message}: ${response.error.data}`;
                 if (errorMessage === '[-32000] Server error: send_tx_commit has timed out.') {
                     throw new TypedError('send_tx_commit has timed out.', 'TimeoutError');
+                } else if (errorMessage === '[-32000] Server error: Timeout') {
+                    throw new TypedError('Timeout', 'TimeoutError');
                 } else {
                     throw new TypedError(errorMessage);
                 }


src.ts/providers/json-rpc-provider.ts
```diff
@@ -80,6 +80,8 @@ export class JsonRpcProvider extends Provider {
                 const errorMessage = `[${response.error.code}] ${response.error.message}: ${response.error.data}`;
                 if (errorMessage === '[-32000] Server error: send_tx_commit has timed out.') {
                     throw new TypedError('send_tx_commit has timed out.', 'TimeoutError');
+                } else if (errorMessage === '[-32000] Server error: Timeout') {
+                    throw new TypedError('Timeout', 'TimeoutError');
```
vgrichina, line 84: why not throw the same error as above? Also why do we still have a kludge, shouldn't nearcore already return error with type?

ailisp: I got a lot errorMessage === '[-32000] Server error: Timeout from devnet, It hit the else case here => return TypedError(errorMessage). And then:
```
export class TypedError extends Error {
    type: string;
    constructor(message?: string, type?: string) {
        super(message);
        this.type = type || 'UntypedError';
    }
}
```
It becomes an "UntypedError", but this should be an 'TimeoutError' type to trigger retryTxResult in account.ts:
```
        try {
            result = await this.connection.provider.sendTransaction(signedTx);
        } catch (error) {
            if (error.type === 'TimeoutError') {
                result = await this.retryTxResult(txHash, this.accountId);
            } else {
                throw error;
            }
        }
```

vgrichina: @ailisp
> I got a lot errorMessage === '[-32000] Server error: Timeout from devnet, It hit the else case here => return TypedError(errorMessage). And then:

that sounds like a problem with nearcore to me. `devnet` should already be returning structured error info, including type. 

so it's necessary to investigate why `throw parseRpcError(response.error.data);` branch isn't triggered

ailisp: So as an example this error message:
```
{'jsonrpc': '2.0', 'error': {'code': -32000, 'message': 'Server error', 'data': 'Timeout'}, 'id': 'dontcare'}
```
from
https://nightly.neartest.com/run/c1ba86d68438f22d25235ba66e2254439d55e9c4_200406_171929/pytest/pytest_sanity_rpc_query_py/stderr

error.data is a string instead of object, therefore
```
        if (response.error) {
            if (typeof response.error.data === 'object') {
                if (typeof response.error.data.error_message === 'string' && typeof response.error.data.error_type === 'string') {
                    // if error data has error_message and error_type properties, we consider that node returned an error in the old format
                    throw new TypedError(response.error.data.error_message, response.error.data.error_type);
                } else {
                    throw parseRpcError(response.error.data);
                }
            } else {
                const errorMessage = `[${response.error.code}] ${response.error.message}: ${response.error.data}`;
                if (errorMessage === '[-32000] Server error: send_tx_commit has timed out.') {
                    throw new TypedError('send_tx_commit has timed out.', 'TimeoutError');
                } else {
                    throw new TypedError(errorMessage);
                }
            }
```
going to else.
I agree this seems a missing one in nearcore :thinking:

vgrichina: @ailisp looks like this case needs to be checked in `parseRpcError` then

vgrichina: ah wait, check happens before `parseRpcError` even called.

I think this check `if (typeof response.error.data === 'object') {` can instead be checking just for presence of `response.error.data`, not need to make sure it's an object

vgrichina:
https://github.com/near/near-api-js/pull/350#issuecomment-649233749

PR #279, opened 2020-04-02
Add more documentation and clean up code

willemneal:

PR #281, opened 2020-04-03
add Typedoc where needed

mikedotexe:
This is branched from https://github.com/near/near-api-js/tree/add_type_docs


mikedotexe:
@willemneal was able to change the base for this PR like we talked about. Turns out you can just edit it in Github 👍

vgrichina:
@mikedotexe see https://github.com/near/near-api-js/pull/279#issuecomment-608670940. I think you are setting up yourself for life of pain by not merging this in smaller pieces to `master`

chadoh:
In `package.json`:

```js
    "doc": "typedoc --out docs/nearlib --theme docusaurus src/ --hideBreadcrumbs",
```

Can we change this to output to `docs/near-api-js` while we're here? Or maybe just `docs`? (Is there any reason to nest it?)

mikedotexe:
> docs/near-api-js

I'll go with `docs/near-api-js` as I don't have a preference at the moment. Thanks for noticing!

mikedotexe:
okay @chadoh ready for re-review. I feel like you're the best authority on this now as you know what we need to do for the automation. Thanks!

chadoh:
@mikedotexe here's the docs site update: https://github.com/nearprotocol/docs/pull/291

Per the description there, let's use this PR as an opportunity to stop storing the output of `yarn doc` in git.

src/account_creator.ts
```diff
@@ -5,7 +5,7 @@ import { fetchJson } from './utils/web';
 import { PublicKey } from './utils/key_pair';
 
 /**
- * Account creator provides interface to specific implementation to acutally create account.
+ * Account creator provides an interface for implementations to actually create accounts.
```
chadoh, line 8: Let's drop the period while we're here

src/wallet-account.ts
```diff
@@ -87,6 +87,11 @@ export class WalletConnection {
         window.location.assign(newUrl.toString());
     }
```
chadoh, line 88: While we're here, can we fix `requestSignIn` and mark `successUrl` and `failureUrl` as optional? In many of our examples we skip these arguments, but if you convert a project to typescript you get yelled at for not providing them.

src/account_creator.ts
bowenwang1996: Nice. Love the grammar fix.

src/wallet-account.ts
mikedotexe: I put those two arguments as optional and ran `env NODE_ENV=ci yarn test` and everything passed.
Cleaned up the typedocs of the `requestSignIn` method, good eye.

src/account_creator.ts
mikedotexe: 🤙

mikedotexe:
> See two small nits below. I'm marking as "Request Changes" because:
> 
> 1. even though the typedocOptions _say_ to put the output in `docs/near-api-js`, it looks like they're still going to `docs/nearlib` – is this configuration option correct?
> 2. once we make sure the docs actually get built to the desired location, we need to add that location to `.gitignore` and delete the folder from source control, as mentioned in my comment above

Hey @chadoh I'm not seeing the issue where the generated docs are getting output to `docs/nearlib`, here's a screenshot of what I see, which is working as expected:

Screenshot 2020-04-08 17 45 16

I added `docs/` to gitignore and `git rm -r --cached docs/` to remove revision of that directory.
This PR is ready for re-review. Thank you for the comments!

PR #283, opened 2020-04-08
Setup separate entry point for browser build

vgrichina:
This is needed to prevent web bundlers from failing with default config because `fs` module is not found.

### DIFF
--- a/browser-exports.js
+++ b/browser-exports.js
@@ -1,3 +1,3 @@
 require('error-polyfill');
-window.nearApi = require('./lib/index');
+window.nearApi = require('./lib/browser-index');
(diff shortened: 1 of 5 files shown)


vgrichina:
@chadoh ping

vgrichina:
@chadoh some bundlers fail immediately (not in runtime), people hit this (in Telegram I think, there is no convenient way to link there).

There is no universal way to solve this in runtime in a way which would work with bundlers where we don't control config.

vgrichina:
this is good example of similar issue:
https://github.com/nuxt-community/dotenv-module/issues/11#issuecomment-376780588

PR #285, opened 2020-04-08
Change test setup to use account helper if available

vgrichina:


mikedotexe:
The CI is getting:

>Sender "test1586329977365474" does not have enough balance 100000000000000000 for operation costing 10000000000000000000000000

mikedotexe:
CI is failing because:
>6:5  error  'testAccount' is assigned a value but never used  no-unused-vars

.travis.yml
```diff
@@ -3,6 +3,8 @@ node_js: 10
 env:
   - NODE_ENV=ci
   - NODE_ENV=ci-betanet
+    #  - NODE_ENV=devnet
+  - NODE_ENV=betanet
```
bowenwang1996, line 7: Not sure whether this is a good idea. If betanet is down, near-api-js development would be blocked.

mikedotexe: Agree with Bowen here, I think it's too early have to have CI check betanet. In time, perhaps

PR #287, opened 2020-04-11
Hotfix/fix dist nearlib ref

mikedotexe:
Changed a bunch of references from `nearlib`
The main thing that's a change is the cdn links will not point to `…/nearlib.min.js`
Also, the keystore prefix will not be `nearlib:…`

Besides that, generally trying to reduce confusion as we move to the new name.


vgrichina:
Closing for https://github.com/near/near-api-js/pull/288

PR #288, opened 2020-04-13
Fix remaining nearlib rename quirks

vgrichina:
Cherry-picked some stuff from https://github.com/near/near-api-js/pull/287 and changed some other (e.g. using `nearApi` instead of `nearAPIJs`).

Local storage prefix cannot be changed this easily (we would effectively nuke existing accounts in wallet).

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   "license": "MIT",
   "scripts": {
     "dist": "yarn browserify && yarn doc",
-    "browserify": "browserify browser-exports.js -i node-fetch -i http -i https -o dist/nearlib.js && browserify browser-exports.js -i node-fetch -g uglifyify -o dist/nearlib.min.js",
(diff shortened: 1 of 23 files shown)


package.json
```diff
@@ -66,7 +66,7 @@
   "bundlewatch": {
     "files": [
       {
-        "path": "dist/nearlib.min.js",
+        "path": "dist/near-api-js.min.js",
```
mikedotexe, line 69: I noticed the `/dist` folder still has nearlib in it. I think we should run `yarn dist` and remove those, right?
https://github.com/near/near-api-js/tree/fix-nearlib-rename/dist

vgrichina: @mikedotexe this is best not done in PRs (results in merge conflicts). Will need rebuild after merge anyway.

mikedotexe: okay, so typically these are pushed to master, then?

vgrichina:
>  I'd like to eventually see the keystore prefixes renamed

@mikedotexe IMO it's not really worth effort by itself (needs data migration and tests for it). IMO makes sense to do when data migration is needed for other reason.

PR #290, opened 2020-04-15
Add test and fix the mishandling of function call with deposit attached

vgrichina:
Context: submitting premium message stopped working in guest book example after testnet update. No idea why it worked before (shouldn't have worked).

### DIFF
--- a/src/wallet-account.ts
+++ b/src/wallet-account.ts
@@ -235,7 +235,10 @@ class ConnectedWalletAccount extends Account {
                     return false;
                 }
                 const [{ functionCall }] = actions;
-                return functionCall && (allowedMethods.length === 0 || allowedMethods.includes(functionCall.methodName));
+                return functionCall &&
+                    functionCall.deposit.toString() === "0" && // TODO: Should support charging amount smaller than allowance?
+                    (allowedMethods.length === 0 || allowedMethods.includes(functionCall.methodName));
+                // TODO: Handle cases when allowance doesn't have enough to pay for gas
             }
         }
         // TODO: Support other permissions than FunctionCall
--- a/test/wallet-account.test.js
+++ b/test/wallet-account.test.js
@@ -106,46 +106,8 @@ it('can request transaction signing', async () => {
     });
 });
 
-it('requests transaction signing automatically when there is no local key', async () => {
-    // TODO: Refactor copy-pasted common setup code
-    let keyPair = nearApi.KeyPair.fromRandom('ed25519');
-    walletConnection._authData = {
-        allKeys: [ 'no_such_access_key', keyPair.publicKey.toString() ],
-        accountId: 'signer.near'
-    };
-    nearFake.connection.provider = {
-        query(path, data) {
-            if (path === 'account/signer.near') {
-                return { };
-            }
-            if (path === 'access_key/signer.near') {
-                return { keys: [{
-                    access_key: {
-                        nonce: 1,
-                        permission: 'FullAccess'
-                    },
-                    public_key: keyPair.publicKey.toString()
-                }] };
-            }
-            fail(`Unexpected query: ${path} ${data}`);
-        },
-        status() {
-            return {
-                sync_info: {
-                    latest_block_hash: BLOCK_HASH
-                }
-            };
-        }
-    };
-
-    try {
-        await walletConnection.account().signAndSendTransaction('receiver.near', [nearApi.transactions.transfer(1)]);
-        fail('expected to throw');
-    } catch (e) {
-        expect(e.message).toEqual('Failed to redirect to sign transaction');
-    }
-
-    const parsedUrl = url.parse(newUrl, true);
+function parseTransactionsFromUrl(urlToParse) {
+    const parsedUrl = url.parse(urlToParse, true);
     expect(parsedUrl).toMatchObject({
         protocol: 'http:',
         host: 'example.com',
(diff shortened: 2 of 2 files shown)


vgrichina:
@chadoh I'm going to merge this ASAP and cut new release as this kinda affects production. Any comments will likely be addressed in follow-up PRs.

test/wallet-account.test.js
```diff
@@ -106,11 +109,26 @@ it('can request transaction signing', async () => {
     });
 });
 
-it('requests transaction signing automatically when there is no local key', async () => {
-    // TODO: Refactor copy-pasted common setup code
-    let keyPair = nearApi.KeyPair.fromRandom('ed25519');
+function parseTransactionsFromUrl(urlToParse) {
+    const parsedUrl = url.parse(urlToParse, true);
+    expect(parsedUrl).toMatchObject({
```
chadoh, line 114: I'm never sure if it makes sense to put expectations inside helper functions like this – seems like a side-effect in an otherwise pure function. It's probably fine for now, but I'm curious if you have the same misgivings, @vgrichina

test/wallet-account.test.js
```diff
@@ -106,11 +109,26 @@ it('can request transaction signing', async () => {
...
+            nearApi.transactions.SCHEMA,
+            nearApi.transactions.Transaction,
+            Buffer.from(txBase64, 'base64')));
+    return transactions;
+}
+
+function setupWalletConnectionForSigning({ allKeys, accountAccessKeys }) {
     walletConnection._authData = {
-        allKeys: [ 'no_such_access_key', keyPair.publicKey.toString() ],
+        allKeys: allKeys,
```
chadoh, line 131: nit: can shorten to `allKeys,`

test/wallet-account.test.js
```diff
@@ -106,11 +106,26 @@ it('can request transaction signing', async () => {
...
-it('requests transaction signing automatically when there is no local key', async () => {
-    // TODO: Refactor copy-pasted common setup code
-    let keyPair = nearApi.KeyPair.fromRandom('ed25519');
+function parseTransactionsFromUrl(urlToParse) {
+    const parsedUrl = url.parse(urlToParse, true);
+    expect(parsedUrl).toMatchObject({
+        protocol: 'http:',
+        host: 'example.com',
+        query: {
+            callbackUrl: 'http://example.com/location'
```
chadoh, line 118: At some point I think it would be nice to move these strings to constants, so it's more clear  that they're coming from stubbing code

test/wallet-account.test.js
```diff
@@ -185,3 +202,69 @@ it('requests transaction signing automatically when there is no local key', asyn
...
+    keyStore.setKey('networkId', 'signer.near', localKeyPair);
+
+    try {
+        await walletConnection.account().signAndSendTransaction('receiver.near', [
+            nearApi.transactions.functionCall('someMethod', new Uint8Array(), new BN('1'), new BN('1'))
+        ]);
+        fail('expected to throw');
+    } catch (e) {
+        expect(e.message).toEqual('Failed to redirect to sign transaction');
+    }
```
chadoh, line 240: I don't mind the `try / catch`, but it may clean it up a bit to use `expect(() => { ... }).toThrow('Failed to redirect...')` 

See https://jestjs.io/docs/en/expect#tothrowerror

test/wallet-account.test.js
```diff
@@ -185,3 +202,69 @@ it('requests transaction signing automatically when there is no local key', asyn
...
+                        receiver_id: 'receiver.near',
+                        method_names: []
+                    }
+                }
+            },
+            public_key: localKeyPair.publicKey.toString()
+        }, {
+            access_key: {
+                nonce: 1,
+                permission: 'FullAccess'
```
chadoh, line 226: Is it an important part of what you're testing that this test sets this `FullAccess` key and the next test does not?

test/wallet-account.test.js
```diff
@@ -154,7 +154,7 @@ function setupWalletConnectionForSigning({ allKeys, accountAccessKeys }) {
             return {
                 transaction_outcome: { outcome: { logs: [] } },
                 receipts_outcome: []
-            }
+            };
```
chadoh, line 157: Thanks for keeping the refactor and the new test+functionality in separate commits! I'm curious how you feel about rebasing. My inclination would have been to rebase these lint fixes into the previous commits, rather than add a third commit. Do you have strong feelings for or against this sort of rebasing?

src/wallet-account.ts
```diff
@@ -235,7 +235,10 @@ class ConnectedWalletAccount extends Account {
                     return false;
                 }
                 const [{ functionCall }] = actions;
-                return functionCall && (allowedMethods.length === 0 || allowedMethods.includes(functionCall.methodName));
+                return functionCall &&
+                    functionCall.deposit.toString() === "0" && // TODO: Should support charging amount smaller than allowance?
```
chadoh, line 239: This now expects `deposit` to be defined. I don't think we required that before, and now any contract calls that only passed one or two arguments are going to start breaking. Is that accurate, @vgrichina ?

test/wallet-account.test.js
vgrichina: Don't see a problem with that. Asserting invariants is common pattern even for non-test code.

test/wallet-account.test.js
vgrichina: >  Do you have strong feelings for or against this sort of rebasing?

I don't think there is much reason to care, I usually rebase in such cases:
1) I'm actually rebasing on upstream branch that did change.
2) commits are total mess because of experimentation, i.e. history is full of stuff like "TMP:...", "WIP: ", "try this thing", etc

rebasing lint fixes is usually PITA as they often apply to multiple previous commits

test/wallet-account.test.js
vgrichina: I don't like this pattern as it is limits you on what you can assert about error (e.g. once we have codes in every error – looks like it won't be convenient to verify both message and code).

src/wallet-account.ts
vgrichina: Thanks for catching this, fixed in follow-up!

PR #295, opened 2020-04-22
improve usability of contract functions

chadoh:
This lays groundwork to improve the ergonomics of https://github.com/near-examples/guest-book/pull/79

Function Binding
----------------

When passing around contract functions throughout an app, I have ended
up in situations where `near-api-js` ends up throwing an error:

    TypeError: Cannot read property 'account' of undefined

Using arrow functions ensures correct binding of these functions, so
that `this` refers to what we expect.

`.name` behavior
----------------

If you call `someFunction.name`, JavaScript will show the name of the
function (see [mdn]). However, this doesn't work as expected with
functions defined using `Object.defineProperty`.

Using an approach found on [StackOverflow], this ensures that the
`.name` property works as expected for these contract functions.
Example:

    const contract = await new Contract(account, contractName, {
      viewMethods: ['getMessages']
    })
    console.log(conract.getMessages.name) // => 'getMessages'

  [mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name
  [StackOverflow]: https://stackoverflow.com/a/41854075/249801

### DIFF
--- a/src/contract.ts
+++ b/src/contract.ts
@@ -3,6 +3,15 @@ import { Account } from './account';
 import { getTransactionLastResult } from './providers';
 import { PositionalArgsError, ArgumentTypeError } from './utils/errors';
 
+// Makes `function.name` return given name
+function nameFunction(name, body) {
+    return {
+        [name](...args) {
+            return body(...args);
+        }
+    }[name];
+}
+
 /**
  * Defines a smart contract on NEAR including the mutable and non-mutable methods
  */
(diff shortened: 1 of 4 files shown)


chadoh:
@willemneal @vgrichina it looks like we do not currently have any tests for `Contract`. Should I add one and test this new behavior?

vgrichina:
> @willemneal @vgrichina it looks like we do not currently have any tests for `Contract`. Should I add one and test this new behavior?

This code is exercised indirectly in account.test.js (and probably other places). Feel free to add more specific tests!

Note that this change seems to be breaking some of existing tests.

chadoh:
@vgrichina I added my own test file and I think it's working. The tests pass locally and in the "Pull Request" build. However, the "Branch" build fails consistently for the `ci-betanet` environment – I think this might be a configuration issue, but haven't yet tracked it down. Are these failures expected right now?

src/contract.ts
```diff
@@ -18,26 +27,26 @@ export class Contract {
...
             });
         });
         changeMethods.forEach((methodName) => {
             Object.defineProperty(this, methodName, {
                 writable: false,
                 enumerable: true,
-                value: async function(args: any, gas?: BN, amount?: BN) {
-                    if (arguments.length > 3) {
+                value: nameFunction(methodName, async (args: object = {}, gas?: BN, amount?: BN) => {
+                    if (Object.prototype.toString.call(args) !== '[object Object]') {
```
vgrichina, line 43: it's still useful to check for argument count as well in case first argument of contract call is also object (though neither guarantees absence of false negatives)

chadoh: I added a test for this, and I moved related tests from `test/account.test.js` to `test/contract.test.js`

PR #300, opened 2020-04-29
Validator tools

ilblackdragon:
This is coming from https://github.com/nearprotocol/near-shell/pull/328:
 - getting validators via provider
 - finding seat threshold
 - diffing validators between current and previous epochs

Next up, emulating auction to identify which proposals are making into the next next epoch.


src/validators.ts
```diff
@@ -0,0 +1,72 @@
...
+ *  Calculation follow the spec: https://nomicon.io/Economics/README.html#validator-selection
+ * @params validators: current or next epoch validators.
+ * @params numSeats: number of seats.
+ */
+export function findSeatPrice(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], numSeats: number): BN {
+    const stakes = validators.map(v => new BN(v.stake, 10)).sort((a, b) => a.cmp(b));
+    const num = new BN(numSeats);
+    const stakesSum = stakes.reduce((a, b) => a.add(b));
+    if (stakesSum.lt(num)) {
+        throw 'Stakes are below seats';
```
vgrichina, line 16: nit: `throw new Error('Stakes are below seats');`

throwing strings directly has some problems (e.g. with stack traces)

src/validators.ts
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+        if (!found) {
+            right = mid;
+        }
+    }
+    return left;
+}
+
+export class ChangedValidatorInfo {
```
vgrichina, line 39: feel like `ChangedValidatorInfo` and `EpochValidatorsDiff` should just be interfaces

src/validators.ts
```diff
@@ -0,0 +1,72 @@
...
+    public current: CurrentEpochValidatorInfo;
+    public next: NextEpochValidatorInfo;
+
+    public constructor(init?: Partial<ChangedValidatorInfo>) {
```
vgrichina, line 43: What is the intent of this construction? Seems like it's basically violating type that you just defined: `Partial<T>` allows nullable fields, `Object.assign` would result in some fields not being set, but this type pretends to have non-nullable fields.

src/validators.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+/** Diff validators between current and next epoch.
+ * Returns additions, subtractions and changes to validator set.
+ * @params 
+ */
+export function diffEpochValidators(currentValidators: CurrentEpochValidatorInfo[], nextValidators: NextEpochValidatorInfo[]): EpochValidatorsDiff {
+    const validatorsMap = new Map<string, CurrentEpochValidatorInfo>();
+    currentValidators.forEach(v => validatorsMap[v.account_id] = v);
```
vgrichina, line 64: I think `[]` operator on `Map` is just using it as plain JS object. Like you need to use `.put` and `.get` if using it as Map (JS is weird). Not like it matters much for string keys, but then you can just use `{}` instead of `new Map()`.

also you can create `Map` based on key value pairs like this:

```
const validatorsMap = new Map(currentValidators.map(v => [v.account_id, v]));
```

src/validators.ts
```diff
@@ -0,0 +1,72 @@
...
+    const nextValidatorsSet = new Set(nextValidators.map(v => v.account_id));
+    return new EpochValidatorsDiff({ 
+        newValidators: nextValidators.filter(v => !(v.account_id in validatorsMap)),
```
vgrichina, line 67: `in` is also operating on object keys, not on `Map` keys.

src/validators.ts
ilblackdragon: I don't understand how to construct an object from an interface though...

Issue #298, opened 2020-04-28
Expose state stake and account balance breakdown


kcole16:
### Overview
We currently do not expose an account's complete balance breakdown, specifically the "state stake". This is necessary to compute the account's spendable balance. 

Both the [Wallet](https://github.com/nearprotocol/near-wallet/issues/481) and Explorer need to retrieve and display the state stake amount, and it's likely app devs will want the same data (so they can determine if users have a sufficient NEAR balance for a TX).

`account/<account_id>` RPC example
```bash
{
  "result": {
    "amount": "10000000999999952626000000",
    "locked": "0",
    "storage_usage": 387
  }
}
```

The spendable balance = `amount` - `state_stake`

The state stake can be calculated from `storage_usage * storage_amount_per_byte`.  

`cost_per_byte` can be retrieved from [RPC method = EXPERIMENTAL_genesis_config](https://docs.nearprotocol.com/docs/api/rpc-experimental#genesis-config) which exposes `runtime_config.storage_amount_per_byte`

cc @Patrick1904 @vgrichina

kcole16:
One way to do this is to add a [getAccountBalance](https://github.com/nearprotocol/near-wallet/blob/c7e2dedc54026f175198e9d8c1852b158604eb98/src/utils/wallet.js#L288) function to the `Account` object: https://github.com/near/near-api-js/blob/master/src/account.ts

> I'd also change it to just return object with all different "types" of balance instead of taking `type` argument.

_Originally posted by @vgrichina in https://github.com/nearprotocol/near-wallet/pull/534#issuecomment-620295470_

vgrichina:
Note that there is generally no changes needed in near-api-js to get info from experimental endpoints.

`sendJsonRpc` can be used directly for such situations:
https://github.com/near/near-api-js/blob/db51150b98f3e55c2893a410ad8e2379c10d8b73/src/providers/json-rpc-provider.ts#L114

bowenwang1996:
@kcole16 for regular users state stake is usually quite small (something like 0.01), do you think it is still good to display that?

kcole16:
@bowenwang1996 We're planning to display as "minimum balance". The total balance will be displayed in most places by default, and spendable will be highlighted on TX and "Send Money" pages.

https://github.com/near/near-wallet/issues/481

PR #316, opened 2020-05-12
Add account balance to account state (#298)

marcinbodnar:


src/account.ts
```diff
@@ -324,4 +335,25 @@ export class Account {
         });
         return result;
     }
+
+    /**
+     * Returns calculated account balance
+     * @returns {Promise<AccountBalance>}
+     */
+    private async getAccountBalance(): Promise<AccountBalance> {
```
vgrichina, line 343: any reason to make this private method?

src/account.ts
```diff
@@ -70,6 +80,7 @@ export class Account {
      */
     async state(): Promise<AccountState> {
         await this.ready;
+        this._state.balance = await this.getAccountBalance();
```
vgrichina, line 83: let's not do this, `state()` is called in many places, better avoid slowing it down by making additional query every time

src/providers/json-rpc-provider.ts
```diff
@@ -106,6 +107,14 @@ export class JsonRpcProvider extends Provider {
         return this.sendJsonRpc('validators', [blockId]);
     }
 
+    /**
+     * Gets EXPERIMENTAL_genesis_config from RPC
+     * @returns {Promise<GenesisConfig>}
+     */
+    async genesisConfig(): Promise<GenesisConfig> {
```
vgrichina, line 114: let's call this `experimental_genesisConfig` to indicate that API isn't final and can be changed

src/account.ts
```diff
@@ -36,6 +36,16 @@ export interface AccountState {
     amount: string;
     staked: string;
     code_hash: string;
+    storage_usage: number;
+    locked: string;
+    balance: AccountBalance;
```
bowenwang1996, line 41: why do we have both `locked` and `staked` here? What does `balance` refer to? how is it different from `amount`?

marcinbodnar: > why do we have both locked and staked here?

we use `locked` to calculate total balance (`amount` + `locked`), `staked` is not used, it looks like both `staked` and `accountId` are not fetched from RPC, and can be removed. 

> What does balance refer to? how is it different from amount?

`balance` contains calculated: total balance, state staked, staked and available balance. `amount` is fetched from RPC and it's used to calculate total balance.

src/account.ts
marcinbodnar: I was planning to add it to the state to avoid calling `getAccountBalance` in Wallet separately.

src/account.ts
marcinbodnar: Good point.

marcinbodnar: Changes are done.

It might be a bit more efficient if we add optional param to `state()` that will decide if the balance should be calculated and added to `_state`. With current changes, we need to call `state()` again in `getAccountBalance` to calculate balance, so for example, in Wallet we will call `state()` then call `getAccountBalance` that will again call `state()`. What do you think?

PR #329, opened 2020-05-26
add public_key field to key file

bowenwang1996:
Currently when we write a `UnencryptedFileSystemKeyStore` to file, we do not write the public key. This can cause problems as ed25519 public key is not recoverable from its private key. Also for nearcore compatibility it would be nice if the private key field is named "secret_key". @vgrichina do we have anything that depends on the field's name?

### DIFF
--- a/test/key_stores/unencrypted_file_system_keystore.test.js
+++ b/test/key_stores/unencrypted_file_system_keystore.test.js
@@ -3,7 +3,9 @@ const rimraf  = require('util').promisify(require('rimraf'));
 
 const nearApi = require('../../lib/index');
 const UnencryptedFileSystemKeyStore = nearApi.keyStores.UnencryptedFileSystemKeyStore;
+const KeyPair = nearApi.utils.KeyPairEd25519;
 const { ensureDir } = require('../test-utils');
+const fs = require('fs');
 
 const KEYSTORE_PATH = '../test-keys';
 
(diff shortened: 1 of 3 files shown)


vgrichina:
> This can cause problems as ed25519 public key is not recoverable from its private key. 

This is not true, see https://github.com/near/near-api-js/blob/543b6fbe6998b0a52d9e6e9b2d13aa97370563ae/src/utils/key_pair.ts#L109

> do we have anything that depends on the field's name?

Obviously reading the keys previously saved with this store depend on the field name. So if you want to change it – it is also necessary to change how the keys are being loaded (preferably to support both old and new name).

bowenwang1996:
@vgrichina I see. I was wrong about the reconstructing the public key, but I still think it's better to have public key in the file so that it is easy to use if people just want to get the public key. I think for naming, we can leave it as is.

vgrichina:
> but I still think it's better to have public key in the file so that it is easy to use if people just want to get the public key.

Makes sense to me. May you also add a test here https://github.com/near/near-api-js/blob/master/test/key_stores/unencrypted_file_system_keystore.test.js?

bowenwang1996:
@vgrichina tests added. Sorry for the delay

test/key_stores/unencrypted_file_system_keystore.test.js
```diff
@@ -16,5 +18,20 @@ describe('Unencrypted file system keystore', () => {
         ctx.keyStore = new UnencryptedFileSystemKeyStore(KEYSTORE_PATH);
     });
 
+    it('test get key', async () => {
```
vgrichina, line 21: nit: I think this is more or less covered in common test case for all key stores

vgrichina:
@bowenwang1996 tests failed on CI

bowenwang1996:
@vgrichina I cannot reproduce the CI failure locally and they seem unrelated to my change.

bowenwang1996:
@vgrichina I cannot understand this magic. I just switched the order of tests and then it worked.

vgrichina:
> I cannot understand this magic. I just switched the order of tests and then it worked.

@bowenwang1996 probably some files were left by previous tests changing behavior?

PR #337, opened 2020-06-09
add rpc for light client proof

bowenwang1996:
Add `experimental_lightClientProof` to provider to support querying light client proofs, which is needed for the Eth <> NEAR bridge to work. It is hard to write a test now because the rpc is not yet supported on testnet.


src/providers/json-rpc-provider.ts
```diff
@@ -115,12 +115,20 @@ export class JsonRpcProvider extends Provider {
         return await this.sendJsonRpc('EXPERIMENTAL_genesis_config', []);
     }
 
+    /**
+     * Gets EXPERIMENTAL_light_client_proof from RPC
```
vgrichina, line 119: Would be nice to provide a bit more info on what is this about (maybe link to relevant nearcore info?).

vgrichina:
@bowenwang1996 check `startFromVersion` in `test/account.test.js` – should be possible to do the same for testing not released yet features

bowenwang1996:
@vgrichina that would be true if we have proper versioning.

vgrichina:
> @vgrichina that would be true if we have proper versioning.

Then we need to fix versioning in nearcore.  @chefsale we need to figure this out. It should be possible to gate tests to run only on newer version of nearcore and looks like according to @bowenwang1996 there is something missing on nearcore side of things.

bowenwang1996:
@vgrichina yes it is blocked partially on the devops side.

PR #345, opened 2020-06-12
feat(JsonRpcProvider): support named params

### LINKED ISSUE #338: Update to use named parameters in json-rpc-provider
See what needs to be updated: https://github.com/near/near-api-js/blob/master/src/providers/json-rpc-provider.ts#L107

Acceptance criteria:
* near-api-js updated to use [this model](https://docs.near.org/docs/api/rpc-params#using-near-final-finality)
* Design discussion headed by @chadoh

chadoh:
Fixes #338 

This updates `JsonRpcProvider.block` to pass named parameters to RPC, and allows fetching blocks by either `blockId` or, new new, `finality`

You can continue to pass a raw `blockId` to this function, but you will get a deprecation warning prompting you to call it with `block({ blockId })` or `block({ finality })` instead

Reviewers: Please also review my commit message! I plan to add [Conventional Commits](https://www.conventionalcommits.org/) tooling to this repository to auto-generate release notes, so I've targeted my commit message at users of `near-api-js` who want to learn more about this change. 

Example of deprecation warning: 

JsonRpcProvider.block(blockId) deprecated use block({ blockId }) or block({ finality }) instead test/providers.test.js:36:40

Yes, this deprecation warning is thrown by the test. I think that's a good thing, as it will remind us to remove this backwards-compatibility in the future.


src/providers/json-rpc-provider.ts
```diff
@@ -81,11 +82,18 @@ export class JsonRpcProvider extends Provider {
     /**
      * Query for block info from the RPC
      * See [docs for more info](https://docs.nearprotocol.com/docs/interaction/rpc#block)
-     * @param blockId Block hash or height
-     * @returns {Promise<BlockResult>}
```
chadoh, line 85: I've deleted these lines because we're using TypeScript & TypeDoc, so these comments provide no information that isn't available via the argument typings. Explaining the new `blockQuery` argument with English seemed harder than just using the typings, so I opted to just remove the comment.

src/providers/json-rpc-provider.ts
```diff
@@ -103,7 +111,7 @@ export class JsonRpcProvider extends Provider {
      * See [docs for more info](https://docs.nearprotocol.com/docs/interaction/rpc#validators)
      * @param blockId Block hash or height, or null for latest.
      */
-    async validators(blockId: BlockId): Promise<EpochValidatorInfo> {
+    async validators(blockId: BlockId | null): Promise<EpochValidatorInfo> {
```
chadoh, line 114: I think we should add `strictNullChecks` to our `tsconfig` to catch issues like this. @vgrichina do you have objections?

src/providers/json-rpc-provider.ts
vgrichina: agree 100% that we shouldn't have docs which merely restate types

however I think in context of this method it is interesting to understand how finality argument would work, it is less obvious that for block id

src/providers/json-rpc-provider.ts
vgrichina: Should be fine I guess, except seems like it might be quite a bit of work to convert existing codebase. 

Also I'd prefer we just make it optional `validators(blockId?: BlockId)` vs use `| null`.

chadoh: The RPC call [requires](https://docs.near.org/docs/interaction/rpc#validators) passing literal `null`:

    http post https://rpc.testnet.near.org jsonrpc=2.0 method=validators params:='[null]' id=dontcare

src/providers/json-rpc-provider.ts
chadoh: Yeah, but that is explained in the docs link

I think we should move such docs inline into these comments (which will add them to the generated near-api-js API docs), and slim down the docs.near.org docs, but that seems like a separate issue. And when we do that, that's when I imagine we will move the explanation of the `finality` argument into this comment.

PR #346, opened 2020-06-15
chore: enforce & use Conventional Commits

chadoh:
This adds some tooling to make sure we use [Conventional Commits] and
use them to their fullest

* use [commitlint] & [husky] to check commit messages & reject any that
  don't fit the Conventional Commits format
* use [standard-version] to automatically update CHANGELOG, set new
  version number, and publish a release to NPM

  [Conventional Commits]: https://www.conventionalcommits.org/
  [commitlint]: https://commitlint.js.org/
  [husky]: https://www.npmjs.com/package/husky
  [standard-version]: https://github.com/conventional-changelog/standard-version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ env:
   - NODE_ENV=ci-betanet
 cache: yarn
 script:
+  - commitlint-travis
   - yarn lint
   - yarn test
   - yarn browserify && yarn run bundlewatch
--- a/README.md
+++ b/README.md
@@ -18,11 +18,16 @@ A JavaScript/TypeScript library for development of DApps on the NEAR platform
 
 # Publish
 
-Prepare `dist` version by running:
+If you are a maintainer of `near-api-js` and you want to publish a new version to npmjs.com, there are some **prerequisites**:
 
-    yarn dist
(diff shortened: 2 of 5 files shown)


vgrichina:
I think it is better to actually run tests to confirm that given change is non-breaking. Like literally run tests for previous release on new code. Any kind of convention in commits would be very reliant on people not making mistakes, we might as well just chose versions manually.

chadoh:
> literally run tests for previous release on new code

I like that idea!

> convention in commits would be very reliant on people not making mistakes

Yes, this approach requires everyone who can commit to understand the convention and make sure it's upheld. But I agree that it would be good to augment that with automated checks.

chadoh:
@vgrichina another thing we could consider: add commit-linting/enforcement for now, but wait a couple versions before adding automation for changelogs & version numbers

vgrichina:
> add commit-linting/enforcement for now, but wait a couple versions before adding automation for changelogs & version numbers

@chadoh IMO without either tests to detect breaking changes or automation to do releases it might be more annoying than useful to have such checks. I think it makes sense to start with closed loop vs just have additional checks for their own sake.

chadoh:
I do not have time to add the suggested tooling. We can revisit this later.

PR #347, opened 2020-06-19
Promote EXPERIMENTAL_light_client_proof to non-experimental

evgenykuzyakov:
Recent change in `nearcore` promoted the RPC which broke `nearlib-tests`: https://github.com/nearprotocol/nearcore/pull/2859


vgrichina:
> Recent change in nearcore promoted the RPC which broke nearlib-tests: nearprotocol/nearcore#2859

@evgenykuzyakov why not keep experimental_ as alias as well? This doesn't look like good upgradability plan.

evgenykuzyakov:
That's what I suggested on the attached `nearcore` change. I also didn't realize that nearlib CI runs on multiple version which requires proper upgradeability. Let's keep the actual RPC endpoint to be `EXPERIMENTAL_`, but the nearlib method can be without `experimental_` prefix.

vgrichina:
> I also didn't realize that nearlib CI runs on multiple version which requires proper upgradeability. 

that is done exactly to prevent stuff like this where RPC method just removed abruptly ;)

> Let's keep the actual RPC endpoint to be EXPERIMENTAL_, but the nearlib method can be without experimental_ prefix.

Let's also keep old method as alias, but use `depd` to show deprecation warning (see https://github.com/near/near-api-js/blob/30a720cf0cb86ec41b459b3b8a8a7a99f51ad07f/src/providers/json-rpc-provider.ts#L91 for example).

Note like it's critical, but makes sense to establish proper deprecation process.

Issue #127, opened 2019-12-03
Gracefully handle RPC timeout during transaction confirmation


kcole16:
Currently, when the network is slow and RPC times out during transaction confirmation, the end result is an error message, with no indication or whether or not the transaction was successful.

See [#258 in near-wallet](https://github.com/nearprotocol/near-wallet/issues/258) for a related error.

kcole16:
@bowenwang1996 Will RPC still return the TX hash in the case of a timeout?

bowenwang1996:
Right now it doesn't but it is pretty easy to make it return the tx hash.

kcole16:
@vgrichina @janedegtiareva Thoughts on this?

vgrichina:
nearlib can returned transaction hash and signed transaction in error. Then caller can chose to retry  / monitor status of transaction depending on situation.

PR #350, opened 2020-06-25
Fix retry on Timeout error

vgrichina:
This is needed because nearcore fails to return proper structured error:
https://github.com/nearprotocol/nearcore/issues/1839

Related to https://github.com/near/near-api-js/issues/127
Fixes https://github.com/near/near-api-js/pull/280

### DIFF
--- a/lib/account.js
+++ b/lib/account.js
@@ -74,6 +74,7 @@ class Account {
      * @returns {Promise<FinalExecutionOutcome>}
      */
     async retryTxResult(txHash, accountId) {
+        console.warn(`Retrying transaction ${accountId}:${serialize_1.base_encode(txHash)} as it has timed out`);
         let result;
         let waitTime = TX_STATUS_RETRY_WAIT;
         for (let i = 0; i < TX_STATUS_RETRY_NUMBER; i++) {
--- a/lib/account.js
+++ b/lib/account.js
@@ -78,8 +78,15 @@ class Account {
         let result;
         let waitTime = TX_STATUS_RETRY_WAIT;
         for (let i = 0; i < TX_STATUS_RETRY_NUMBER; i++) {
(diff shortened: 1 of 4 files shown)


vgrichina:
@ailisp had to redo https://github.com/near/near-api-js/pull/280 as Git couldn't figure out merge/rebase after source folder rename.

.travis.yml
```diff
@@ -3,7 +3,7 @@ node_js: 10
 env:
   - NODE_ENV=ci
   - NODE_ENV=ci-betanet
-    #  - NODE_ENV=devnet
+  - NODE_ENV=devnet
   - NODE_ENV=betanet
```
bowenwang1996, line 7: Doesn't look like `betanet` and `devnet` environments are useful as tests always fail there.

vgrichina: They shouldn't fail. They only fail because of Timeout and to fix retrying on Timeout we need to figure out what to do with https://discordapp.com/channels/490367152054992913/557983089704566787/725585821515907143

mikedotexe:
devnet and betanet failing. I see there's a small debate about whether to include them. I don't think we've included those environments in any other project's CI, so don't know why we're starting now. Or else we should have a more broad discussion about changing many repo's CI

vgrichina:
> devnet and betanet failing. I see there's a small debate about whether to include them. I don't think we've included those environments in any other project's CI, so don't know why we're starting now. Or else we should have a more broad discussion about changing many repo's CI

@mikedotexe this is introduced by https://github.com/near/near-api-js/pull/285

We need to move from relying on keys to test accounts being exposed (someone can just drain account and screw up CI environments in that case).

In any case this is disabled for now because contract helper on betanet/devnet times out way too often to be practical before this particular change lands on contract helper side.

PR #351, opened 2020-06-25
hack to fix nearcore ci

bowenwang1996:
Nearcore CI is broken because after https://github.com/near/near-api-js/pull/285, there is no default master account for tests and because nearcore CI uses `local` as env, it will try to load key from `.near/validator_key.json`, which is not the right place since the unittest script uses `testdir` as the home directory. Not sure what the best way to fix it, but we can potentially add another environment `unittest`.

### DIFF
--- a/test/test-utils.js
+++ b/test/test-utils.js
@@ -15,9 +15,10 @@ async function setUpTestConnection() {
         deps: { keyStore },
     });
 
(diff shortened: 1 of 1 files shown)


bowenwang1996:
looks like it's resolved in https://github.com/nearprotocol/nearcore/pull/2885

PR #354, opened 2020-06-30
Don't serialize args in account.ts

vgrichina:


src/transaction.ts
```diff
@@ -51,8 +51,11 @@ export function deployContract(code: Uint8Array): Action {
     return new Action({ deployContract: new DeployContract({code}) });
 }
 
-export function functionCall(methodName: string, args: Uint8Array, gas: BN, deposit: BN): Action {
-    return new Action({functionCall: new FunctionCall({methodName, args, gas, deposit }) });
+export function functionCall(methodName: string, args: Uint8Array | object, gas: BN, deposit: BN): Action {
```
janedegtiareva, line 54: Do you mind adding comments to the function? Specifically a note about the possible formats for args would be very helpful. Thanks!

janedegtiareva:
Looks like there are test failures to look at

src/transaction.ts
vgrichina: Done

vgrichina:
> Looks like there are test failures to address

looks like these were just lint failures, fixed now

PR #362, opened 2020-07-21
Adding transaction id to errors returned from signAndSendTransaction

janedegtiareva:
This will include a transaction id to the error results. Note that this happens even if a transaction was not included in the chain at all (e.g. insufficient balance cases). If clients want to e.g. output an explorer link, they need to check the validity of transaction via some other means. 

Tested: 
 console.log of an error with timeout 
      TypedError: Exceeded 10 status check attempts for transaction GRPWi935e8Cm2PmPqrcrAyBuM9N13h5kHJNqhfVcY33q.
          at Account.retryTxResult (C:\Users\janed\near\near-api-js\lib\account.js:97:15)
          at Account.signAndSendTransaction (C:\Users\janed\near\near-api-js\lib\account.js:119:22)
          at Object.<anonymous> (C:\Users\janed\near\near-api-js\test\account.test.js:38:5) {
        type: 'RetriesExceeded',
        txnId: 'GRPWi935e8Cm2PmPqrcrAyBuM9N13h5kHJNqhfVcY33q'
      }

Console log of some other error:
      NotEnoughBalance [Error]: Sender test-1595357554416-24383 does not have enough balance {} for operation costing {}
          at Object.parseRpcError (C:\Users\janed\near\near-api-js\lib\utils\rpc_errors.js:38:19)
          at JsonRpcProvider.sendJsonRpc (C:\Users\janed\near\near-api-js\lib\providers\json-rpc-provider.js:146:40)
          at processTicksAndRejections (internal/process/task_queues.js:97:5)
          at Account.signAndSendTransaction (C:\Users\janed\near\near-api-js\lib\account.js:115:22)
          at Object.<anonymous> (C:\Users\janed\near\near-api-js\test\account.test.js:38:5) {
        type: 'NotEnoughBalance',
        txnId: 'GkVb7Sd48MEPvaogY3xwXtmxW2q8eAf3hrfWaehZDqMt',
        balance: '500000000000000000000000000',
        cost: '500000000004309233884375010',
        signer_id: 'test-1595357554416-24383'
      }


janedegtiareva:
Do we have a good way to test error cases right now?

src/account.ts
```diff
@@ -159,6 +159,7 @@ export class Account {
             if (error.type === 'TimeoutError') {
                 result = await this.retryTxResult(txHash, this.accountId);
             } else {
+                error.transactionHash = base_encode(txHash);
```
vgrichina, line 162: Should we move this downstream to provider? Looks like right now we won't catch cases when `this.retryTxResult` throws any other error except `RetriesExceeded`.

vgrichina: Or maybe just handle it here, just make sure this situation is covered.

src/utils/errors.ts
```diff
@@ -12,8 +12,10 @@ export class ArgumentTypeError extends Error {
 
 export class TypedError extends Error {
     type: string;
-    constructor(message?: string, type?: string) {
+    transactionHash?: string;
+    constructor(message?: string, type?: string, transactionHash?: string) {
```
vgrichina, line 16: I feel it makes sense to wrap `transactionHash` into object so that we can later add more context (like `accountId`) later easily.

So that error is basically created like:

`new TypedError(message, type, { transactionHash })`

resulting object field can be something like `data` or `context`

vgrichina:
> Do we have a good way to test error cases right now?

for this case should be possible to inject mock `provider` object that throws errors when needed?

src/account.ts
janedegtiareva: Downstream provider doesn't have txnId readily available, so I have a slight preference to keep as and cover the extra cases.

src/utils/errors.ts
janedegtiareva: done

janedegtiareva:
> > Do we have a good way to test error cases right now?
> 
> for this case should be possible to inject mock `provider` object that throws errors when needed?

The question was more like is that something we are already doing somewhere, or to be added?

PR #366, opened 2020-07-21
Add janedegtiareva to CODEOWNERS as per devx ownership splitup

janedegtiareva:


CODEOWNERS
```diff
@@ -1 +1,2 @@
 *       @vgrichina
+*       @janedegtiareva
```
chadoh, line 2: You can add these on one line,

    * @vgrichina @janedegtiareva

PR #370, opened 2020-07-22
Update how transactions are sent

janedegtiareva:


janedegtiareva:
This is superseded by https://github.com/near/near-api-js/pull/374

PR #371, opened 2020-07-23
more allowances in test

bowenwang1996:
Add more allowances to access key in test to make it work with https://github.com/nearprotocol/nearcore/pull/3028

### DIFF
--- a/test/account.access_key.test.js
+++ b/test/account.access_key.test.js
@@ -20,7 +20,7 @@ beforeEach(async () => {
 
 test('make function call using access key', async() => {
     const keyPair = nearApi.utils.KeyPair.fromRandom('ed25519');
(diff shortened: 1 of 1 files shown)


chadoh:
Are the test failures legit?

vgrichina:
@bowenwang1996 we need to change default attached gas I think as well, as we won't be giving 2 NEAR allowances in wallet (that is too much).

bowenwang1996:
@chadoh I think it is just flaky.

bowenwang1996:
@vgrichina sure. What do you want to change it to? Also that can be done in a separate PR

bowenwang1996:
I'll merge this to unblock the nearcore PR

PR #372, opened 2020-07-23
Remove mentions of devnet, since it's no longer running

janedegtiareva:
https://github.com/near/devx/issues/244

### DIFF
--- a/src/key_stores/browser_local_storage_key_store.ts
+++ b/src/key_stores/browser_local_storage_key_store.ts
@@ -15,7 +15,7 @@ export class BrowserLocalStorageKeyStore extends KeyStore {
 
     /**
      * Sets a local storage item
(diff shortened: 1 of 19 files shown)


src/key_stores/browser_local_storage_key_store.ts
```diff
@@ -15,7 +15,7 @@ export class BrowserLocalStorageKeyStore extends KeyStore {
 
     /**
      * Sets a local storage item
-     * @param networkId The targeted network. (ex. default, devnet, betanet, etc…)
+     * @param networkId The targeted network. (ex. default, testnet, betanet, etc…)
```
chadoh, line 18: `default` == `testnet` right now, so I don't think this makes sense. Here and in all these other files, I think it would be better to just delete `devnet`

PR #374, opened 2020-07-27
Cache nonce, retry on nonce errors

vgrichina:


vgrichina:
WARNING: no specific tests yet. Normal tests are technically covering the code, but best to test specific situations.

src/account.ts
```diff
@@ -144,28 +112,55 @@ export class Account {
...
-        );
+            const nonce = ++accessKey.nonce;
+            [txHash, signedTx] = await signTransaction(
+                receiverId, nonce, actions, base_decode(status.sync_info.latest_block_hash), this.connection.signer, this.accountId, this.connection.networkId
+            );
+
+            try {
+                const result = await exponentialBackoff(TX_STATUS_RETRY_WAIT, TX_STATUS_RETRY_NUMBER, TX_STATUS_RETRY_WAIT_BACKOFF, async () => {
+                    try {
+                        return await this.connection.provider.sendTransaction(signedTx);
```
bowenwang1996, line 134: If I understand this logic correctly, we should not retry that frequently, especially at the beginning of the backoff.

vgrichina: @bowenwang1996 why? We are only doing new `broadcast_tx_commit` call after previous one returned timeout (my understanding is that it takes 10 seconds).

bowenwang1996: ah I got confused.

Issue #373, opened 2020-07-24
Contract methods with deposits fail if user doesn't have full access key for account


mattlockyer:
***Describe the bug***
 With the addition of 2FA we have accounts that ONLY have limited access keys.
 In an app this makes signing of contract changeMethods with deposits impossible (requires full access key).
 Scenario:
 You approve an access key to call methods of a contract. All good. Now the app wants to call a method with a deposit. Requires a full access key on the account. `near-api-js` currently throws an error in the app that it doesn't have any access keys that can sign this type of transaction.
 However, we could prepare a multisig request in the wallet that would allow the user to call the method with the deposit attached.

***To Reproduce***
 1\. create an account and enable 2FA using this render instance: [https://near-wallet-pr-700.onrender.com/](https://near-wallet-pr-700.onrender.com/)
 2\. create-near-app some sample app
 3\. change the config.js to point to this wallet instance: [https://near-wallet-pr-700.onrender.com/](https://near-wallet-pr-700.onrender.com/)
 4\. sign in, complete 2FA flow
 5\. try to change the greeting with the deposit amount "1" (the default)
 6\. see exception in app, `"Cannot find matching key for transaction sent to <span class="error">[contractName]</span>"`

This is because your account only has limited access keys to your multisig contract AND the current app. Meaning that no method calls with attached deposits are possible. However, creating a multisig request for a function call with an attached deposit and approving that IS POSSIBLE.

***Expected behavior***
 We should not throw an exception and instead continue to redirect the user to `/sign?tx=...` for their wallet.

***Additional context***
 [https://github.com/near/near-api-js/blob/3993ebe9142e8d15e55eb54f94d8c852df669722/src/wallet-account.ts#L185](https://github.com/near/near-api-js/blob/3993ebe9142e8d15e55eb54f94d8c852df669722/src/wallet-account.ts#L185)

Should actually still redirect to wallet.

This is what @kcole16 and myself came up with as lowest hanging fruit for now given wallet with 2FA is launching soon and these transactions could be important for staking, transferring Near in apps.

@vgrichina had some ideas on how to override Account / signAndSendTransaction, curious if there's a better solution?

mattlockyer:
WIP branch here: https://github.com/near/near-api-js/blob/43022551758e9a4b934b0156f5058dd36644bebe/src/wallet-account.ts#L115-L128

tests passed, still need to conduct e2e testing, ETA EOD 07/25 or will have to be 07/28

[...]

PR #379, opened 2020-08-03
No Full Access Key but Has 2FA Key

mattlockyer:
This PR allows near-api-js to redirect to the wallet if the user has 2FA key.
Previously, if the user had no function call key or full access key an error would be thrown in the app using near-api-js.
See issue here: https://github.com/near/near-api-js/issues/373

I verified with a local dist that it does what we expect it to do from the wallet side.

Added 2 new tests demonstrating this functionality:
1. working correctly with the right key
2. not working with the wrong key

Hoping to merge soon, let me know if there is anything additional required.


test/wallet-account.test.js
```diff
@@ -243,6 +243,71 @@ it('requests transaction signing automatically when function call has attached d
...
+    try {
+        res = await walletConnection.account().signAndSendTransaction('receiver.near', [
+            nearApi.transactions.functionCall('someMethod', new Uint8Array(), new BN('1'), new BN('1'))
+        ]);
+    } catch (e) {
+        console.log(e.message);
+    }
+
+    expect(res).toHaveProperty('transaction_outcome');
+    expect(res).toHaveProperty('receipts_outcome');
```
janedegtiareva, line 277: It would be helpful to give an example of expected result here. If it's hard to compare to a specific object in an assert, maybe leave a comment?

test/wallet-account.test.js
```diff
@@ -243,6 +243,71 @@ it('requests transaction signing automatically when function call has attached d
...
+    setupWalletConnectionForSigning({
+        allKeys: [ walletKeyPair.publicKey.toString() ],
+        accountAccessKeys: [{
+            access_key: {
+                nonce: 1,
+                permission: {
+                    FunctionCall: {
+                        allowance: '1000000000',
+                        receiver_id: 'signer.near',
+                        method_names: ['add_request_and_confirm']
```
janedegtiareva, line 258: for readability: it would be helpful to pull the string out into a constant in src/wallet-account.ts and reference that constant here, so it's clear that the method name has to be something specific.

src/wallet-account.ts
```diff
@@ -230,6 +230,10 @@ class ConnectedWalletAccount extends Account {
 
         if (permission.FunctionCall) {
             const { receiver_id: allowedReceiverId, method_names: allowedMethods } = permission.FunctionCall;
+            // check if access_key has 2fa method
+            if (allowedReceiverId === this.accountId && allowedMethods.includes('add_request_and_confirm')) {
```
janedegtiareva, line 234: This may be a question to clarify in a tutorial/how to doc: does the contract have to implement add_request_and_confirm?

test/wallet-account.test.js
vgrichina: I think might be better to keep it in tests (even if extracting into constant), as this serves similar purpose as double entry accounting (avoid accidental rename in this case).

test/wallet-account.test.js
```diff
@@ -243,6 +243,71 @@ it('requests transaction signing automatically when function call has attached d
...
+    });
+    keyStore.setKey('networkId', 'signer.near', localKeyPair);
+
+    let res;
+    try {
+        res = await walletConnection.account().signAndSendTransaction('receiver.near', [
+            nearApi.transactions.functionCall('someMethod', new Uint8Array(), new BN('1'), new BN('1'))
+        ]);
+    } catch (e) {
+        console.log(e.message);
```
vgrichina, line 273: why swallow error?

src/wallet-account.ts
```diff
@@ -230,6 +230,10 @@ class ConnectedWalletAccount extends Account {
 
         if (permission.FunctionCall) {
             const { receiver_id: allowedReceiverId, method_names: allowedMethods } = permission.FunctionCall;
+            // check if access_key has 2fa method
```
vgrichina, line 233: I think it's worth adding more expanded comment / linking to multisig contract code.

test/wallet-account.test.js
```diff
@@ -243,6 +243,71 @@ it('requests transaction signing automatically when function call has attached d
...
+                        allowance: '1000000000',
+                        receiver_id: 'signer.near',
+                        method_names: ['add_request_and_confirm']
+                    }
+                }
+            },
+            public_key: localKeyPair.publicKey.toString()
+        }]
+    });
+    keyStore.setKey('networkId', 'signer.near', localKeyPair);
```
vgrichina, line 265: missing `await`

test/wallet-account.test.js
```diff
@@ -243,6 +243,71 @@ it('requests transaction signing automatically when function call has attached d
...
+                        allowance: '1000000000',
+                        receiver_id: 'signer.near',
+                        method_names: ['not_a_valid_2fa_method']
+                    }
+                }
+            },
+            public_key: localKeyPair.publicKey.toString()
+        }]
+    });
+    keyStore.setKey('networkId', 'signer.near', localKeyPair);
```
vgrichina, line 299: missing `await`

src/wallet-account.ts
mattlockyer: expanded comment

test/wallet-account.test.js
mattlockyer: ok

mattlockyer:
@janedegtiareva @vgrichina thanks for the review, edits are made. Let me know if you have further questions.

test/wallet-account.test.js
```diff
@@ -2,6 +2,8 @@ const url = require('url');
 const localStorage = require('localstorage-memory');
 const BN = require('bn.js');
 
+const MULTISIG_HAS_METHOD = 'add_request_and_confirm';
```
janedegtiareva, line 5: It would be helpful to understand where the value is coming from: 
maybe reference the const from wallet-account.ts or leave a comment that this is where this value is defined?

mattlockyer: added similar comment from wallet-account.ts to this test

PR #386, opened 2020-08-16
Record current directory in file system keystore

ilblackdragon:
If application changes the path, keystore stops to work as it recorded a relative path.
This updates to record absolute path at the creation and leverage that during the app lifetime.

### DIFF
--- a/src/key_stores/unencrypted_file_system_keystore.ts
+++ b/src/key_stores/unencrypted_file_system_keystore.ts
@@ -58,12 +58,7 @@ export class UnencryptedFileSystemKeyStore extends KeyStore {
 
     constructor(keyDir: string) {
         super();
-        // Check if path is absolute by resolving it and comparing to normailzed path.
-        if (path.resolve(keyDir) !== path.normalize(keyDir)) {
-            // If path is not absolute, add current working directory.
-            keyDir = path.join(process.cwd(), keyDir);
-        }
-        this.keyDir = keyDir;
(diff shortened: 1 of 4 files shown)


src/key_stores/unencrypted_file_system_keystore.ts
```diff
@@ -57,6 +58,11 @@ export class UnencryptedFileSystemKeyStore extends KeyStore {
 
     constructor(keyDir: string) {
         super();
+        // Check if path is absolute by resolving it and comparing to normailzed path.
+        if (path.resolve(keyDir) !== path.normalize(keyDir)) {
```
vgrichina, line 62: there is https://nodejs.org/api/path.html#path_path_isabsolute_path

src/key_stores/unencrypted_file_system_keystore.ts
```diff
@@ -57,6 +58,11 @@ export class UnencryptedFileSystemKeyStore extends KeyStore {
...
+            // If path is not absolute, add current working directory.
+            keyDir = path.join(process.cwd(), keyDir);
```
vgrichina, line 64: why not just use `path.resolve`?  You can shorten whole thing to:

```
this.keyDir = path.resolve(process.cwd(), keyDir);
```

it will automatically do the right thing for `keyDir` being absolute:
https://nodejs.org/api/path.html#path_path_resolve_paths

Issue #331, opened 2020-05-26
Don't rely on classes for error handling


vgrichina:
Currently structured errors are represented by custom TypeScript classes for every error type. This has few drawbacks:
- checking if smth is instance of type is unreliable in JS. It breaks down easily once there are multiple versions of library in given environment, similar to problems with serialization (see https://github.com/near/near-api-js/issues/330)
- a lot of boilerplate (including some code gen) is necessary to support this
- every time nearcore introduces new error types – there is a need to do codegen, make new near-api-js release, etc

Given dynamic nature of JS we would be better of just doing generic `TypedError` with following fields:
- `type` – error type string. Used to check which type of error it is.
- `data` – whatever detailed data JSON object attached to error. Used as parameters when formatting errors for UI, etc

PR #388, opened 2020-08-18
Propagate tx info into ServerErrors

ilblackdragon:


src/utils/rpc_errors.ts
```diff
@@ -57,6 +69,8 @@ function walkSubtype(errorObj, schema, result, typeName) {
         }
         return walkSubtype(error, schema, result, errorTypeName);
     } else {
+        // TODO: is this the right thing to do?
```
vgrichina, line 72: unfortunately I don't know, as I have no understanding of what are possible responses from nearcore side. 

also I think we should completely remove everything related to error schema https://github.com/near/near-api-js/issues/331

mikedotexe:
@vgrichina this repo is important in the dependency chain for the EVM tooling. Can we prioritize getting this looked at this week, please?

src/utils/rpc_errors.ts
mikedotexe: @bowenwang1996 I feel like you may know a bit about the nearcore errors. Would you mind giving us your two cents when you get a moment this week?

bowenwang1996: an error only has `kind` if it is an action error but there are many other errors out there

mikedotexe:
Initially I thought we simply needed the section where exports happen in the second commit here:
https://github.com/near/near-api-js/pull/388/commits/81f6403eef19dfc35d7dc0e02542c96f6ae61c28#diff-fff0ee1c855d7fdef5ec027268d27066a50d0c0dc4e4832b51f8a2e40129e177R37-R44

because without that, Balancer tests fail saying:

>ERROR:  TypeError: nearAPI.transactions.FunctionCall is not a constructor

Adding those exports alone solves that, but we do need this full PR in order to get the best results from tests.
Included here are two text files showing the Balancer test results

[9-failing-with_account-errors_branch.txt](https://github.com/near/near-api-js/files/5455277/9-failing-with_account-errors_branch.txt)
[39-failing-only-with-export.txt](https://github.com/near/near-api-js/files/5455278/39-failing-only-with-export.txt)

I'm wondering how we can get this PR moving along. I don't know a ton about errors, except that it might be a heavy lift to address the line that we're commenting about [here](https://github.com/near/near-api-js/pull/388#pullrequestreview-483012001) We're trying to keep a pretty aggressive timeline in terms of having the EVM-related PRs buttoned up ASAP. Wondering how I can help, and what the expected lift is for this. I'll tag @djsatok so we can make sure this project is tracked

mikedotexe (approved):
Ran a bunch of other checks in addition to the rather thorough tests we have here.
Nonexistent account
`./bin/near state nonexistent`
When no connection to RPC is there:
`NEAR_ENV=local ./bin/near state nonexistent`
When a wrong argument type is given:
`./bin/near view oracle.chainlink.testnet get_all_requests '{"max_num_accounts": "100", "max_requests": "abc"}'`
Calling nonexistent method:
./bin/near view oracle.chainlink.testnet ZZZget_all_requests '{"max_num_accounts": "100", "max_requests": "abc"}'

The only test that had issues was getting cross-contract failures, but finally figured it out that we needed to explicitly add the error.type and error.message and now everything is behaving perfectly.

Also, talked to Vlad about this yesterday, and while there is discussion about completely removing the class-based way that errors are handled currently, it's better to have that be a separate task.

This PR addresses EVM closely, especially with "revert" errors, so this PR represents a major improvement for the EVM and we'd like to merge it asap.

mikedotexe:
Note: we are going to hold off on merging this until we've resolved this issue:
https://github.com/near/nearcore/issues/3544

PR #389, opened 2020-08-19
Export Signature class

anbud:
Add `Signature` class from `transaction.ts` to exported classes. 

Exporting the `Signature` class allows us to construct a new `SignedTransaction` object by providing a signature from a hardware wallet (such as _Ledger Nano S_).


vgrichina:
@anbud are you planning to use this with Ledger? If so – take a look at near-shell and near-wallet, as they already have Ledger integration.

PR #396, opened 2020-09-01
docs: add walletUrl to NearConfig type

chadoh:
While not used directly within the `src/near.ts` file, `src/connection.ts` expects this value to be set on this config, and provides no default


vgrichina:
`walletUrl` isn't required if you don't use `WalletConnection`, should be optional here or will get in the way for anyone building CLI or backend.

chadoh:
Do you have any good ideas on how to add types to describe that behavior? Should we alter the type for WalletConnection?

PR #403, opened 2020-09-16
Adding AccountMultisig

mattlockyer:
This should provide a generic multisig account class that extends account.

TODO
- make class self aware it is multisig
- multisig provider e.g. contract helper methods
- e2e test with wallet


src/account_multisig.ts
```diff
@@ -0,0 +1,94 @@
...
+// helpers
+const getContract = (account) => {
+    return new MultisigContract(account, account.accountId, {
+        viewMethods: VIEW_METHODS,
+        changeMethods: METHOD_NAMES_LAK,
+    });
+};
+
+const convertPKForContract = (pk) => {
+    if (typeof pk !== 'string') {
```
vgrichina, line 60: nit: not need for this check, just always call .toString()

PR #414, opened 2020-09-29
AccountMultisig - update for near-cli support

mattlockyer:
This PR makes AccountMutlisig aware if it's running in the browser or node.

When running in node it uses:
- an inMemory cache of the current multisig request to confirm
- node-fetch library to send CH requests


mattlockyer:
Open to suggestions regarding the node-fetch package. Basically needed a quick way to communicate with CH that works in browser/node at the same time.

dist/near-api-js.js
```diff
@@ -414,12 +414,25 @@ exports.MULTISIG_DEPOSIT = new bn_js_1.default('0');
...
 ;
+// in memory request cache for node w/o localStorage
+let __multisigRequest = null;
 class AccountMultisig extends account_1.Account {
     constructor(connection, accountId) {
         super(connection, accountId);
         this.contract = getContract(this);
     }
+    // overrides
+    // async findAccessKey(receiverId, actions) {
```
vgrichina, line 427: is this some debug leftover?

src/account_multisig.ts
```diff
@@ -9,6 +9,13 @@ import { PublicKey } from './utils/key_pair';
 import { Action, addKey, deleteKey, deployContract, functionCall, functionCallAccessKey } from './transaction';
 import { FinalExecutionOutcome } from './providers';
 
+const IS_BROWSER = typeof window !== 'undefined'
+
+let fetch = IS_BROWSER && window.fetch
```
vgrichina, line 14: why is this needed? Should this use same `web.ts` helper that all other code?

src/account_multisig.ts
```diff
@@ -118,11 +129,17 @@ export class AccountMultisig extends Account {
     }
 
     getRequest() {
-        return JSON.parse(localStorage.getItem(`__multisigRequest`) || `{}`)
```
vgrichina, line 121: I think it's better to pass `storage` as optional parameter (e.g. as part of 3rd `options` parameter) when creating `AccountMultisig`. In browser user of this class can pass `localStorage` if needed.

Then you don't have to use same key in localstorage for every `AccountMultisig` instance in browser as any custom implementation can be passed.

PR #424, opened 2020-10-03
Fixing up addKey signature to support many method names

ilblackdragon:
Additionally, fixing up logic in the MultisigAccount for AddKey.

This is fixing `add-key` command in near-shell

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -305,17 +305,20 @@ export class Account {
     /**
      * @param publicKey A public key to be associated with the contract
      * @param contractId NEAR account where the contract is deployed
-     * @param methodName The method name on the contract as it is written in the contract code
+     * @param methodNames The method names on the contract that this key will have access to
      * @param amount Payment in yoctoⓃ that is sent to the contract during this function call
      * @returns {Promise<FinalExecutionOutcome>}
      * TODO: expand this API to support more options.
      */
-    async addKey(publicKey: string | PublicKey, contractId?: string, methodName?: string, amount?: BN): Promise<FinalExecutionOutcome> {
+    async addKey(publicKey: string | PublicKey, contractId?: string, methodNames?: string[], amount?: BN): Promise<FinalExecutionOutcome> {
         let accessKey;
         if (contractId === null || contractId === undefined || contractId === '') {
             accessKey = fullAccessKey();
         } else {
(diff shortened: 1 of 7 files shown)


src/account_multisig.ts
```diff
@@ -43,11 +43,16 @@ export class AccountMultisig extends Account {
         this.contract = <MultisigContract>getContract(this);
     }
 
-    async addKey(publicKey: string | PublicKey, contractId?: string, methodName?: string, amount?: BN): Promise<FinalExecutionOutcome> {
-        if (contractId) {
-            return super.addKey(publicKey, contractId, MULTISIG_CHANGE_METHODS.join(), MULTISIG_ALLOWANCE)
+    /** 
+     * If contractId matches given accountId, mutlisig only address another multisig access key.
```
vgrichina, line 47: @ilblackdragon not sure I understand the intention here. This seems to break everything, e.g. you cannot login to apps anymore. AFAIK this is perfectly possible with multisig contract (I just tried logging in into guest-book today).

Also shouldn't it even be possible to add another full access key (e.g. for seed phrase recovery)?

src/account.ts
```diff
@@ -305,17 +305,20 @@ export class Account {
...
      * @param publicKey A public key to be associated with the contract
      * @param contractId NEAR account where the contract is deployed
-     * @param methodName The method name on the contract as it is written in the contract code
+     * @param methodNames The method names on the contract that this key will have access to
      * @param amount Payment in yoctoⓃ that is sent to the contract during this function call
      * @returns {Promise<FinalExecutionOutcome>}
      * TODO: expand this API to support more options.
      */
-    async addKey(publicKey: string | PublicKey, contractId?: string, methodName?: string, amount?: BN): Promise<FinalExecutionOutcome> {
+    async addKey(publicKey: string | PublicKey, contractId?: string, methodNames?: string[], amount?: BN): Promise<FinalExecutionOutcome> {
```
vgrichina, line 313: type needs to be `string[] | string` or it's going to break users on TypeScript

I can fix this up if necessary.

vgrichina (changes requested):
Not clear what are intentions with `AccountMultisig.addKey` changes, other than that I can fix it up myself and merge.

vgrichina:
Closing as similar change is included in https://github.com/near/near-api-js/pull/453

PR #426, opened 2020-10-07
feat: allow borsh-serialized contract fn arguments

chadoh:
Let's say you have a `contract` object instantiated in your code, and the contract it wraps has an `add_person` function which receives a borsh-serialized struct. The following code now works:

    import { serialize } from 'near-api-js/lib/utils/serialize'

    class Person {
      constructor (proof) {
        Object.assign(this, proof)
      }
    }

    const personSchema = new Map([
      [Person, {
        kind: 'struct',
        fields: [
          ['age', 'u64'],
          ['name', 'String'],
        ]
      }]
    ])

    const person = new Person({ age: 33, name: 'Jesus' })

    contract.add_person(serialize(personSchema, person))

### DIFF
--- a/src/contract.ts
+++ b/src/contract.ts
@@ -12,6 +12,12 @@ function nameFunction(name, body) {
     }[name];
 }
 
+const isUint8Array = (x: any) =>
+    x && x.byteLength !== undefined && x.byteLength === x.length;
+
+const isObject = (x: any) =>
+    Object.prototype.toString.call(x) === '[object Object]';
+
 /**
  * Defines a smart contract on NEAR including the mutable and non-mutable methods
  */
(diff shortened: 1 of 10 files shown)


src/contract.ts
```diff
@@ -40,7 +47,10 @@ export class Contract {
                 writable: false,
                 enumerable: true,
                 value: nameFunction(methodName, async (args: object = {}, gas?: BN, amount?: BN, ...ignored) => {
-                    if (ignored.length || Object.prototype.toString.call(args) !== '[object Object]') {
+                    if (
```
vgrichina, line 50: nit: this formatting of `if (` almost gives me physical pain lol

vgrichina (approved):

vgrichina:
@chadoh need to fix test failures

src/contract.ts
chadoh: hahaha, ok, I'll find a way to shorten it enough to fit on one line comfortably

chadoh:
@vgrichina merged; could you cut a release?

PR #428, opened 2020-10-07
updated AccountMultisig with callbacks

mattlockyer:
No need to override signAndSendTransaction

More of the sendCode, getCode, verifyCode flow is handled here by default.

Pass callbacks to tap into what you want to change i.e. change the getCode UI/UX only:

```
constructor(connection: Connection, accountId: string, options: any) {
    super(connection, accountId);
    this.storage = options.storage;
    this.sendCode = options.sendCode || this.sendCodeDefault;
    this.getCode = options.getCode || this.getCodeDefault;
    this.onResult = options.onResult;
    this.contract = <MultisigContract>getContract(this);
}
```


src/account_multisig.ts
```diff
@@ -157,7 +216,34 @@ export class AccountMultisig extends Account {
...
-    async verifyRequestCode(securityCode: string) {
+    async getCodeDefault(): Promise<string> {
+        if (typeof window === 'undefined') {
+            throw new Error('There is no getCodeFallback for Node env. Please provide your own in AccountMultisig constructor options argument.');
+        }
+        const method = await this.get2faMethod();
+        if (!method || !method.kind) {
+            throw new Error('no active 2fa method found');
+        }
+        const code = window.prompt(`Enter security code sent to ${ method.kind.split('2fa-')[1] }`)
```
vgrichina, line 227: let's not introduce dependency on browser here. I think it's fine if default is just throwing an error if callback is not set

src/account_multisig.ts
```diff
@@ -21,13 +21,56 @@ export const MULTISIG_CHANGE_METHODS = ['add_request', 'add_request_and_confirm'
...
 export const MULTISIG_CONFIRM_METHODS = ['confirm'];
 
+/********************************
+This method can be used to mod a near instance .account helper to return AccountMultisig if the account has multisig deployed
+It takes an options object where you can provide callbacks for:
+- sendCode: how to send the 2FA code in case you don't use NEAR Contract Helper
+- getCode: how to get code from user (use this to provide custom UI/UX for prompt of 2FA code)
+- onResult: the tx result after it's been confirmed by NEAR Contract Helper
+********************************/
+export const modIfMultisig = (near, options) => {
```
vgrichina, line 31: I think it's best if it is just default behavior of `near.account`. In a sense that `await near.account(accountId)` just figures out what is proper instance for given `accountId` and returns it.

However this means that stuff like callback to get code will need to be passed through main config when calling `connect`

src/account_multisig.ts
```diff
@@ -157,7 +216,34 @@ export class AccountMultisig extends Account {
...
+        return code
+    }
+
+    async promptAndVerify() {
+        const method = await this.get2faMethod();
+        const securityCode = await this.getCode(method)
+        try {
+            const { success, res: result } = await this.verifyCode(securityCode);
+            if (!success || result === false) {
+                throw new Error('Request failed with error: ' + JSON.stringify(result));
```
vgrichina, line 237: what kind of errors won't result in `this.verifyCode` throwing exception?

src/account_multisig.ts
```diff
@@ -157,7 +216,34 @@ export class AccountMultisig extends Account {
...
+            }
+            return typeof result === 'string' && result.length === 0 ? 'true' : result;
```
vgrichina, line 239: why are we using `'true'`  string as return value?

vgrichina (commented):
See comments + needs tests

src/account_multisig.ts
mattlockyer: sorry it's cause it was going to be printed, will fix

src/account_multisig.ts
mattlockyer: I'm trying to avoid a backwards compat issue.

src/account_multisig.ts
mattlockyer: If contract panics, RPC will throw in CH and set success to false with JSON stringified result.

Sometimes a contract will return boolean false. This will be in result.

mattlockyer: maybe I shouldn't include the false condition. Since some contracts might return false, but intention is that the request was processed just returned false.

src/account_multisig.ts
vgrichina: @mattlockyer should we add option to config? `near.connect({...config, 2fa: true })` or something like that?

PR #429, opened 2020-10-09
feat: enable view calls to deserialize with borsh

chadoh:
Say you have a `contract` object instantiated in your code, and this contract has a `get_age` endpoint which returns a borsh-serialized number. The following code now works:

    import {
      deserialize as deserializeBorsh
    } from 'near-api-js/lib/utils/serialize'

    class BorshPerson {
      constructor (args) {
        Object.assign(this, args)
      }
    }

    const schema = new Map([
      [BorshPerson, {
        kind: 'struct',
        fields: [
          ['get_age', 'u64']
        ]
      }]
    ])

    function deserializePerson (raw) {
      return deserializeBorsh(schema, Person, raw)
    }

    // a small helper class to wrap a near-api-js Contract instance
    export default class Person {
      constructor (contract) {
        this.contract = contract
      }

      async getAge () {
        const deserialized = await this.contract.get_age(
          {},
          { parse: deserializePerson }
        )
        return deserialized.get_age.toNumber()
      }
    }

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -67,6 +67,10 @@ interface ReceiptLogWithFailure {
     failure: ServerError;
 }
 
+function parseJsonFromRawResponse (response: Uint8Array): any {
+    return JSON.parse(response.toString());
+}
+
 /**
  * More information on [the Account spec](https://nomicon.io/DataStructures/Account.html)
  */
@@ -112,7 +116,7 @@ export class Account {
         }
     }
 
-    private printLogs(contractId: string, logs: string[], prefix: string = '') {
+    private printLogs(contractId: string, logs: string[], prefix = '') {
         for (const log of logs) {
             console.log(`${prefix}Log [${contractId}]: ${log}`);
         }
@@ -182,7 +186,7 @@ export class Account {
             }
         });
         if (!result) {
-            throw new TypedError(`nonce retries exceeded for transaction. This usually means there are too many parallel requests with the same access key.`, 'RetriesExceeded');
+            throw new TypedError('nonce retries exceeded for transaction. This usually means there are too many parallel requests with the same access key.', 'RetriesExceeded');
         }
 
         const flatLogs = [result.transaction_outcome, ...result.receipts_outcome].reduce((acc, it) => {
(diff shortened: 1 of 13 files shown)


chadoh:
@vgrichina tests added

src/account.ts
```diff
@@ -67,6 +67,10 @@ interface ReceiptLogWithFailure {
     failure: ServerError;
 }
 
+function parseJsonFromRawResponse (response: Uint8Array): any {
+    return JSON.parse(response.toString());
```
vgrichina, line 71: nit: you cannot assume that `Uint8Array.toString()` does expected thing. Need to either use `Buffer.from` to be explicit

test/contract.test.js
```diff
@@ -56,6 +56,32 @@ const contract = new Contract(account, null, {
...
+            1,
+            'lol',
+            [],
+            new Date(),
+            null,
+            new Set(),
+        ].map(async badArgs => {
+            try {
+                await contract.viewMethod({ a: 1 }, badArgs);
+                throw new Error(`Calling \`contract.viewMethod({ a: 1 }, ${badArgs})\` worked. It shouldn't have worked.`);
```
vgrichina, line 77: nit: more idiomatic would be to use `fail`

test/contract.test.js
```diff
@@ -56,6 +56,32 @@ const contract = new Contract(account, null, {
...
 
+describe('viewMethod', () => {
+    test('passes options through to account viewFunction', async () => {
+        function customParser () {}
+        const stubbedReturnValue = await contract.viewMethod({}, { parse: customParser });
+        expect(stubbedReturnValue.options.parse).toBe(customParser);
+    });
+
+    test('throws PositionalArgsError if second argument is not an object', () => {
+        return Promise.all([
```
vgrichina, line 67: nit: seems like a use case for https://jestjs.io/docs/en/api#describeeachtablename-fn-timeout

vgrichina (approved):
lgtm except few nits

src/account.ts
chadoh: The `Buffer.from` happens in `viewFunction`:

```js
async viewFunction(
    contractId: string,
    methodName: string,
    args: any,
    { parse = parseJsonFromRawResponse } = {}
): Promise<any> {
    args = args || {};
    this.validateArgs(args);
    const result = await this.connection.provider.query(`call/${contractId}/${methodName}`, base_encode(JSON.stringify(args)));
    if (result.logs) {
        this.printLogs(contractId, result.logs);
    }
    return result.result && result.result.length > 0 && parse(Buffer.from(result.result));
}
```

Would we want to do it again here, or pass raw `result.result` to `parse`?

test/contract.test.js
chadoh: Not sure I understand. We want to await the call, and if it succeeds, we throw an error. I spent a significant amount of time trying to find a more idiomatic-to-Jest way of doing this in the similar test above, but ultimately landed back on this approach, which I stole from elsewhere in this codebase.

src/account.ts
vgrichina: I understand that it happens upstream in this case. But declared input type doesn't require it happening. So you either want to be explicit about accepting only `Buffer` or do `Buffer.from` (best option as it works in runtime)

PR #430, opened 2020-10-10
default gas missing a zero

ALuhning:
Looks like a typo on the default_func_call_gas.


chadoh:
@vgrichina looks like you purposely removed a 0 from this value in ccd872615; should we change the comment above it to match?

vgrichina:
> @vgrichina looks like you purposely removed a 0 from this value in [ccd8726](https://github.com/near/near-api-js/commit/ccd87261539a0992b5a40aa14fcfb8dccbce67b7); should we change the comment above it to match?

yes, I think we need to change a comment

PR #436, opened 2020-10-14
add support for multiple methodNames to Account.addKey

mattlockyer:
This PR adds support for `Account.addKey` that _also_ accepts an array of methodNames vs. just a string of 1 methodName. Matches more closely to how `functionCallAccessKey` works.

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -305,14 +305,14 @@ export class Account {
     /**
      * @param publicKey A public key to be associated with the contract
      * @param contractId NEAR account where the contract is deployed
-     * @param methodName The method name on the contract as it is written in the contract code
+     * @param methodNames The method names on the contract that should be allowed to be called. Pass null or '' for no method names and [] for any method names.
      * @param amount Payment in yoctoⓃ that is sent to the contract during this function call
      * @returns {Promise<FinalExecutionOutcome>}
      * TODO: expand this API to support more options.
      */
     async addKey(publicKey: string | PublicKey, contractId?: string, methodNames?: string|string[], amount?: BN): Promise<FinalExecutionOutcome> {
         if (!methodNames) {
-            methodNames = []
+            methodNames = ['']
         }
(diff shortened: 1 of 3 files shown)


vgrichina (commented):
@mattlockyer check out previous PR on the topic https://github.com/near/near-api-js/pull/424/files

I think Illia had some unanswered questions there, makes sense to check

mattlockyer:
> I think Illia had some unanswered questions there, makes sense to check
Yes I have the correct type and not trying to make it multisig specific. It's still generic.

I have the same check for undefined / null and add extra Array.isArray (fairly safe IMHO)

src/account.ts
```diff
@@ -310,12 +310,18 @@ export class Account {
      * @returns {Promise<FinalExecutionOutcome>}
      * TODO: expand this API to support more options.
      */
-    async addKey(publicKey: string | PublicKey, contractId?: string, methodName?: string, amount?: BN): Promise<FinalExecutionOutcome> {
+    async addKey(publicKey: string | PublicKey, contractId?: string, methodNames?: string|string[], amount?: BN): Promise<FinalExecutionOutcome> {
+        if (!methodNames) {
+            methodNames = []
```
vgrichina, line 315: @mattlockyer are you sure this is the same as `null` / `undefined`?  I think empty list might mean "no methods are allowed" and `null` means "any method is allowed".

mattlockyer: In the core:
Empty list means "all methods" at protocol level.
Empty string `''` means no methods.
https://nomicon.io/DataStructures/AccessKey.html#accesskeypermissionfunctioncall

It seems like we have options how to interpret this?

AFAIK we have to end up with an Array in TS/JS?
https://github.com/near/near-api-js/blob/cf003f1cadb5bfa93dcde165f94dfadd91a9f82e/src/transaction.ts#L9-L13

PR #437, opened 2020-10-15
hotfix: missing await in getRequestNonce and getRequestIds

mattlockyer:
missing await so not often requestId is not set properly because it's still receiving promise from contract call vs. value

Fixes so requestId will always be value:

https://github.com/near/near-api-js/blob/93e8a6025cb278cbdfaf97983aea514ccfa1a016/src/account_multisig.ts#L61-L62

### DIFF
--- a/src/account_multisig.ts
+++ b/src/account_multisig.ts
@@ -21,7 +21,6 @@ export const MULTISIG_CHANGE_METHODS = ['add_request', 'add_request_and_confirm'
 export const MULTISIG_VIEW_METHODS = ['get_request_nonce', 'list_request_ids'];
 export const MULTISIG_CONFIRM_METHODS = ['confirm'];
 
-
 interface MultisigContract {
     get_request_nonce(): any,
     list_request_ids(): any,
@@ -59,6 +58,9 @@ export class AccountMultisig extends Account {
         await this.deleteUnconfirmedRequests()
 
         const requestId = await this.getRequestNonce()
+
+        console.log('getRequestNonce', requestId)
+
         this.setRequest({ accountId, requestId, actions });
 
         const args = new Uint8Array(new TextEncoder().encode(JSON.stringify({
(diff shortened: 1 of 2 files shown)


src/account_multisig.ts
```diff
@@ -115,11 +115,11 @@ export class AccountMultisig extends Account {
     // helpers
 
     async getRequestNonce(): Promise<Number> {
-        return this.contract.get_request_nonce();
+        return await this.contract.get_request_nonce();
     }
 
     async getRequestIds(): Promise<string> {
-        return this.contract.list_request_ids();
+        return await this.contract.list_request_ids();
```
chadoh, line 122: I just want to note that this is the same as removing the `async` and keeping the `return` as it was

chadoh (approved):

src/account_multisig.ts
mattlockyer: I know, but I had trouble with that before... perhaps that was JS and with TypeScript it's more clear and safer with typed return values.

chadoh: Sorry, I should have made it clear that I don't much care either way; not sure which is more idiomatic tbh 😄

chadoh (approved):

vgrichina:
@mattlockyer @chadoh are you sure it's actually fixing some issue?

```
> f=() => new Promise((resolve) => setTimeout(resolve, 1000))
[Function: f]
> g=async () => f()
[AsyncFunction: g]
> (async()=> { console.log(Date.now());  await g(); console.log(Date.now()) })()
1602797869907
Promise { <pending> }
> 1602797870911

```

seems like it waits properly even when you return promise directly from `async` function (which makes sense to me)

mattlockyer:
> seems like it waits properly even when you return promise directly from `async` function (which makes sense to me)
Ya... I might be after the wrong issue.

I can't explain it, but rarely getRequestNonce doesn't return any value and what ends up getting stored for requestId is undefind or null.

Then you pass this to contract helper and because requestId is undefined it thinks you're init the 2fa method, and the user ends up verifying "adding ... as 2fa method"

It's really weird.

vgrichina:
> I can't explain it, but rarely getRequestNonce doesn't return any value and what ends up getting stored for requestId is undefind or null.

@mattlockyer  How to check that it works then? We need to understand what is going on. Maybe the issue is in something else?

PR #444, opened 2020-10-31
leave ledger key untouched when deployMutlisig

mattlockyer:
This PR filters out ledger key from keys that are converted to limited access keys, when user deploys multisig contract.


vgrichina:
@mattlockyer looks good, but please add regression test

vgrichina (changes requested):
Needs a test demonstrating fixed bug

test/test-utils.js
```diff
@@ -27,8 +27,8 @@ function generateUniqueString(prefix) {
     return `${prefix}-${Date.now()}-${Math.round(Math.random() * 1000000)}`;
 }
 
-async function createAccount(near) {
-    const newAccountName = generateUniqueString('test');
+async function createAccount(near, str = 'test') {
```
vgrichina, line 30: not sure what is the reason for the change? Does it matter for test what is account name prefix?

test/account.test.js
```diff
@@ -223,3 +224,32 @@ describe('with deploy contract', () => {
...
+            data: originalKeys.map(({ public_key: publicKey }, i) => ({ publicKey, kind: i > 0 ? 'ledger' : 'phrase' }))
+        });
+        accountMultisig.postSignedJson = () => ({
+            publicKey: KeyPair.fromRandom('ed25519').getPublicKey()
+        });
+        try {
+            await accountMultisig.deployMultisig([...fs.readFileSync('./test/data/main.wasm')]);
+        } catch (e) {
+            if (e.message.indexOf('Contract method is not found') === -1) {
+                throw(e);
```
vgrichina, line 247: nit: it's `throw e;`

test/account.test.js
```diff
@@ -223,3 +224,32 @@ describe('with deploy contract', () => {
...
+
+    test('skip access key if recovery method is "ledger"', async () => {
+        const account = await testUtils.createAccount(nearjs, 'multisig');
+        await account.addKey(KeyPair.fromRandom('ed25519').getPublicKey());
+        const originalKeys = await account.getAccessKeys();
+        const { multisig: { AccountMultisig } } = nearApi;
+        const accountMultisig = new AccountMultisig(nearjs.connection, account.accountId, {});
+        // modifiers to return the key material we want to test against
+        accountMultisig.getRecoveryMethods = () => ({
+            data: originalKeys.map(({ public_key: publicKey }, i) => ({ publicKey, kind: i > 0 ? 'ledger' : 'phrase' }))
```
vgrichina, line 238: there should be non-ledger and non-phrase key as well to test that it gets converted properly

test/account.test.js
```diff
@@ -223,3 +224,32 @@ describe('with deploy contract', () => {
...
+        });
+        try {
+            await accountMultisig.deployMultisig([...fs.readFileSync('./test/data/main.wasm')]);
+        } catch (e) {
+            if (e.message.indexOf('Contract method is not found') === -1) {
+                throw(e);
+            }
+        }
+        const currentKeys = await accountMultisig.getAccessKeys();
+        expect(currentKeys[0].access_key.permission).toEqual('FullAccess');
```
vgrichina, line 251: this should check that public keys are as desired as well

mattlockyer:
Merged in #453

PR #445, opened 2020-11-02
Updates to Account Multisig

mattlockyer:
This PR updates account multisig:
- handle more flow of 2FA using default CH methods
- override via config passed in to near.connect
- add enable/disable via disable and update to deployMultisig


test/account_multisig.test.js
```diff
@@ -0,0 +1,25 @@
...
+    nearjs = await testUtils.setUpTestConnection();
+    workingAccount = await testUtils.createAccountMultisig(nearjs, {
+        getCode: () => 'test'
+    });
+    let nodeStatus = await nearjs.connection.provider.status();
+    startFromVersion = (version) => semver.gte(nodeStatus.version.version, version);
+    console.log(startFromVersion);
+});
+
+test('view pre-defined account works and returns correct name', async () => {
```
vgrichina, line 21: @mattlockyer do you plan to add more tests?

mattlockyer: Yes working on this right now since it's more critical to release:
https://github.com/near/near-api-js/pull/444#pullrequestreview-522312440

Will be adding more tests tonight.

src/account_multisig.ts
```diff
@@ -20,14 +20,36 @@ export const MULTISIG_DEPOSIT = new BN('0');
 export const MULTISIG_CHANGE_METHODS = ['add_request', 'add_request_and_confirm', 'delete_request', 'confirm'];
 export const MULTISIG_VIEW_METHODS = ['get_request_nonce', 'list_request_ids'];
 export const MULTISIG_CONFIRM_METHODS = ['confirm'];
+export const MULTISIG_CONTRACT_HASHES = process.env.MULTISIG_CONTRACT_HASHES || [
```
vgrichina, line 23: @mattlockyer probably need to parse `process.env.MULTISIG_CONTRACT_HASHES`? Though I'd prefer we don't rely on env variables in near-api-js (stuff should be passed through config).

src/account_multisig.ts
```diff
@@ -20,14 +20,36 @@ export const MULTISIG_DEPOSIT = new BN('0');
...
+    // https://github.com/near/core-contracts/blob/636e7e43f1205f4d81431fad0be39c5cb65455f1/multisig/src/lib.rs
+    '8DKTSceSbxVgh4ANXwqmRqGyPWCuZAR1fCqGPXUjD5nZ',
+    // https://github.com/near/core-contracts/blob/f93c146d87a779a2063a30d2c1567701306fcae4/multisig/res/multisig.wasm
+    '55E7imniT2uuYrECn17qJAk9fLcwQW4ftNSwmCJL5Di',
+];
 
+/********************************
+This method can be used to detect if an account on a particular network (connection) is a multisig account
+********************************/
+export const isAccountMultisig = async (connection: Connection, accountId: string): Promise<boolean> => {
```
vgrichina, line 37: what is intended usage pattern for this? Because e.g. if you want to first do `isAccountMultisig` and then create new `AccountMultisig` – you might end up having many extra RPC queries

src/account_multisig.ts
```diff
@@ -36,10 +58,24 @@ let storageFallback = {
...
     public storage: any;
+    public sendCode: sendCodeFunction;
+    public getCode: getCodeFunction;
+    public verifyCode: verifyCodeFunction;
+    public onResult: Function;
 
-    constructor(connection: Connection, accountId: string, storage: any) {
+    /********************************
+    AccountMultisig has options object where you can provide callbacks for:
+    - sendCode: how to send the 2FA code in case you don't use NEAR Contract Helper
```
vgrichina, line 68: @mattlockyer this seems to be very specific to our 2FA implementation vs multisig in general. Should we split into 2 classes with `Account2FA` extending `AccountMultsig`?

vgrichina: (at the very least we should rename, so that it's possible to introduce `AccountMultsig` in follow-up without breaking API

src/account_multisig.ts
```diff
@@ -150,7 +217,27 @@ export class AccountMultisig extends Account {
...
+    async getCodeDefault(method: any): Promise<string> {
+        throw new Error('There is no getCode callback provided. Please provide your own in AccountMultisig constructor options. It has a parameter method where method.kind is "email" or "phone".');
+    }
+
+    async promptAndVerify() {
+        const method = await this.get2faMethod();
+        const securityCode = await this.getCode(method)
+        try {
+            const { success, res: result } = await this.verifyCode(securityCode);
+            console.log(success, result)
```
vgrichina, line 229: nit: looks like debug leftover

src/account_multisig.ts
```diff
@@ -150,7 +217,27 @@ export class AccountMultisig extends Account {
...
+            if (!success || result === false) {
+                throw new Error('Request failed with error: ' + JSON.stringify(result));
+            }
+            return typeof result === 'string' && result.length === 0 ? 'true' : result;
```
vgrichina, line 233: Why are we returning `'true'` string here?

Generally it's very hard to understand what is going on with the types here. Why is there separate `success` and `result`? Why result can be `false` or string?

test/account_multisig.test.js
```diff
@@ -0,0 +1,126 @@
...
+        // auto accept "code"
+        verifyCode: () => ({ success: true, res: '' }),
+        onResult: async () => {
+            const { requestId } = accountMultisig.getRequest();
+            console.log('requestId', requestId);
+            // set confirmKey as signer
+            const originalSigner = nearjs.connection.signer;
+            const tempKeyStore = new InMemoryKeyStore();
+            await tempKeyStore.setKey(nearjs.connection.networkId, accountId, accountMultisig.confirmKey);
+            nearjs.connection.signer = new InMemorySigner(tempKeyStore);
```
vgrichina, line 39: seems like this pattern is repeating enough times that makes sense to extract as utility. `InMemorySigner.fromKeyPair` maybe?

test/account_multisig.test.js
```diff
@@ -0,0 +1,126 @@
...
+    const keys = await account.getAccessKeys();
+    const accountMultisig = new AccountMultisig(nearjs.connection, accountId, {
+        // skip this (not using CH)
+        getCode: () => {},
+        sendCode: () => {},
+        // auto accept "code"
+        verifyCode: () => ({ success: true, res: '' }),
+        onResult: async () => {
+            const { requestId } = accountMultisig.getRequest();
+            console.log('requestId', requestId);
```
vgrichina, line 34: if we are going to keep the log in committed code – let's add some specific message. Otherwise best to cleanup debug logs

test/account_multisig.test.js
```diff
@@ -0,0 +1,126 @@
...
+    });
+    accountMultisig.confirmKey = KeyPair.fromRandom('ed25519');
+    accountMultisig.postSignedJson = () => ({ publicKey: accountMultisig.confirmKey.getPublicKey() });
+    accountMultisig.getRecoveryMethods = () => ({
+        data: keys.map(keyMapping)
+    });
+    try {
+        await accountMultisig.deployMultisig([...fs.readFileSync('./test/data/multisig.wasm')]);
+    } catch (e) {
+        console.log(e);
```
vgrichina, line 55: `console.error`? Also seems like makes sense to log it only when not throwing

test/account_multisig.test.js
```diff
@@ -0,0 +1,126 @@
...
+        if (e.message.indexOf('Contract method is not found') === -1) {
+            throw(e);
```
vgrichina, line 57: nit: just `throw e;`

test/account_multisig.test.js
```diff
@@ -0,0 +1,126 @@
...
+        expect(currentKeys.find(({ public_key }) => keys[0].public_key === public_key).access_key.permission).toEqual('FullAccess');
+        expect(currentKeys.find(({ public_key }) => keys[1].public_key === public_key).access_key.permission).toEqual('FullAccess');
+    });
+
+    test('access key if method is "phone"', async () => {
+        const account = await testUtils.createAccount(nearjs);
+        const keys = await account.getAccessKeys();
+        const accountMultisig = await AccountMultisigInstance(account);
+        const currentKeys = await accountMultisig.getAccessKeys();
+        expect(currentKeys.find(({ public_key }) => keys[0].public_key === public_key).access_key.permission).not.toEqual('FullAccess');
```
vgrichina, line 90: I think it's actually best to join these 2 tests so that you check that decision is made per each key and not for account as a whole.

test/account_multisig.test.js
```diff
@@ -0,0 +1,126 @@
...
+        const accountMultisig = await AccountMultisigInstance(
+            account,
+            ({ public_key: publicKey }, i) => ({ publicKey, kind: i > 0 ? 'ledger' : 'phrase' })
+        );
+        const currentKeys = await accountMultisig.getAccessKeys();
+        expect(currentKeys.find(({ public_key }) => keys[0].public_key === public_key).access_key.permission).toEqual('FullAccess');
+        expect(currentKeys.find(({ public_key }) => keys[1].public_key === public_key).access_key.permission).toEqual('FullAccess');
+    });
+
+    test('access key if method is "phone"', async () => {
```
vgrichina, line 85: it's not clear where `phone` method is setup in the test

test/account_multisig.test.js
```diff
@@ -0,0 +1,126 @@
...
+describe('accountMultisig transactions', () => {
+
+    test('add app key', async() => {
+        let account = await testUtils.createAccount(nearjs);
+        account = await AccountMultisigInstance(account);
+        const appPublicKey = KeyPair.fromRandom('ed25519').getPublicKey();
+        const appMethodNames = ['some_app_stuff', 'some_more_app_stuff'];
+        await account.addKey(appPublicKey, 'foobar', appMethodNames.join(), new BN(parseNearAmount('0.25')));
+        const keys = await account.getAccessKeys();
+        // console.log(keys.map((k) => ({
```
vgrichina, line 104: why is this commented out? Debug leftover?

src/account_multisig.ts
mattlockyer: removed, will detect outside of near-api-js

src/account_multisig.ts
mattlockyer: removed

src/account_multisig.ts
mattlockyer: result is the result from the chain, success was if contract helper could verify the code

test/account_multisig.test.js
vgrichina: moved to separate issue

vgrichina (approved):

PR #449, opened 2020-11-07
Fix adding app key to multisig account

vgrichina:


mattlockyer (approved):
Looks good.

I think this was in one of the other PRs we didn't have time to merge before token transfers were enabled.

I'll add to regression tests for `account_multisig.test.js` today.

vgrichina:
> I think this was in one of the other PRs we didn't have time to merge before token transfers were enabled.
> 

Which of them? This one https://github.com/near/near-api-js/pull/445/files doesn't seem to touch `getKey` method.

> I'll add to regression tests for `account_multisig.test.js` today.

Yeah, I think we need to get full coverage for `account_multisig`. 

Main problem I guess is that we didn't really use apps before Berry Club 🥑

mattlockyer:
Tested recoveryMethods in wallet linked to this branch. Looks like things are working.

The addKey override was likely left over from when everything was in wallet `twoFactor.js`.

Everything that's wallet specific, e.g. adding recovery methods, is handled now in `wallet.js > addAccessKey` but we may want to clean this up in the future.

Issue #450, opened 2020-11-10
Extract creating single-key signer as common utility


vgrichina:
seems like this pattern is repeating enough times that makes sense to extract as utility. `InMemorySigner.fromKeyPair` maybe?

_Originally posted by @vgrichina in [https://github.com/near/near-api-js/pull/445#r520200050_](https://github.com/near/near-api-js/pull/445#r520200050_)

vgrichina:
@mattlockyer as I understand this is done on near-api-js side already, but we still need to convert some of the wallet code, etc. Correct?

mattlockyer:
Yes, it's been working in a test for account_multisig, we can start using it in wallet.

Should we add explicit regression test for this?

vgrichina:
@mattlockyer yeah, makes sense to do a test that it can sign.

PR #451, opened 2020-11-11
add fromKeyPair

mattlockyer:
for #450

### DIFF
--- a/src/signer.ts
+++ b/src/signer.ts
@@ -1,6 +1,6 @@
 import sha256 from 'js-sha256';
 import { Signature, KeyPair, PublicKey } from './utils/key_pair';
-import { KeyStore } from './key_stores';
+import { KeyStore, InMemoryKeyStore } from './key_stores';
 
 /**
  * General signing interface, can be used for in memory signing, RPC singing, external wallet, HSM, etc.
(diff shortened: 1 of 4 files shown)


src/signer.ts
```diff
@@ -12,6 +12,14 @@ export abstract class Signer {
      */
     abstract async createKey(accountId: string, networkId?: string): Promise<PublicKey>;
 
+    /**
+     * Creates a public key for the account, network and keyPair provided
+     * @param accountId accountId to retrieve from.
+     * @param networkId The targeted network. (ex. default, betanet, etc…)
+     * @param keyPair The keyPair to use for signing
+     */
+    abstract async fromKeyPair(accountId: string, networkId: string, keyPair: KeyPair): Promise<PublicKey>;
```
vgrichina, line 21: @mattlockyer let's keep the same order of `networkId` / `accountId` as in `setKey` to avoid additional confusion

mattlockyer:
Not sure why pr build isn't working? Ran tests locally all passed.

src/signer.ts
```diff
@@ -50,6 +58,18 @@ export class InMemorySigner extends Signer {
...
     }
+    
+    /**
+     * Creates a public key for the account, network and keyPair provided
+     * @param networkId The targeted network. (ex. default, betanet, etc…)
+     * @param accountId The NEAR account to assign a public key to
+     * @param keyPair The keyPair to use for signing
+     * @returns {Promise<PublicKey>}
+     */
+    async fromKeyPair(networkId: string, accountId: string, keyPair: KeyPair): Promise<PublicKey> {
```
vgrichina, line 69: @mattlockyer sorry that missed it the first time, but I think following API makes most sense:

`InMemorySigner.fromKeyPair(networkId: string, accountId: string, keyPair: KeyPair): Promise<Signer>`

This is useful every time we need to have throwaway signer. 

Basically create signer instance as a whole from single KeyPair. Generally this only makes sense for in-memory signer, i.e. it won't work for Ledger, etc.

vgrichina:
> Not sure why pr build isn't working? Ran tests locally all passed.

Looks like a flake

PR #452, opened 2020-11-11
Implement account.viewState which allows querying contract state directly

vgrichina:


src/account.ts
```diff
@@ -373,6 +374,19 @@ export class Account {
         return result.result && result.result.length > 0 && parse(Buffer.from(result.result));
     }
 
+    async viewState(prefix: string | Uint8Array, blockQuery: { blockId: BlockId } | { finality: Finality } ) {
+        const { blockId, finality } = blockQuery as any || {};
+        const { values } = await this.connection.provider.query({
+            request_type: 'view_state', 
+            block_id: blockId,
+            finality: blockId ? undefined : finality || 'optimistic',
```
chadoh, line 382: There's a pretty tricky interplay between `blockId` & `finality` here that I'd like to see documented more explicitly. Can we add a code comment above `viewState` to get picked up by typedoc, as well as throwing an error message when you pass a nonsensical combination of the two?

src/account.ts
```diff
@@ -373,6 +374,19 @@ export class Account {
...
+            account_id: this.accountId,
+            prefix_base64: Buffer.from(prefix,).toString('base64')
```
chadoh, line 384: What's with the comma in `(prefix,)`?

src/account.ts
```diff
@@ -373,6 +374,19 @@ export class Account {
...
+        });
+
+        return values.map(({key, value}) => ({ key: Buffer.from(key, 'base64'), value: Buffer.from(value, 'base64')}))
```
chadoh, line 387: Please break up this long line

src/providers/json-rpc-provider.ts
```diff
@@ -68,13 +68,17 @@ export class JsonRpcProvider extends Provider {
 
     /**
      * Query the RPC as [shown in the docs](https://docs.nearprotocol.com/docs/interaction/rpc#query)
-     * @param path Path parameter for the RPC (ex. "contract/my_token")
-     * @param data Data parameter (ex. "", "AQ4", or whatever is needed)
```
chadoh, line 72: Please add documentation to describe the new interface & behavior

src/providers/provider.ts
```diff
@@ -241,6 +242,7 @@ export abstract class Provider {
 
     abstract async sendTransaction(signedTransaction: SignedTransaction): Promise<FinalExecutionOutcome>;
     abstract async txStatus(txHash: Uint8Array, accountId: string): Promise<FinalExecutionOutcome>;
+    abstract async query(params: object): Promise<any>;
     abstract async query(path: string, data: string): Promise<any>;
```
chadoh, line 246: Do we want to have both versions of `query` here?

src/providers/provider.ts
```diff
@@ -26,6 +26,7 @@ type BlockHash = string;
 type BlockHeight = number;
 export type BlockId = BlockHash | BlockHeight;
 
+// TODO: Remove near-final?
```
chadoh, line 29: `near-final` used to be in the docs, but I don't see it anymore. Maybe this was removed. Who can we ask?

Seems like we could make a separate PR that just removes this option, and request review from people who would know enough to approve or reject the change.

chadoh (changes requested):
Thanks for this! Excited to see it get in. I'd like to see some changes related to documentation & devx before it gets merged, but nothing that should take long. See below.

src/providers/json-rpc-provider.ts
vgrichina: @chadoh what inft do you think is going to be useful outside of link to corresponding RPC docs?

If I added a doc in style of the docs we had before it would be something like:
```
@param params parameters for the RPC
```

which I don't think is adding value

src/providers/provider.ts
vgrichina: Yes to avoid breaking backwards compatibility

src/providers/provider.ts
vgrichina: No idea, that's why I just added TODO

src/account.ts
vgrichina: yeah, I forgot about the doc here, need to add

vgrichina: > as well as throwing an error message when you pass a nonsensical combination of the two?

Types kinda describe it. We aren't really checking types during runtime in most other places. TBH this is something I really don't understand about TypeScript. Like doing this whole effort with types and then not having easy way to check in runtime automatically 🤷‍♂️

vgrichina: created issue to track https://github.com/near/near-api-js/issues/457

src/providers/provider.ts
vgrichina: Seems like this is secret feature 🤷‍♂️

vgrichina:
Added docs and tests, etc @chadoh 
Going to merge

PR #453, opened 2020-11-12
hotfix: 2FA key rotations, Account.addKey, InMemorySigner.fromKeyPair

mattlockyer:
- fix deployMultisig to only convert full access keys to multisig limited access keys
- fix to Account.addKey where you can add multiple methodNames
- fix to add one off signer from keyPair

### DIFF
--- a/test/account_multisig.test.js
+++ b/test/account_multisig.test.js
@@ -82,6 +82,17 @@ describe('deployMultisig key rotations', () => {
 
 describe('account2fa transactions', () => {
 
+    test('add app key before deployMultisig', async() => {
+        let account = await testUtils.createAccount(nearjs);
+        const appPublicKey = KeyPair.fromRandom('ed25519').getPublicKey();
+        const appMethodNames = ['some_app_stuff','some_more_app_stuff'];
+        await account.addKey(appPublicKey.toString(), 'foobar', appMethodNames.join(), new BN(parseNearAmount('0.25')));
+        account = await getAccount2FA(account);
+        const keys = await account.getAccessKeys();
+        expect(keys.find(({ public_key }) => appPublicKey.toString() === public_key)
+            .access_key.permission.FunctionCall.method_names.toString()).toEqual(appMethodNames.toString());
+    });
+
     test('add app key', async() => {
         let account = await testUtils.createAccount(nearjs);
         account = await getAccount2FA(account);
--- a/test/account_multisig.test.js
+++ b/test/account_multisig.test.js
@@ -81,7 +81,7 @@ describe('account2fa transactions', () => {
     test('add app key before deployMultisig', async() => {
         let account = await testUtils.createAccount(nearjs);
         const appPublicKey = KeyPair.fromRandom('ed25519').getPublicKey();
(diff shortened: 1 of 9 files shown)


test/account_multisig.test.js
```diff
@@ -82,6 +82,17 @@ describe('deployMultisig key rotations', () => {
...
 describe('account2fa transactions', () => {
 
+    test('add app key before deployMultisig', async() => {
+        let account = await testUtils.createAccount(nearjs);
+        const appPublicKey = KeyPair.fromRandom('ed25519').getPublicKey();
+        const appMethodNames = ['some_app_stuff','some_more_app_stuff'];
+        await account.addKey(appPublicKey.toString(), 'foobar', appMethodNames.join(), new BN(parseNearAmount('0.25')));
+        account = await getAccount2FA(account);
+        const keys = await account.getAccessKeys();
+        expect(keys.find(({ public_key }) => appPublicKey.toString() === public_key)
```
vgrichina, line 92: @mattlockyer need to check the receiver is correct as well

test/account_multisig.test.js
```diff
@@ -82,6 +82,17 @@ describe('deployMultisig key rotations', () => {
 
 describe('account2fa transactions', () => {
 
+    test('add app key before deployMultisig', async() => {
+        let account = await testUtils.createAccount(nearjs);
+        const appPublicKey = KeyPair.fromRandom('ed25519').getPublicKey();
+        const appMethodNames = ['some_app_stuff','some_more_app_stuff'];
+        await account.addKey(appPublicKey.toString(), 'foobar', appMethodNames.join(), new BN(parseNearAmount('0.25')));
```
vgrichina, line 89: not sure why `.join()` call is needed on `appMethodNames`. Same about `.toString()` BTW.

test/account_multisig.test.js
```diff
@@ -82,6 +82,17 @@ describe('deployMultisig key rotations', () => {
...
+        account = await getAccount2FA(account);
+        const keys = await account.getAccessKeys();
+        expect(keys.find(({ public_key }) => appPublicKey.toString() === public_key)
+            .access_key.permission.FunctionCall.method_names.toString()).toEqual(appMethodNames.toString());
```
vgrichina, line 93: should actually check arrays are equal instead of checking `.toString()` is equal, which I assume is passing the test disregarding wrong input (`appMethodNames.join()`).

@mattlockyer I guess it is needed to merge https://github.com/near/near-api-js/pull/436/files for multiple methods support?

mattlockyer: Yes

test/account_multisig.test.js
mattlockyer: odd behavior with near-api-js returning arrays that don't deepEqual

vgrichina: @mattlockyer but you basically end up testing adding a key for one method_name with comma inside instead of testing multiple method names. This is definitely not what is desired.

mattlockyer: Needs 436 I think?

PR #454, opened 2020-11-12
Update README.md

TrevorJTClarke:
because as much as i like ocumentation, i figured it could be better for searching `doc...`


chadoh (approved):
❤️

PR #461, opened 2020-11-30
fix(utils/web): Fix typo in Content-Type header

rlisowski:
Reasons:
- RFC says it's Content-Type, not Content-type
- node-fetch package is checking for the `Content-Type` header in a case sensitive manner and overwrites it when cannot find it https://github.com/node-fetch/node-fetch/blob/master/src/request.js#L71
- because of the node-fetch Content-Type check, the header is sometimes overwritten and near-api-js call is not always recognized as application/json request


frol (approved):

frol:
@vgrichina @bowenwang1996 do you know what is going on with the betanet CI?

bowenwang1996:
Not sure. @chefsale @mhalambek can you take a look?

PR #462, opened 2020-12-01
add newLocalPublicKey to disable so we can rotate

mattlockyer:
This cleans up disable method and adds a new arg `newLocalPublicKey` so we can rotate in a new localStorage key from the wallet.


mattlockyer:
ci-betanet is failing... tests run locally

src/account_multisig.ts
```diff
@@ -159,54 +159,96 @@ export class Account2FA extends AccountMultisig {
...
+            deleteKey(localPublicKey),
+            addKey(newLocalPublicKey, null),
             ...lak2fak.map(({ public_key }) => deleteKey(public_key)),
             ...lak2fak.map(({ public_key }) => addKey(public_key, null)),
             deployContract(contractBytes),
         ]
         console.log('disabling 2fa for', accountId)
-        return await this.signAndSendTransaction(accountId, actions)
+        const result = await this.signAndSendTransaction(accountId, actions)
+        // check if newLocalPublicKey was added to account, if not whole tx failed
```
vgrichina, line 245: @mattlockyer not sure I understand this logic? `signAndSendTransaction` should be throwing error when transaction failed. Is there some case where it doesn't throw properly? Basically how to trigger this code?

src/account_multisig.ts
```diff
@@ -159,54 +159,108 @@ export class Account2FA extends AccountMultisig {
...
             ...lak2fak.map(({ public_key }) => addKey(public_key, null)),
             deployContract(contractBytes),
         ]
         console.log('disabling 2fa for', accountId)
-        return await this.signAndSendTransaction(accountId, actions)
+        let result
+        try {
+            result = await this.signAndSendTransaction(accountId, actions)
+        } catch (e) {
+            if (!/RetriesExceeded/.test(e.toString())) {
```
vgrichina, line 250: nit: this is thrown right out of near-api-js, so it has `type` set, no need for regexps

src/account_multisig.ts
```diff
@@ -159,54 +159,108 @@ export class Account2FA extends AccountMultisig {
...
+                throw e
+            }
+        }
+        return await this.hasPublicKey(newLocalPublicKey) && result
```
vgrichina, line 254: I think this has to happen on the side of the caller app. Retrying immediately if transaction gets delayed is unlikely to help, but you'll need to deal with stuff like keep key backup in case it succeeded later.

PR #467, opened 2020-12-10
feat: add `completedTransactions` cache

chadoh:
For an example of how this would be used, see https://github.com/near/rainbow-bridge-frontend/pull/48

Sometimes a transaction requires confirmation in NEAR Wallet.

1. `Person` takes action in `App`
2. `App` calls `Function` in `Contract` on behalf of `Person`
3. `Function` requires confirmation from `Person`. This could happen if:
   * `Function` has an attached deposit
   * `Person` signed into `App` using a contract other than `Contract` (`walletConnection.requestSignIn(contractName)` was given a `contractName` other than the name of `Contract`)
   * `Person` signed into `App` using `Contract`, which grantend a FunctionCall access key to `App`, but this access key ran out of allowance
4. near-api-js redirects to NEAR Wallet
5. `Person` clicks Approve or Deny
6. NEAR Wallet redirects back to `App`
7. `App` needs to figure out if the call to `Function` succeeded or failed

near-api-js provides two main ways for `App` to call `Function`:

* `Account.functionCall`
* A `changeMethod` on a `Contract` – under the hood, this calls `Account.functionCall`

Both of these previously accepted positional arguments for `gas` and `deposit`. Instead, now they accept an `options` object, with `gas` and `deposit` keys. In addition, you can provide a `meta` key, with its value a free-form object. If `meta` is provided, this transaction will be saved to a cache, along with your `meta` data. Whether or not the transaction results in a redirect to NEAR Wallet, once complete you can check the outcome in `walletConnection.completedTransactions`.

For now, `completedTransactions` only provides one way to look up a specific completed transaction: the `remove` method. Your app will only want to process a completed transaction once, so this method automatically removes a found transaction from the underlying cache.

Where is the cache? If in a browser, near-api-js will use localStorage. If used from Node, near-api-js imports the `node-localstorage` module, which writes the data to a file. By default, this file will live in `~/.near-api-js`, but you can pick a different folder by setting the `NEAR_CACHE_DIR` environment variable.

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -24,17 +24,10 @@ import { PublicKey } from './utils/key_pair';
 import { PositionalArgsError } from './utils/errors';
 import { parseRpcError, parseResultError } from './utils/rpc_errors';
 import { ServerError } from './generated/rpc_error_types';
-import { cacheTransaction, markTransactionFailed, markTransactionSucceeded } from './cached-transactions';
+import { ChangeMethodOptions } from './wallet-account';
+import { BasicCachedTransaction, TxMetadata } from './cached-transactions';
 
 import exponentialBackoff from './utils/exponential-backoff';
-
-type TxMetadata = { [key: string]: string | number };
-
-export type ChangeMethodOptions = {
-    gas?: BN;
-    deposit?: BN;
-    meta?: TxMetadata;
-}
 
 // Default amount of gas to be sent with the function calls. Used to pay for the fees
 // incurred while running the contract execution. The unused amount will be refunded back to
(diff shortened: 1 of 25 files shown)


src/contract.ts
```diff
@@ -45,14 +44,42 @@ export class Contract {
...
+                    async (
+                        args: object = {},
+                        optionsOrGas?: ChangeMethodOptions | BN,
+                        deposit?: BN,
+                        ...ignored
+                    ) => {
+                        const options = isObject(optionsOrGas)
+                            ? optionsOrGas as ChangeMethodOptions
+                            : {
+                                gas: (() => {
```
vgrichina, line 58: this construction seems a bit complicated. Why not just have `if` at top-level?

src/contract.ts
```diff
@@ -45,14 +44,42 @@ export class Contract {
...
-                        throw new PositionalArgsError();
+                value: nameFunction(
+                    methodName,
+                    async (
+                        args: object = {},
+                        optionsOrGas?: ChangeMethodOptions | BN,
+                        deposit?: BN,
+                        ...ignored
+                    ) => {
+                        const options = isObject(optionsOrGas)
```
vgrichina, line 55: seems like can be a problem in case when `BN` is used for gas (isn't it an object?)

src/account.ts
```diff
@@ -144,20 +153,38 @@ export class Account {
...
-        let txHash, signedTx;
+        let txHash: Uint8Array, signedTx: SignedTransaction;
         // TODO: TX_NONCE (different constants for different uses of exponentialBackoff?)
         const result = await exponentialBackoff(TX_STATUS_RETRY_WAIT, TX_NONCE_RETRY_NUMBER, TX_STATUS_RETRY_WAIT_BACKOFF, async () => {
             [txHash, signedTx] = await this.signTransaction(receiverId, actions);
             const publicKey = signedTx.transaction.publicKey;
 
+            if (meta) {
+                cacheTransaction({
+                    hash: txHash.toString(),
```
vgrichina, line 176: I wouldn't rely on `txHash.toString()` as `.toString()` is different between `Buffer` and `Uint8Array` which can be fun. I'd better convert explicitly like `Buffer.from(txHash).toString('hex')`

src/account.ts
```diff
@@ -144,20 +153,38 @@ export class Account {
...
+                    meta,
+                    publicKey: publicKey.toString(),
+                    receiverId,
+                    signedTx: signedTx.encode().toString(),
```
vgrichina, line 180: same as above – probably better to convert explicitly using `Buffer`

src/account.ts
```diff
@@ -144,20 +153,38 @@ export class Account {
...
+                });
+            }
+
             try {
                 const result = await exponentialBackoff(TX_STATUS_RETRY_WAIT, TX_STATUS_RETRY_NUMBER, TX_STATUS_RETRY_WAIT_BACKOFF, async () => {
                     try {
+                        // TODO: Where does redirect to NEAR Wallet happen?
```
vgrichina, line 187: @chadoh redirect happens here basically:
https://github.com/near/near-api-js/blob/0aefdb01a151f7361463f3ff65c77dbfeee83200/src/wallet-account.ts#L208
https://github.com/near/near-api-js/blob/0aefdb01a151f7361463f3ff65c77dbfeee83200/src/wallet-account.ts#L107

src/storage.ts
```diff
@@ -0,0 +1,37 @@
+if (typeof localStorage === 'undefined' || localStorage === null) {
```
vgrichina, line 1: this definitely should be something you can pass in a constructor, etc vs always using same global object. Like e.g. when you are running tests you most likely would want to use in-memory storage. Some web apps would want to use `sessionStorage` instead of `localStorage`, etc.

src/contract.ts
chadoh: Updated to:

```ts
let options = optionsOrGas as ChangeMethodOptions | undefined;
if (!options && !deposit) {
    options = {} as ChangeMethodOptions;
} else if (!options || !(options.gas || options.deposit || options.meta)) {
    // passed positional gas or deposit
    const deprecate = depd('positional gas & deposit amount');
    deprecate('pass { gas: inUnits, deposit: inYoctoNEAR } instead');
    options = {
        gas: optionsOrGas as BN,
        deposit
    } as ChangeMethodOptions;
}
```

src/storage.ts
chadoh: That seems like it would add a bit of complexity to a currently simple interface, when we're not sure anyone cares. Unless I'm mistaken, the tests should pass just fine even if they write some data to a file.

But thanks for the comment! I noticed that I set the localStorage directory to `./scratch` by accident, which is not what I meant to do. 

I've updated it to this instead, which will allow CLI people to control where the cache directory goes while providing a reasonable default:

```ts
if (typeof localStorage === 'undefined' || localStorage === null) {
    const { LocalStorage } = require('node-localstorage'); // eslint-disable-line @typescript-eslint/no-var-requires
    localStorage = new LocalStorage(process.env.NEAR_CACHE_DIR || '~/.near-api-js');
}
```

Marking this as resolved, but feel free to reopen the discussion if you feel strongly that we need to provide more configurability for the initial release of this feature.

vgrichina:
@chadoh BTW, I assume you plan to add tests later for this PR, correct?

chadoh:
@vgrichina yes

chadoh:
We need this one

PR #468, opened 2020-12-16
Error schema and types synced with nearcore

volovyk-s:


chadoh (approved):
Thank you!

What was the process of generating this change? It seems like something we should automate.

volovyk-s:
> Thank you!
> 
> What was the process of generating this change? It seems like something we should automate.

It's half automated but still confusing. 
There is an instruction at the bottom of the README file.
The downsides are next:
- schema should be regenerated manually on the `nearcore` side. It means that it can be, and it is, outdated.
- it takes a long time to merge it to the `nearcore` (https://github.com/near/nearcore/pull/3727) @evgenykuzyakov 
- we should update `gen_error_types.js` with the new commit manually and run the script to generate new error types.

Looks like an attempt to automate it will be too complicated. The good news is that hopefully this logic will be moved to  `nearcore` itself, so it should solve the issue. I'm trying to speed up this process.

vgrichina:
@volovyk-s @chadoh I think we should retire this whole auto-generation approach as it doesn't really play well with JS (because of stuff like loading same class twice, etc) https://github.com/near/near-api-js/issues/331

PR #469, opened 2020-12-16
Adding keyPath uses that keystore first

mikedotexe:
I've noticed this on and off and finally figured out it's a one-liner.
When I create a function-call access key for an account and then specify it as the `keyPath`, the full access key was being used.
When a person specifies the keyPath, I believe that means, "Please use this keystore first, as I'm specifying to use it." Instead it's added to the end of the MergeKeyStore, so it's the last keystore to be checked.
This simply flips that so the first keystore used is the one specified by keyPath.

### DIFF
--- a/src/near.ts
+++ b/src/near.ts
@@ -112,7 +112,7 @@ export async function connect(config: ConnectConfig): Promise<Near> {
                 if (!config.masterAccount) {
                     config.masterAccount = accountKeyFile[0];
                 }
-                config.deps.keyStore = new MergeKeyStore([config.deps.keyStore, keyPathStore]);
+                config.deps.keyStore = new MergeKeyStore([keyPathStore, config.deps.keyStore]);
                 console.log(`Loaded master account ${accountKeyFile[0]} key from ${config.keyPath} with public key = ${keyPair.getPublicKey()}`);
             }
         } catch (error) {


chadoh (approved):
I agree with this, thank you! I think we should mark this as a breaking change. Do you mind trying out the [Conventional Commits](https://www.conventionalcommits.org/) way of marking such a change, even though this project has no tooling to automatically generate changelogs with such commits? I believe we would want something like:

```
feat!: added keyPath becomes primary keystore

BREAKING CHANGE: if your code relies on the primary keystore to stay unchanged after adding a new keyPath, you will need to update it to work with this new feature. [and then maybe give an example here]
```

Such a commit and PR description will still show up in our release notes, and make it easier to write useful notes.

volovyk-s (approved):

vgrichina (changes requested):
I think we shouldn't make this change (at least as is right now). What will happen as result is that all keys created during near-cli command execution will be written in `InMemoryKeyStore` created in this method and will be lost. 

We can add another argument to `MergeKeyStore` to specify keyStore for writing keys separately from read priority.

mikedotexe:
> I think we shouldn't make this change (at least as is right now). What will happen as result is that all keys created during near-cli command execution will be written in `InMemoryKeyStore` created in this method and will be lost.
> 
> We can add another argument to `MergeKeyStore` to specify keyStore for writing keys separately from read priority.

Interesting. So you're talking about a command like `near generate-key…` ?
I mean, if someone has gone out of their way to specify a `keyPath`, doesn't that mean that the only difference this PR changes is an Unencrypted File Keystore takes precedence? I don't understand the problem, perhaps. Do you have an example of when adding a custom `keyPath` would result in a command causing a key to get lost because it was written to memory?

vgrichina:
> I mean, if someone has gone out of their way to specify a `keyPath`, doesn't that mean that the only difference this PR changes is an Unencrypted File Keystore takes precedence? I don't understand the problem, perhaps. Do you have an example of when adding a custom `keyPath` would result in a command causing a key to get lost because it was written to memory?

`keyPathStore` is an `InMemoryKeyStore`. It gets precedence so all new keys are _written_ to it. So if you e.g. try to create new account while specifying `keyPath` to provide key for you `masterAccount` you would lose a key to newly created account.

PR #490, opened 2021-01-27
feat(travis): add fossa integration to travis

chefsale:


chadoh (changes requested):
Sorry, I haven't been able to work on NAJ for about a month. When I get back to it, one of the first orders of business is removing dependency on Travis. Travis has quit working for various NEAR project multiple times over the past couple months, and it does so in the worst way. Rather than failing all builds with a message explaining how to make them pass again, it just stops running altogether. Since we don't have other steps in the CI process for many projects, this makes it seem like the builds pass, because there aren't any. And then tools like Dependabot which are set up to auto-merge if builds pass start merging a bunch of breaking changes.

So we need to change to GitHub Actions or some other CI provider, and integrate fossa with it.

.travis.yml
```diff
@@ -1,12 +1,27 @@
...
+  include:
+    - name: yarn
+      script:
+      - yarn lint
+      - yarn test
+      - yarn browserify && yarn run bundlewatch
+
+    - name: fossa     
+      before_script:
+        - "curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh | sudo bash"
```
chadoh, line 23: Piping `curl` into `sudo bash`! Fun times. Should this at least use a specific (vetted) commit, rather than the hot version from `master`? Seems like a security issue, even if it is only running in a CI environment.

chefsale: Pinned to a specific commit now than.

chefsale:
Can we land this for now and once you plan to migrate this, you let me know and I can do it post landing this PR. We really shouldn't be commiting anything which is not licence compliant or a security issue and this will make it prevent from happening. Later on we can migrate the CI which shouldn't be a big problem, until then this is the state of the world.

chadoh (approved):
Travis isn't even running on this repository right now. Seems like they decided to nickle-and-dime for running Windows builds. Or macOS builds? Either way, Travis isn't running. 

I have required any new commits to `master` to have an approving PR review from me. And I'm not checking this repository again until Rainbow Bridge UI lands on mainnet and Rainbow Bridge Lib v1 exists.

So this commit won't do anything!

That said, having it merged will make it easier to ensure that `fossa` checks come along for the ride when I migrate to a new CI environment.

Thanks!

chadoh:
Hmm, looks like Travis started again? 😂🤷🏻

chefsale:
Not really sure what you mean they are not running, I can clearly see the build in travis: https://travis-ci.com/github/near/near-api-js/builds/214956011, as well all the reports for my PR and other are generated in fossa: https://app.fossa.com/projects/custom%2B21311%2Fgithub.com%2Fnear%2Fnear-api-js/history/builds. So, at the moment it seems like they are running fine. Thank you for the approval.

chefsale:
Yes, they are alive 😺

chadoh:
oh, lol, everything I said applies to create-near-app, not near-api-js 🤦🏻‍♂️😖

thanks for bearing with me Sandi (and any other confused onlookers)

chefsale:
Haha, no worries. I assumed you are talking about another repo, as it didn't make much sense. At least you made my laugh in the end 💯

PR #492, opened 2021-02-03
Don't fetch account state pre-emptively

vgrichina:
This creates extra RPC requests and seems to be there only for historic reasons.

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -1,6 +1,7 @@
 'use strict';
 
 import BN from 'bn.js';
+import depd from 'depd';
 import {
     transfer,
     createAccount,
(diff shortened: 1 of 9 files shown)


src/account.ts
```diff
@@ -78,33 +78,18 @@ function parseJsonFromRawResponse (response: Uint8Array): any {
...
         this.accountId = accountId;
     }
 
-    /**
-     * Helper function when getting the state of a NEAR account
-     * @returns Promise<void>
-     */
-    async fetchState(): Promise<void> {
-        this._state = await this.connection.provider.query(`account/${this.accountId}`, '');
-    }
```
chadoh, line 99: > BREAKING CHANGE: This removes some methods and fields from `Account` which potentially can be used somewhere (like `.fetchState()`
> 
> Change potentially can be made non-breaking if we replace these by no-ops. 

Making it non-breaking and adding a `depd` deprecation warning for `fetchState` and `ready` seems easy enough. When we publish a v1, we can drop all such `depd` warnings.

chadoh (changes requested):

src/account.ts
vgrichina: @chadoh seems good to me, gonna do it before getting out of draft

PR #496, opened 2021-02-09
Types added to some JSON RPC errors

volovyk-s:
To avoid the situations where we are matching string errors with the use of `match`, `indexOf`, or `includes` we can do this only once in JSON RPC provider.
This logic can be deleted when `nearcore` will start returning typed errors.


vgrichina:
@volovyk-s let's make sure we have tests for these

Issue #397, opened 2020-09-07
Support or document a use in mixed (node.js and browser) environments (Can't resolve 'fs')


frol:
**Is your feature request related to a problem? Please describe.**

I am frustrated to see that every project that uses near-api-js in mixed environments (e.g. Next.js transparently enables developers to write server-side-renderable application) bumps into weird:

```
./node_modules/near-api-js/lib/key_stores/unencrupted_file_system_keystore.js:7:0
Module not found: Can't resolve 'fs'
```

![image](https://user-images.githubusercontent.com/304265/92389223-121d9500-f0ef-11ea-9203-3f3c3d36fe68.png)

**Describe the solution you'd like**

Ideally, I would like to avoid a need to have any configuration on the developer side.

**Describe alternatives you've considered**

The alternative is to have it properly documented and have the solution immediately discoverable. For Next.js, you have to tweak webpack by conditionally resolve "fs" to an empty module:

```js
node = {
  fs: "empty",
}
```

See [more details on StackOverflow](https://stackoverflow.com/questions/39249237/node-cannot-find-module-fs-when-using-webpack/).

Here is how we deal with it on Explorer frontend (Next.js): https://github.com/near/near-explorer/blob/745e9c6db01aad2f0feb8d1e2ab9f61538c6565d/frontend/next.config.js#L37-L47

/cc @mikedotexe @vgrichina @chadoh

vgrichina:
@frol we have separate entry point for browser which doesn't include `fs` (`browser-index`). Is Next.js ignoring `browser` setting in `package.json`  (https://github.com/near/near-api-js/blob/master/package.json#L11)?

If so – this seems like Next.js bug. Are there other tools suffering the same issue? Is it a problem with some underlying bundler?

vgrichina:
Looks like some stuff was pulled internally that ended up bringing `fs` even after refactoring obvious entry point (`keyStores`)

PR #499, opened 2021-02-10
Make sure that webpack build doesn’t end up requiring `fs`

vgrichina:
Fixes https://github.com/near/near-api-js/issues/397

### DIFF
--- a/src/signer.ts
+++ b/src/signer.ts
@@ -1,6 +1,7 @@
 import sha256 from 'js-sha256';
 import { Signature, KeyPair, PublicKey } from './utils/key_pair';
-import { KeyStore, InMemoryKeyStore } from './key_stores';
+import { KeyStore } from './key_stores/keystore';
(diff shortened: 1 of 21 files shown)


src/signer.ts
```diff
@@ -1,6 +1,7 @@
 import sha256 from 'js-sha256';
 import { Signature, KeyPair, PublicKey } from './utils/key_pair';
-import { KeyStore, InMemoryKeyStore } from './key_stores';
```
volovyk-s, line 3: What is the reason to do this?

volovyk-s (approved):

src/signer.ts
vgrichina: This bug https://github.com/near/near-api-js/issues/397

Apparently we pack doesn’t care that only specific classes are imported, key stores module ends up depending on fs.

PR #506, opened 2021-02-18
requestSignIn: deprecate title param, fix typing

chadoh:
Also:

* Export ConnectedWalletAccount so that TypeScript projects can
  properly add typings for the return value of
  `(new WalletConnection()).account()`
* Make `findAccessKey` non-private, since it's the best way I've found
  to determine if the user is authenticated against the desired
  contract. JS doesn't care that it's marked private and allows anyone
  to use it anyhow, but TypeScript projects throw warnings when trying
  to use this method.


src/wallet-account.ts
```diff
@@ -60,28 +61,36 @@ export class WalletConnection {
     /**
      * Redirects current page to the wallet authentication page.
      * @param contractId The NEAR account where the contract is deployed
-     * @param title Name of the application that will appear as requesting access in Wallet
-     * @param successUrl Optional url to redirect upon success
-     * @param failureUrl Optional url to redirect upon failure
-     * 
+     * @param options An optional options object
+     * @param options.successUrl URL to redirect upon success. Default: current url
```
vgrichina, line 65: given we are changing this wonder if we should go with https://github.com/near/NEPs/pull/10/files and use one callback URL + return error info from wallet

chadoh: I would prefer to avoid a breaking change if we can. I believe the change proposed here would easily allow adding `callbackUrl` to the `options` later, in support of NEP-10. Right?

vgrichina: @chadoh right, didn't mean it to be a breaking change. Not a blocker

vgrichina:
@chadoh looks good overall but compilation fails for some reason:
https://travis-ci.com/github/near/near-api-js/jobs/484091428

frol (approved):

chadoh:
Should we close this in favor of #524 ?

PR #507, opened 2021-02-18
New error handling design

volovyk-s:


src/account.ts
```diff
@@ -184,7 +182,7 @@ export class Account {
                     `Transaction ${result.transaction_outcome.id} failed. ${result.status.Failure.error_message}`,
                     result.status.Failure.error_type);
             } else {
-                throw parseResultError(result);
+                throw parseRpcResultError(result);
```
frol, line 185: Let's call it `parseTransactionExecutionFailure`

src/utils/rpc_errors.ts
```diff
@@ -2,30 +2,50 @@
 import Mustache from 'mustache';
 import schema from '../generated/rpc_error_schema.json';
 import messages from '../res/error_messages.json';
-import * as CLASSMAP from '../generated/rpc_error_types';
-import { ServerError } from '../generated/rpc_error_types';
+import { TypedError } from './errors';
+import { ExecutionOutcomeWithIdView } from '../providers/provider';
 
-export * from '../generated/rpc_error_types';
+export class ServerError extends TypedError {
```
frol, line 8: I suggest merging `ServerError` and `ServerTransactionError` into `TransactionExecutionFailure` type with required `transaction_outcome` (for backward compatibility, if needed) and `context` fields.

src/utils/rpc_errors.ts
```diff
@@ -2,30 +2,50 @@
...
+    constructor(transactionHash?: string, errorPath?: Record<string, any>) {
+        this.transactionHash = transactionHash;
+        this.errorPath = errorPath;
+    }
+}
 
-class ServerTransactionError extends ServerError {
-    public transaction_outcome: any;
+export class ServerTransactionError extends ServerError {
+    public transaction_outcome: ExecutionOutcomeWithIdView; //TODO: should it be a part of context?
```
frol, line 31: ^^

src/utils/rpc_errors.ts
frol: I also suggest renaming the file (`rpc_errors.ts`) or reorganizing it altogether (e.g. somehow merge it with the `transaction.ts`)

volovyk-s: Looks like a merge of these two classes is not possible with the current architecture, as well as part of the renaming that you have mentioned.
Basically, we are using `ServerError` for these type of errors:
`{"jsonrpc":"2.0","error":{"code":-32000,"message":"Server error","data":{"TxExecutionError":{"InvalidTxError":{"InvalidNonce":{"tx_nonce":2,"ak_nonce":28}}}}},"id":124}`

And `ServerTransactionError` for these type of errors:

`Response.result:
{
  status: { Failure: { ActionError: [Object] } },
  transaction: {},
  transaction_outcome: {},
  receipts_outcome: []
}`

Also, let's not rename `type` to `error_kind` in this PR because it would be a breaking change. Let's deliver this change in a separate PR after all changes in `nearcore`.

volovyk-s:
I Will request feedback after these changes: https://github.com/near/near-api-js/discussions/515#discussioncomment-543925

mikedotexe:
Update: Serhii will break out some code into separate PRs.

PR #508, opened 2021-02-19
feat: Exposed experimental_protocolConfig provider method

frol:
It is a follow up on the new experimental method in nearcore JSON RPC: https://github.com/near/nearcore/pull/3919

# Testing plan

* [x] Tested with CI
* [x] Manually tested against current testnet, which has the `EXPERIMENTAL_protocol_config` support
* [x] Manually tested against current mainnet, which currently does not have `EXPERIMENTAL_protocol_config` method


frol:
@vgrichina This is future and backward compatible implementation for `EXPERIMENTAL_protocol_config` API. I don't think we need nearcore release to support this scenario. /cc @bowenwang1996

Issue #335, opened 2020-05-27
Log in without AccessKey


ilblackdragon:
There are cases when application needs to have access to account id of the user but doesn't need to have FunctionCall AccessKey.
For example if UI uses NEP-21 and doesn't want to claim custody of this token.

We need a flow for this to inform app's UI about account id (and let user create account).

PR #524, opened 2021-03-05
 Allow wallet sign in without asking for access key

vgrichina:
Fixes https://github.com/near/near-api-js/issues/335

Needs https://github.com/near/near-wallet/pull/1523 on wallet side


chadoh (approved):
Thank you!

PR #528, opened 2021-03-11
getNetwork deleted from Provider

### LINKED ISSUE #282: Fix hardcoded getNetwork to return current net
https://github.com/near/near-api-js/blob/master/src/providers/json-rpc-provider.ts#L26-L35

```
async getNetwork(): Promise<Network> {
        return {
            name: 'test',
            chainId: 'test'
        };
```

With other nets like betanet coming on, seems like we need a way to determine this. I see there's a `url` available but parsing that is probably not the best approach.

volovyk-s:
closes #282 

`getNetwork()` deleted from `Provider` interface and `JsonRpcProvider`.
@vgrichina @mikedotexe it's a breaking change, but we don't have any mentions about `getNetwork` in our docs. I have also deleted `Network` interface since it's never used (YAGNI).


volovyk-s:
@vgrichina we can choose a softer way: https://github.com/near/near-api-js/pull/531
And merge this PR in a few months.

vgrichina (approved):

PR #533, opened 2021-03-17
Feat/json rpc query types

### LINKED ISSUE #525: getAccountDetails returns empty transactions array
**Describe the bug**
`getAccountDetails` function claims to return transactions but always returns nothing.
See https://github.com/near/near-api-js/blob/960b798fe009fe0658637803af9645f550f139cb/src/account.ts#L402

**To Reproduce**
1. Create `let account = new Account(account_id);`
2. Call `account.getAccountDetails()`
3. Observe the empty `transactions` in the result

**Expected behavior**
Should remove this field and update comments as it's confusing

mehtaphysical:
Fixes #525

In this PR:

* RPC query request typing in `src/providers/provider.ts` see `RpcQueryRequest`
* RPC query response typing in `src/providers/provider.ts` see`QueryResponseKind`
* update `src/account.ts` to use params object to take advantage of the request typing
* update `src/accounts.ts` to use `query<T>` response typing

Follow up: It would be nice to figure out how to separate the types from the `provider.ts` file. Maybe into separate files by type. `rpc-block-views.ts`, `rpc-query-views.ts`, etc. But this may leave some of the files a bit anemic.

Very open to any feedback on this.

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -14,12 +14,11 @@ import {
     deleteKey,
     stake,
     deleteAccount,
-    AccessKey,
     Action,
     SignedTransaction
 } from './transaction';
 import { FinalExecutionOutcome, TypedError, ErrorContext } from './providers';
-import { Finality, BlockId } from './providers/provider';
+import { Finality, BlockId, ViewStateResult, AccountView, AccessKeyView, CodeResult, AccessKeyList, AccessKeyInfoView, FunctionCallPermissionView } from './providers/provider';
 import { Connection } from './connection';
 import { baseDecode, baseEncode } from 'borsh';
 import { PublicKey } from './utils/key_pair';
@@ -46,18 +45,17 @@ const TX_NONCE_RETRY_WAIT = 500;
 // Exponential back off for waiting to retry.
 const TX_NONCE_RETRY_WAIT_BACKOFF = 1.5;
 
-export interface AccountState {
-    amount: string;
-    code_hash: string;
-    storage_usage: number;
-    locked: string;
-}
-
 export interface AccountBalance {
     total: string;
     stateStaked: string;
     staked: string;
     available: string;
+}
+
+export interface AccountAuthorizedApp {
+    contractId: string;
+    amount: string;
+    publicKey: PublicKey;
 }
 
 interface ReceiptLogWithFailure {
@@ -95,10 +93,10 @@ export class Account {
 
     /**
      * Returns the state of a NEAR account
-     * @returns {Promise<AccountState>}
+     * @returns {Promise<AccountView>}
      */
-    async state(): Promise<AccountState> {
-        return await this.connection.provider.query(`account/${this.accountId}`, '');
+    async state(): Promise<AccountView> {
+        return await this.connection.provider.query<AccountView>(`account/${this.accountId}`, '');
     }
 
     private printLogsAndFailures(contractId: string, results: [ReceiptLogWithFailure]) {
@@ -189,9 +187,9 @@ export class Account {
         return result;
     }
 
-    accessKeyByPublicKeyCache: { [key: string]: AccessKey } = {}
+    accessKeyByPublicKeyCache: { [key: string]: AccessKeyView } = {}
 
-    private async findAccessKey(receiverId: string, actions: Action[]): Promise<{publicKey: PublicKey; accessKey: AccessKey}> {
+    private async findAccessKey(receiverId: string, actions: Action[]): Promise<{publicKey: PublicKey; accessKey: AccessKeyView}> {
         // TODO: Find matching access key based on transaction
         const publicKey = await this.connection.signer.getPublicKey(this.accountId, this.connection.networkId);
         if (!publicKey) {
@@ -204,7 +202,7 @@ export class Account {
         }
 
         try {
-            const accessKey = await this.connection.provider.query(`access_key/${this.accountId}/${publicKey.toString()}`, '');
+            const accessKey = await this.connection.provider.query<AccessKeyView>(`access_key/${this.accountId}/${publicKey.toString()}`, '');
             this.accessKeyByPublicKeyCache[publicKey.toString()] = accessKey;
             return { publicKey, accessKey };
         } catch (e) {
@@ -345,7 +343,7 @@ export class Account {
     ): Promise<any> {
         args = args || {};
         this.validateArgs(args);
-        const result = await this.connection.provider.query(`call/${contractId}/${methodName}`, baseEncode(JSON.stringify(args)));
+        const result = await this.connection.provider.query<CodeResult>(`call/${contractId}/${methodName}`, baseEncode(JSON.stringify(args)));
         if (result.logs) {
             this.printLogs(contractId, result.logs);
         }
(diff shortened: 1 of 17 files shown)


mehtaphysical:
It's a great point @vgrichina. I'm guessing you are mostly concerned about the `AccessKey` to `AccessKeyView` change?

The problem is that in `near-api-js` there is right now (before this change) two types of `AccessKey` that is being hidden because of the use of the `any` type. These are

* the `AccessKey` as defined here: https://github.com/near/near-api-js/blob/master/src/transaction.ts#L22
* the `AccessKey` that is returned by the RPC which I've named `AccessKeyView` in this PR following `nearcore` https://github.com/near/nearcore/blob/master/core/primitives/src/views.rs#L184

For the `AccessKey` it is a type used for the borsh serialization and has something like this shape:

```
    "accessKey": {
      "nonce": "00",
      "permission": {
        "fullAccess": {},
        "enum": "fullAccess"
      }
    }
```

You can see this here: https://github.com/near/near-api-js/blob/master/src/transaction.ts#L17 and https://github.com/near/near-api-js/blob/master/src/transaction.ts#L164.

To use this type in the Typescript it would mean doing something like `accessKey.permissions.fullAccess` and `accessKey.permissions.functionCall`. You can sort of see this here: https://github.com/near/near-api-js/blob/master/src/account_multisig.ts#L108

This is not the type of thing the RPC returns and not how the `AccessKey`s are used in many places: https://github.com/near/near-api-js/blob/master/src/account_multisig.ts#L195. That is because the access keys returned by the `account.getAccessKeys` method is not the same as the `AccessKey` type. The access keys returned by the RPC look like:

```json
          {
                "public_key": "ed25519:LjcEBmDxwEA1mbQYxz2By3iRHGaWvYJycXX1nQdRm7z",
                "access_key": {
                    "nonce": 4,
                    "permission": "FullAccess"
                }
            },
            {
                "public_key": "ed25519:WMd8rgbwB2KQ26Fc31zGn64Qc1xaTWbN2kknfK4sesq",
                "access_key": {
                    "nonce": 0,
                    "permission": {
                        "FunctionCall": {
                            "allowance": "250000000000000000000000",
                            "receiver_id": "undefined",
                            "method_names": []
                        }
                    }
                }
            },
```

mehtaphysical:
Ohhh sorry. I didn't really consider the `AccountState` type. Is that the breaking change you were worried about @vgrichina?

I've updated this PR so that `AccountState` is still exported from account.ts preventing a breaking change.

I think i was worried about the `AccessKey` change because I didn't plan on having to consider that while  initially thinking about this change.

src/account.ts
```diff
@@ -393,24 +410,23 @@ export class Account {
...
         const accessKeys = await this.getAccessKeys();
-        const result: any = { authorizedApps: [], transactions: [] };
-        accessKeys.map((item) => {
-            if (item.access_key.permission.FunctionCall !== undefined) {
-                const perm = item.access_key.permission.FunctionCall;
-                result.authorizedApps.push({
-                    contractId: perm.receiver_id,
-                    amount: perm.allowance,
+        const authorizedApps = accessKeys
+            .filter(item => item.access_key.permission['FunctionCall'] !== undefined)
```
vgrichina, line 420: nit: seems like it makes sense to have `FunctionCall` as a field in a type used for `permission`

src/account_multisig.ts
```diff
@@ -175,48 +176,49 @@ export class Account2FA extends AccountMultisig {
...
-            access_key.permission.FunctionCall.method_names.length === 4 &&
-            access_key.permission.FunctionCall.method_names.includes('add_request_and_confirm')    
-        )
-        const confirmOnlyKey = PublicKey.from((await this.postSignedJson('/2fa/getAccessKey', { accountId })).publicKey)
+        const { accountId } = this;
+        const accessKeys = await this.getAccessKeys();
+        const lak2fak = accessKeys
+            .filter(({ access_key }) => access_key.permission['FunctionCall'] !== undefined)
+            .filter(({ access_key }) => {
+                const perm = (access_key.permission as FunctionCallPermissionView).FunctionCall;
```
vgrichina, line 193: I think it makes sense to change type for `permission` to define fields like `FunctionCall` so that there is no need to use `['FunctionCall']` and do a cast to `FunctionCallPermissionView`

test/providers.test.js
```diff
@@ -60,6 +59,102 @@ test('json rpc query account', withProvider(async (provider) => {
...
+        storage_paid_at: 0,
+    });
+}));
+
+test('json rpc query view_code', withProvider(async (provider) => {
+    const near = await testUtils.setUpTestConnection();
+    const account = await testUtils.createAccount(near);
+    const contract = await testUtils.deployContract(account, testUtils.generateUniqueString('test'));
+
+    return testUtils.waitFor(async() => {
```
vgrichina, line 110: not clear why `.wairFor` is needed vs jest's native timeout

src/account.ts
```diff
@@ -95,10 +93,14 @@ export class Account {
...
-     * @returns {Promise<AccountState>}
+     * @returns {Promise<AccountView>}
      */
-    async state(): Promise<AccountState> {
-        return await this.connection.provider.query(`account/${this.accountId}`, '');
+    async state(): Promise<AccountView> {
+        return this.connection.provider.query<AccountView>({
+            request_type: 'view_account',
+            account_id: this.accountId,
+            finality: 'optimistic'
```
vgrichina, line 102: are we sure that default was optimistic?

vgrichina (commented):
Looks good overall, but I have some minor comments.

src/account.ts
mehtaphysical: I believe so:

* using the older style falls back to this: https://github.com/near/nearcore/blob/master/chain/jsonrpc-primitives/src/types/query.rs#L146
* `Finality::None` is this https://github.com/near/nearcore/blob/master/core/primitives/src/types.rs#L23

test/providers.test.js
mehtaphysical: ah, i borrowed the idea from react testing library: https://testing-library.com/docs/dom-testing-library/api-async#waitfor

It waits for the condition to not throw an exception before passing the test. Its like what happens here: https://github.com/near/near-api-js/blob/master/test/providers.test.js#L93.

It seems like the type of thing that will come up repeatedly and i thought this may be a nice general solution.

src/account_multisig.ts
mehtaphysical: the trouble is that the field `permission` returned by the RPC is either an object or a string. We need to have some check to see which we have.

Could change it to: `typeof access_key.permission !== 'string'`

src/account.ts
```diff
@@ -393,24 +410,23 @@ export class Account {
...
         const accessKeys = await this.getAccessKeys();
-        const result: any = { authorizedApps: [], transactions: [] };
-        accessKeys.map((item) => {
-            if (item.access_key.permission.FunctionCall !== undefined) {
-                const perm = item.access_key.permission.FunctionCall;
-                result.authorizedApps.push({
-                    contractId: perm.receiver_id,
-                    amount: perm.allowance,
+        const authorizedApps = accessKeys
+            .filter(item => item.access_key.permission !== 'FullAccess')
```
mehtaphysical, line 420: Check if the permission is `FullAccess`, if not then it has to be `FunctionCallPermissionView`.

src/account_multisig.ts
vgrichina: @mehtaphysical got it. Now I understand it's unfortunate format choice on backend, so it's not really possible to avoid explicit cast for `permission`

vgrichina (approved):

mehtaphysical:
Can i merge this thing? Travis is failing because of fossa token issue @vgrichina

Issue #419, opened 2020-09-29
Contract class lacks API documentation


miohtama:
_**Is your feature request related to a problem? Please describe.**_ 

Contract class does not describe how it is used:

[https://github.com/near/near-api-js/blob/master/src/contract.ts#L15](https://github.com/near/near-api-js/blob/master/src/contract.ts#L15)

 _**Describe the solution you'd like**_ 

At minimum describe

1\. How methods for the contract are generated and why 2\. How methods are called 3\. What error handling the caller should be aware of when using `Contract` class 4\. How call the contract methods using different accounts (for testing)

PR #542, opened 2021-03-28
Fix/contract callback url ignored

### LINKED ISSUE #520: callbackUrl support is broken

I've been trying to work out how to use the Near Wallet's callbackUrl query arg, to pass my Near app's state across the wallet transaction without writing to the blockchain or local storage.  walletConnection.requestSignTransaction() has a callbackUrl argument specifically to support this.  But the only caller of that method is walletConnection.signAndSendTransaction(), which only allows the callbackUrl to be window.location.href, thereby blocking use of this feature.

One solution would be for the autogenerated contract methods on the contract object to accept an optional extra options argument -- an object which could specify callbackUrl or other useful things in the future.

**Describe alternatives you've considered**

I can try to hack/patch this in myself and send a pull req, if this is a desired feature.  But if this is deprecated, if there's some other preferable approach to the problem of passing client state through a Near wallet transaction, please point me at it.  Thanks!

**Additional context**

http://beta.plantary.art -- after purchase of an NFT we want to redirect the user to the link for that NFT.

mehtaphysical:
Fixes #520

Here is a proposed fix for `callbackUrl` and adding `meta`.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,15 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v3
+        with:
+          stale-pr-message: 'This PR is stale because it has been open for 7 days with no activity. Remove stale label or comment or this will be closed in another 7 days.'
+          days-before-pr-stale: 7
+          days-before-pr-close: 7
+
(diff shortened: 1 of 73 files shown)


mehtaphysical:
@vgrichina how does this look?

mehtaphysical:
Also fixes #419

mehtaphysical:
bump @vgrichina @mikedotexe @thisisjoshford

thisisjoshford:
> bump @vgrichina @mikedotexe @thisisjoshford

going to defer to @vgrichina and @mikedotexe on this one :)

volovyk-s (approved):

PR #551, opened 2021-04-08
bump bn version number to 5.1.0 to match types

### LINKED ISSUE #550: bn.js and @types/bn.js mis-versioned
`bn.js` and `@types/bn.js` have different versions. This causes an error when using `BN` in typescript: https://discord.com/channels/490367152054992913/714898996832895027/829525916380168215

mehtaphysical:
Fixes #550

### DIFF
--- a/lib/browser-connect.d.ts
+++ b/lib/browser-connect.d.ts
@@ -3,9 +3,7 @@
  *
  * {@link ConnectConfig.networkId} and {@link ConnectConfig.nodeUrl} are required.
  *
- * To sign transactions you can also pass:
- * 1. {@link ConnectConfig.keyStore}
- * 2. {@link ConnectConfig.deps | ConnectConfig.deps.keyStore}
+ * To sign transactions you can also pass: {@link ConnectConfig.keyStore}
  *
  * Both are passed they are prioritize in that order.
  *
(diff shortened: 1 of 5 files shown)


mehtaphysical:
Test fails because after updating BN the bundle size increased to ove 95kb:

```
FAIL dist/near-api-js.min.js: 99.54KB > 95KB (gzip)
```

mehtaphysical:
do we even use the browserified bundle? Could the size check and all the browserify stuff be removed?

vgrichina:
> do we even use the browserified bundle? Could the size check and all the browserify stuff be removed?

the check is useful even if we don't user browserified bundle, as we don't want near-api-js bundle footprint to grow

also judging by https://github.com/near/near-api-js/pull/563 some people want to have browserified bundle

vgrichina:
@mehtaphysical looks latest version is 5.2.0, is there any reason we don't want to use it?

also TBH I would really prefer to switch to `big.js` for cleaner API, but it seems like pretty large undertaking

mehtaphysical:
@vgrichina updated to v5.2.

Also, using `big.js` sounds like a good idea. Would be a great chance to make sure we are using numbers consistently throughout `near-api-js`.

thisisjoshford (approved):

PR #553, opened 2021-04-09
Feat/batch transactions wallet

### LINKED ISSUE #356: Send multiple TXs to Wallet to be signed
There are in many cases need to send a massive number of transactions. 

The current API of an Account is limited to single access key which means that needs to wait for transaction to process to avoid nonce collision.

First change would be to adjust `Signer` to also optionally take `public_key` and support multiple keys per account id.
 Add a way to retrieve list of public keys for a given account id.

The second is support batch actions, e.g.
 ```
 account.batch(<span class="error">[...]</span>.map(() => 

 { account.prepareFunctionCall(...) } 

));
 ```

or something, where it accumulates prepared transactions, signs them with available public key and sends them.

### LINKED ISSUE #504: Document usage of requestSignTransactions
`WalletConnection.requestSignTransactions` function has unclear signature, as it's requires transaction with signer_id and signer_pk. Developers are not sure what to fill in when signing a transaction.

We should document usage of this function in more detail, as these arguments are not required because wallet will be replacing them.

doriancrutcher:
Created Batch Transaction function to make it easier for  users to send multiple transactions at once without having to hand roll transactions, or having multiple redirections to the Near wallet 

fixes #504 
fixes #356

### DIFF
--- a/src/wallet-account.ts
+++ b/src/wallet-account.ts
@@ -217,7 +217,8 @@ export class WalletConnection {
             }
             const publicKey = PublicKey.from(accessKey.public_key);
         
-            const nonce = accessKey.access_key.nonce + index + 1;
+            const nonce = accessKey.access_key.nextNonce();
+            
             return createTransaction(accountId, publicKey, receiverId, nonce, actions, blockHash);
         }));
         
--- a/src/wallet-account.ts
+++ b/src/wallet-account.ts
@@ -22,6 +22,10 @@ const MULTISIG_HAS_METHOD = 'add_request_and_confirm';
 const LOCAL_STORAGE_KEY_SUFFIX = '_wallet_auth_key';
 const PENDING_ACCESS_KEY_PREFIX = 'pending_key'; // browser storage key for a pending access key (i.e. key has been generated but we are not sure it was added yet)
 
+interface TransactionOptions {
+    receiverId: string;
+    actions: Action[];
+}
 interface SignInOptions {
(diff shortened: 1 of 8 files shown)


vgrichina:
thanks, I think I'll need a bit more time to think on how to integrate with related changes I'm also making in https://github.com/near/near-api-js/pull/557/files

src/wallet-account.ts
```diff
@@ -169,6 +179,51 @@ export class WalletConnection {
...
+                            
+        const transactions: Transaction[] = await Promise.all(transactionOptions.map(async({ receiverId, actions }, index) => {
+            // TODO: Cache & listen for nonce updates for given access key
+            const accessKey = await this.account().accessKeyForTransaction(receiverId, actions, localKey);
+            if (!accessKey) {
+                throw new Error(`Cannot find matching key for transaction sent to ${receiverId}`);
+            }
+            const publicKey = PublicKey.from(accessKey.public_key);
+        
+            const nonce = accessKey.access_key.nonce + index + 1;
```
mehtaphysical, line 220: we actually need to increment the nonce here instead of using the index. Maybe:

```js
const nonce = ++accessKey.access_key.nonce
```

I wonder if we actually want a function to do this? like:

```js
const nonce = accessKey.access_key.nextNonce();
```

doriancrutcher: Yeah I see what you mean. The first one would definitely work

PR #555, opened 2021-04-15
remove node-fetch added to global in node env

### LINKED ISSUE #415: TypeError: fetch is not a function when trying to run Jest + NodeJs
**Describe the bug**

When using near-js-api in the jest + Node I get the following

``` FAIL  src/pool.test.js (6.979 s)
  ● Create pool

    TypeError: fetch is not a function

      at node_modules/near-api-js/lib/utils/web.js:48:36
      at Object.exponentialBackoff [as default] (node_modules/near-api-js/lib/utils/exponential-backoff.js:7:30)
      at Object.fetchJson (node_modules/near-api-js/lib/utils/web.js:46:57)
      at JsonRpcProvider.sendJsonRpc (node_modules/near-api-js/lib/providers/json-rpc-provider.js:139:38)
      at JsonRpcProvider.query (node_modules/near-api-js/lib/providers/json-rpc-provider.js:66:35)
      at Account.fetchState (node_modules/near-api-js/lib/account.js:47:54)
      at Account.get ready [as ready] (node_modules/near-api-js/lib/account.js:40:67)
      at Account.signAndSendTransaction (node_modules/near-api-js/lib/account.js:77:20)
      at Account.createAccount (node_modules/near-api-js/lib/account.js:195:21)
      at LocalAccountCreator.createAccount (node_modules/near-api-js/lib/account_creator.js:24:34)```
```

I assume the kludge here is not working with Jest test environment: https://github.com/near/near-api-js/blob/7fdc2131c1b509882122a9e923791fdf57525c01/src/utils/web.ts#L19

**To Reproduce**

```sh
git clone git@github.com:miohtama/advanced-fungible-token.git
git co jest-bug
yarn install 
npx jest --runInBand
```

**Expected behavior**

nest-js-api works in the test environment without modifications or it is documented how it is supposed to be run.

**Additional context**

Maybe just remove the in-house kludge and use one of the proven isomorphic fetch NPM packages that automatically do any kind of feature detection. Also then hav this package in `package.json` dependencies.

mehtaphysical:
Fixes #415 and removes a todo.

Adds the import of `node-fetch` into a separate module and attaches `node-fetch` to `global` only in the node import path.

### DIFF
--- a/lib/browser-connect.d.ts
+++ b/lib/browser-connect.d.ts
@@ -3,9 +3,7 @@
  *
  * {@link ConnectConfig.networkId} and {@link ConnectConfig.nodeUrl} are required.
  *
- * To sign transactions you can also pass:
- * 1. {@link ConnectConfig.keyStore}
- * 2. {@link ConnectConfig.deps | ConnectConfig.deps.keyStore}
+ * To sign transactions you can also pass: {@link ConnectConfig.keyStore}
  *
  * Both are passed they are prioritize in that order.
  *
@@ -22,7 +20,7 @@
  *
  * @module browserConnect
  */
-import { Near, NearConfig } from './near';
+import { Near, NearConfig } from "./near";
 export interface ConnectConfig extends NearConfig {
     /** @hidden */
     keyPath?: string;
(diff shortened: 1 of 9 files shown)


vgrichina:
@mehtaphysical 
> and attaches node-fetch to global only in the node import path.

not sure I understand the fix completely, shouldn't it also remove some old way of using `fetch`? maybe not all of the changed files were committed?

mehtaphysical:
Yes thanks. nice catch and sorry about that!

files pushed now

PR #557, opened 2021-04-16
Prototype support for generating wallet URLs outside of browser

vgrichina:


src/account.ts
```diff
@@ -123,27 +125,30 @@ export class Account {
...
-
-    /**
-     * Create a signed transaction which can be broadcast to the network
-     * @param receiverId NEAR account receiving the transaction
-     * @param actions list of actions to perform as part of the transaction
-     * @see {@link JsonRpcProvider.sendTransaction}
-     */
-    protected async signTransaction(receiverId: string, actions: Action[]): Promise<[Uint8Array, SignedTransaction]> {
+    
+    async createTransaction(receiverId: string, actions: Action[]): Promise<Transaction> {
```
volovyk-s, line 129: @vgrichina let's try to keep functions well documented

src/account.ts
```diff
@@ -153,7 +158,7 @@ export class Account {
      * @param receiverId NEAR account receiving the transaction
      * @param actions list of actions to perform as part of the transaction
      */
-    protected async signAndSendTransaction(receiverId: string, actions: Action[]): Promise<FinalExecutionOutcome> {
+    async signAndSendTransaction(receiverId: string, actions: Action[]): Promise<FinalExecutionOutcome> {
```
mehtaphysical, line 161: this is potentially a breaking change  as anyone extending the `Account` class with a custom implementation of `signAndSendTransaction` will have to remove the `protected` accessor.

This is fine, and something that is often requested anyways. But we need to make not of this before we publish this.

src/wallet-account.ts
```diff
@@ -66,18 +67,29 @@ export class WalletConnection {
...
+            try {
+                this._authData = JSON.parse(window.localStorage.getItem(this._authDataKey));
+            } catch (e) {
+                console.warn(`Error parsing WalletConnection JSON for ${this._authDataKey} key`);
+            }
+            this._authData = this._authData || { allKeys: [] };
+            if (!this.isSignedIn()) {
+                this._completeSignInWithAccessKey();
+            }
         }
```
mehtaphysical, line 93: should we throw a helpful error here if there is no `authData` and no `window`?

mehtaphysical:
I wonder if the `WalletConnection` code should be split into a few different classes:

* `abstract class BaseWalletConnection` that is used to create wallet urls and defines the interface
* `class WalletConnection extends BaseWalletConnection` that is used to open up urls in a browser window

This will allow us to customize the wallet experience better. For example, I could easily make a `class PopupWalletConnection extends BaseWalletConnection` class that opens up the wallet in a new window or tab. Or a `class PrintWalletConnection extends BaseWalletConnection` that `console.logs` the wallet urls.

src/account.ts
vgrichina: @mehtaphysical I don't think anybody besides wallet is doing it for now, do you know of any use cases? Also kinda trivial change for them anyway (and doesn't matter for JS at all, so it only affects peeps who rebuild from TS).

vgrichina:
> I wonder if the `WalletConnection` code should be split into a few different classes:
> 
> * `abstract class BaseWalletConnection` that is used to create wallet urls and defines the interface
> * `class WalletConnection extends BaseWalletConnection` that is used to open up urls in a browser window
> 
> This will allow us to customize the wallet experience better. For example, I could easily make a `class PopupWalletConnection extends BaseWalletConnection` class that opens up the wallet in a new window or tab. Or a `class PrintWalletConnection extends BaseWalletConnection` that `console.logs` the wallet urls.

I think it's great idea, especially if you want to take on it.

But it's out of scope for this PR I think.

src/account.ts
mehtaphysical: Agree. It seems pretty unlikely that someone is re-implementing this. Also, very true it only matters to TS projects.

@raymondfeng is the only person I know of who's been messing around with this. And he initially requested a change like  this anyways.

mehtaphysical: Also, I think this is a good change. Just need to make sure that we make note of it in the release notes.

mehtaphysical:
Fixes #568

mehtaphysical:
@vgrichina do you mind if i fix these merge conflicts?

zavodil:
I'm suffering without this feature.
@vgrichina @mehtaphysical could you resolve conflicts and push it please?

volovyk-s:
@mehtaphysical can you please resolve some of the conversations above? We need to merge this PR.

mehtaphysical:
@volovyk-s @zavodil. I'm not sure about this PR. I don't know about the use case here (though I get why you may want to generate wallet links).

I wonder if another approach is better. For example, maybe a `WalletLinkGenerator` class. Then the `WalletConnection` class can use the `WalletLinkGenerator` class to create wallet links. But other people, who just want the links, can use the `WalletLinkGenerator` class directly.

PR #561, opened 2021-04-17
update txStatus to take a string or a buffer

### LINKED ISSUE #560: Getting a transaction status requires an
Expect to be able to get a transaction status by its hash:

```js
const response = await near.connection.provider.txStatus(
   "AgrX82Q8SzHxSmMfcDtbDw54RMXpmU2TCme6emnXSFmM",
    "sender.testnet"
  );
```

But right now I have to first `bs58.decode` the hash:

```js
const response = await near.connection.provider.txStatus(
    bs58.decode("AgrX82Q8SzHxSmMfcDtbDw54RMXpmU2TCme6emnXSFmM"),
    "sender.testnet"
  );
```

So to use this a developer has to first decode the hash, then the provider will re-enoced the hash: const response = await near.connection.provider.txStatus(
    bs58.decode("AgrX82Q8SzHxSmMfcDtbDw54RMXpmU2TCme6emnXSFmM"),
    "sender.testnet"
  );

mehtaphysical:
Fixes #560 . This method is currently a bit confusing because it expects a `Uint8Array`, but the actual RPC method wants a hash:

```
{
  "jsonrpc": "2.0",
  "id": "dontcare",
  "method": "tx",
  "params": ["6zgh2u9DqHHiXzdy9ouTP7oGky2T4nugqzqt9wJZwNFm", "sender.testnet"]
}
```

This update makes it so you can pass `txStatus` a hash string or `Uint8Array`.


src/providers/json-rpc-provider.ts
```diff
@@ -98,10 +98,22 @@ export class JsonRpcProvider extends Provider {
      * @param txHash The hash of the transaction
      * @param accountId The NEAR account that signed the transaction
      */
-    async txStatus(txHash: Uint8Array, accountId: string): Promise<FinalExecutionOutcome> {
+    async txStatus(txHash: any, accountId: string): Promise<FinalExecutionOutcome> {
```
vgrichina, line 101: nit: why not `Uint8Array | string`?

also need to document what format is expected for string (is it base64 or base58?)

vgrichina (approved):
Looks good but can improve on type / docs

src/providers/json-rpc-provider.ts
mehtaphysical: Great point. I'll update

PR #563, opened 2021-04-19
fix: configure browser to use dist/near-api-js.min.js

raymondfeng:
This is needed to allow usage of near-api-js in html pages, such as:

```
<script src="https://cdn.jsdelivr.net/npm/near-api-js@0.39.0"></script>
```

Without this fix, it fails to load `near-api-js` as `exports` is not defined.

Signed-off-by: Raymond Feng <enjoyjava@gmail.com>

CC: @mattlockyer

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     },
     "homepage": "https://github.com/near/near-api-js",
     "main": "lib/index.js",
-    "browser": "lib/browser-index.js",
+    "browser": "dist/near-api-js.min.js",
     "types": "lib/index.d.ts",
     "dependencies": {
         "@types/bn.js": "^5.1.0",


mehtaphysical:
I think the issue with doing this is that it may also change how bundlers (e.g. webpack and parcel) behave with the library.

Two options here:

1. Classic JavaScript:

```html
<script src="https://cdn.jsdelivr.net/npm/near-api-js@0.39.0/dist/near-api-js.min.js"></script>
```

2. Module JavaScript

```html
<script type="module" src="https://cdn.jsdelivr.net/npm/near-api-js@0.39.0"></script>
```

raymondfeng:
I also tried `<script type="module" src="https://cdn.jsdelivr.net/npm/near-api-js@0.39.0"></script>` and it gives me the following error:

```
browser-index.js:24 Uncaught ReferenceError: exports is not defined
    at browser-index.js:24
```

vgrichina:
> I also tried `<script type="module" src="https://cdn.jsdelivr.net/npm/near-api-js@0.39.0"></script>` and it gives me the following error:
> 
> ```
> browser-index.js:24 Uncaught ReferenceError: exports is not defined
>     at browser-index.js:24
> ```

check this JS fiddle on how to use near-api-js as in `<script>` https://jsfiddle.net/uL7169px/6/

PR #570, opened 2021-04-28
Generated Error classes deleted

volovyk-s:
#closing https://github.com/near/near-api-js/issues/331


README.md
```diff
@@ -35,14 +35,13 @@ Start the node by following instructions from [nearcore](https://github.com/near
...
-# Update error messages
+# Update error schema
 
 Follow next steps:
 
 1. [Change hash for the commit with errors in the nearcore](https://github.com/near/near-api-js/blob/master/gen_error_types.js#L7-L9)
-2. Generate new types for errors: `node gen_error_types.js`
-3. `yarn fix` fix any issues with linter.
-4. `yarn build` to update `lib/**.js` files
+2. Fetch new schema: `node fetch_error_schema.js`
```
vgrichina, line 43: @volovyk-s why fetching schema is still needed?

vgrichina (approved):

README.md
volovyk-s: We still need it here to determine the error type here: https://github.com/near/near-api-js/blob/master/src/utils/rpc_errors.ts#L55

It should be deleted after `nearcore` starts returning the error type in their responses.

PR #572, opened 2021-04-30
2FA robustness improvements

vgrichina:
This is expected to make 2FA requests a bit faster (do not block for delete) and more robust (@mattlockyer's change fixes race condition).

### DIFF
--- a/src/account_multisig.ts
+++ b/src/account_multisig.ts
@@ -66,8 +66,6 @@ export class AccountMultisig extends Account {
         }
         await this.deleteUnconfirmedRequests();
 
-        const requestId = await this.getRequestNonce();
-        this.setRequest({ accountId, requestId, actions });
         const args = Buffer.from(JSON.stringify({
             request: {
                 receiver_id: receiverId,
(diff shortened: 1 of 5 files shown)


vgrichina:
@mehtaphysical this works fine with https://github.com/near/near-wallet/pull/1706

src/account_multisig.ts
```diff
@@ -53,36 +44,63 @@ export class AccountMultisig extends Account {
...
+                return await this.signAndSendTransaction(receiverId, actions);
+            }
+            throw e;
+        }
+
+        // TODO: Are following even needed? Seems like it throws on error already
+        if (!result.status) {
+            throw new Error('Request failed');
+        }
+        const status: any = Object.assign({}, result.status);
```
mehtaphysical, line 71: nit:

```suggestion
        const status: any = { ...result.status };
```

src/account_multisig.ts
```diff
@@ -209,9 +220,6 @@ export class Account2FA extends AccountMultisig {
     async sendCodeDefault() {
         const { accountId } = this;
         const { requestId, actions } = this.getRequest();
```
mehtaphysical, line 222: Fix linting:

```suggestion
        const { requestId } = this.getRequest();
```

mehtaphysical (approved):
Looks good. One nit and one linting change to fix tests.

Also looks like there is a merge conflict

mehtaphysical:
@vgrichina I forgot, there was a change so we require 2 approvals now

vgrichina:
> @vgrichina I forgot, there was a change so we require 2 approvals now

but why?

mehtaphysical:
@mikedotexe can probably add more context about this change.

PR #577, opened 2021-05-04
Adding keyPath uses that keystore first

mehtaphysical:
See #469

From @mikedotexe 

> I've noticed this on and off and finally figured out it's a one-liner.
When I create a function-call access key for an account and then specify it as the keyPath, the full access key was being used.
When a person specifies the keyPath, I believe that means, "Please use this keystore first, as I'm specifying to use it." Instead it's added to the end of the MergeKeyStore, so it's the last keystore to be checked.
This simply flips that so the first keystore used is the one specified by keyPath.

Keep in mind this comment: https://github.com/near/near-api-js/pull/469#issuecomment-772398493

> keyPathStore is an InMemoryKeyStore. It gets precedence so all new keys are written to it. So if you e.g. try to create new account while specifying keyPath to provide key for you masterAccount you would lose a key to newly created account.

### DIFF
--- a/src/key_stores/merge_key_store.ts
+++ b/src/key_stores/merge_key_store.ts
@@ -34,17 +34,22 @@ import { KeyPair } from '../utils/key_pair';
  * const near = await connect(config)
  * ```
  */
+
+interface MergeKeyStoreOptions {
+    writeKeyStoreIndex: number;
+}
+
 export class MergeKeyStore extends KeyStore {
-    private writeIndex: number;
+    private options: MergeKeyStoreOptions;
     keyStores: KeyStore[];
 
     /**
      * @param keyStores read calls are attempted from start to end of array
-     * @param writeIndex the keystore index that will receive all write calls
+     * @param options.writeKeyStoreIndex the keystore index that will receive all write calls
      */
-    constructor(keyStores: KeyStore[], writeIndex = 0) {
+    constructor(keyStores: KeyStore[], options: MergeKeyStoreOptions) {
         super();
-        this.writeIndex = writeIndex;
+        this.options = options;
         this.keyStores = keyStores;
     }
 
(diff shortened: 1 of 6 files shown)


src/key_stores/merge_key_store.ts
```diff
@@ -35,13 +35,16 @@ import { KeyPair } from '../utils/key_pair';
...
+    private writeIndex: number;
     keyStores: KeyStore[];
 
     /**
-     * @param keyStores first keystore gets all write calls, read calls are attempted from start to end of array
+     * @param keyStores read calls are attempted from start to end of array
+     * @param writeIndex the keystore index that will receive all write calls
      */
-    constructor(keyStores: KeyStore[]) {
+    constructor(keyStores: KeyStore[], writeIndex = 0) {
```
vgrichina, line 45: I think we should pass `writeKeyStore` instance explicitly instead of passing index

vgrichina: as if there is a number passed as an argument I have to go an look to docs, etc to understand what this number represents vs if you pass keystore directly it seems to be more intuitive 

though maybe it's just me

mehtaphysical: yes good call.

mehtaphysical: Ok, I've experimented with this a bit. It may be a bit awkward because it makes it so the `writeKeyStore` may not be in the list of `keyStores`:

We can check for this:

```ts
config.keyStore = new MergeKeyStore([keyPathStore, (config.keyStore || config.deps.keyStore)], keyPathStore);
```

```ts
    constructor(keyStores: KeyStore[], writeKeyStore?: KeyStore) {
        super();
        // probably best to check that writeKeyStore is in keystores
        if(!keyStores.some(ks => ks === writeKeyStore)) {
            throw new Error('keyStores is missing writeKeyStore');
        }

        this.writeKeyStore = writeKeyStore || keyStores[0];
        this.keyStores = keyStores;
    }
```

Maybe the best solution is to make the second param to the constructor an options object so we can:
 
```ts
config.keyStore = new MergeKeyStore([keyPathStore, (config.keyStore || config.deps.keyStore)], { writeKeyStoreIndex: 1 });
```

vgrichina (approved):
lgtm, solution with options explains argument better and gives us flexibility in future

PR #579, opened 2021-05-05
Update to typescript 4.2

mehtaphysical:
Seems like this is relatively safe. Not `.d.ts` files change after this update.

### DIFF
--- a/lib/browser-index.js
+++ b/lib/browser-index.js
@@ -14,14 +14,15 @@ var __setModuleDefault = (this && this.__setModuleDefault) || (Object.create ? (
 var __importStar = (this && this.__importStar) || function (mod) {
     if (mod && mod.__esModule) return mod;
     var result = {};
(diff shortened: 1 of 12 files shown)


vgrichina:
@mehtaphysical looks like the only concern here is users who themselves use older version of TypeScript with near-api-js. This has previously caused some problems, so just wanna make sure it's on your radar.

volovyk-s:
@mehtaphysical let's postpone this merge for a few months, there is some breaking changes in this version: https://devblogs.microsoft.com/typescript/announcing-typescript-4-2/#breaking-changes
We should give developers time to upgrade their TS version first

mikedotexe:
Put a calendar event for us to revisit in a month.

volovyk-s:
@mehtaphysical @mikedotexe month has passed.

volovyk-s:
@mehtaphysical let's fix conflicts here and merge this PR.

PR #594, opened 2021-05-18
Add Logger class and configuration

### LINKED ISSUE #564: Add option to remove console logs
Near-api-js logs a transaction status. It feels more like debug info that has been left in. 

Add option that allows you to set debugging logs

**Additional context**
Screenshot 2021-04-20 at 11 21 47

mehtaphysical:
Fixes #564


src/account_multisig.ts
```diff
@@ -110,7 +111,7 @@ export class AccountMultisig extends Account {
                 await super.signAndSendTransaction(this.accountId, 
                     [functionCall('delete_request', { request_id: requestIdToDelete }, MULTISIG_GAS, MULTISIG_DEPOSIT)]);
             } catch(e) {
-                console.warn('Attempt to delete an earlier request before 15 minutes failed. Will try again.');
+                getLogger().debug('Attempt to delete an earlier request before 15 minutes failed. Will try again.');
```
volovyk-s, line 114: Looks like we should check `errorType` here :/ Not a blocker for this PR.
Why `debug` instead of `warn`?

src/utils/web.ts
```diff
@@ -33,15 +34,15 @@ export async function fetchJson(connection: string | ConnectionInfo, json?: stri
             });
             if (!response.ok) {
                 if (response.status === 503) {
-                    console.warn(`Retrying HTTP request for ${url} as it's not available now`);
+                    getLogger().debug(`Retrying HTTP request for ${url} as it's not available now`);
```
volovyk-s, line 37: Looks like a `warning`.

src/utils/web.ts
```diff
@@ -33,15 +34,15 @@ export async function fetchJson(connection: string | ConnectionInfo, json?: stri
...
                     return null;
                 }
                 throw createError(response.status, await response.text());
             }
             return response;
         } catch (error) {
             if (error.toString().includes('FetchError') || error.toString().includes('Failed to fetch')) {
-                console.warn(`Retrying HTTP request for ${url} because of error: ${error}`);
+                getLogger().debug(`Retrying HTTP request for ${url} because of error: ${error}`);
```
volovyk-s, line 45: This one too looks like a warning.

test/account.test.js
```diff
@@ -116,6 +117,8 @@ describe('with deploy contract', () => {
         const newPublicKey = await nearjs.connection.signer.createKey(contractId, testUtils.networkId);
         const data = [...fs.readFileSync(HELLO_WASM_PATH)];
         await workingAccount.createAndDeployContract(contractId, newPublicKey, data, HELLO_WASM_BALANCE);
+        setLogLevel(LogLevel.INFO);
```
volovyk-s, line 120: Should we do the same in all test files?

volovyk-s (commented):
Looks cool. Have you tested if Logging levels are set outside of `near-api-js`?

src/utils/logger.ts
```diff
@@ -0,0 +1,86 @@
...
+  constructor(logLevel: LogLevel = LogLevel.WARN) {
+      this.logLevel = logLevel;
+  }
+
+  setLogLevel(logLevel: LogLevel) {
+      this.logLevel = logLevel;
+  }
+
+  deprecate(moduleName: string, msg: string) {
+      if(this.shouldLog(LogLevel.ERROR)) {
```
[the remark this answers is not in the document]
mehtaphysical: This makes it so if the `LogLevel` is at least `ERROR` that the deprecation messages will show.

`LogLevel` let's a developer configure hiding some logs.

src/account.ts
```diff
@@ -119,8 +119,7 @@ export class Account {
 
     /** @hidden */
     protected get ready(): Promise<void> {
-        const deprecate = depd('Account.ready()');
-        deprecate('not needed anymore, always ready');
+        getLogger().deprecate('Account.ready()', 'not needed anymore, always ready');
```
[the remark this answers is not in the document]
mehtaphysical: that could be interesting.

src/utils/web.ts
mehtaphysical: I chose debug for this one because it automatically retries. Most developers probably won't care.

src/utils/web.ts
mehtaphysical: Same as above. Because it automatically retries it seems like a debug message. Then if after the retries it fails it should log an error message

test/account.test.js
mehtaphysical: hmm. I'm not sure what is better. This one needed it because we are checking that some logs are printed.

But right now the tests print too much info, so I sort of like this only here.

src/account_multisig.ts
mehtaphysical: i think you're right.

src/account.ts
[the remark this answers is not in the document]
volovyk-s: @thisisjoshford what is your opinion on this?
Also, should we add some examples of how users should use it during development here in README or somewhere else?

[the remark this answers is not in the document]
mehtaphysical: I wonder if this should be in a different PR. I bet @MaximusHaximus has some thoughts on deprecation messaging as well.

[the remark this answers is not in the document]
volovyk-s: Ok, let's not block it. But I would love to hear some opinions here,

MaximusHaximus (changes requested):
Comments in-line -- I don't necessarily want to block the PR but I think we should be careful about tight coupling between a logger and `depd` as coupling them together makes it a permanent requirement that we must manage the interactions between them.  I think of deprecation logging as a wholly independent thing from application logging, and I feel like the fact that `depd` has it's own API for controlling deprecation logging means that people really should be using `depd`'s API for this rather than combining the concerns into a single logging variable everywhere :). We just would need to explicitly document the fact that we use `depd` for deprecation logging in our e.g. README.md and explain that it can be controlled by the `depd` API, but for application-level logs, they are controlled by logLevel.

Also, I think it would be nice if the global singleton instance would _default_ its logLevel to whatever is in e.g. `process.env.NEAR_API_LOG_LEVEL`, if it has been set, so people could toggle the value without having to write code (lets you use different .env files to control verbosity and inject it into an e.g. CI pipeline _externally_, without needing to add calling JS methods on our `near-api-js` logger singleton into the consuming codebase.

volovyk-s:
@mehtaphysical do you have time to finish it? It's getting staler and staler.

mikedotexe:
Feels like there's a lot of great comments here from Daryl. Seeing as this isn't blocking critical functionality, are we down to take our time and address these? @MaximusHaximus humongous thank you for being so thorough and thoughtful here, holy smokes

mehtaphysical:
I think we should close this. It seems like it should be a part of a larger design decision.

volovyk-s:
@mehtaphysical ok, let's do the design work first.

Issue #593, opened 2021-05-18
u3 package is broken


vans163:
```
[!] (plugin commonjs) SyntaxError: Deleting local variable in strict mode (8:8) in 
../node_modules/u3/lib/cache.js
node_modules/u3/lib/cache.js (8:8)
```
It seems the u3 package used is outdated and does not follow modern javascript rules.

mehtaphysical:
It doesn't seem like this codebase uses u3. It's not in our `yarn.lock`. Are you maybe using a different dependency that uses it?

thanhlmm:
@vans163 I face the same problem, so I need to override the `u3` package by using `patch-package`

Here is the modified one
```
var cache = function (fn) {
    var called = false,
        store;

    if (!(fn instanceof Function)) {
        called = true;
        store = fn;
        fn = null;
    }

    return function () {
        if (!called) {
            called = true;
            store = fn.apply(this, arguments);
            fn = null;
        }
        return store;
    };
};

module.exports = cache;
```

volovyk-s:
@mehtaphysical looks like it is in our lock file and dependency tree. Try `npm list --depth=1` to find it. It's a dependency of  `error-polyfill`, but both `error-polyfill` and `u3` have the latest versions from `npm`.

mehtaphysical:
@volovyk-s I see. We should consider removing this dependency. According to their GitHub repo it is:

> INACTIVE (unstable) - Javascript Error Polyfill - Implementing the V8 Error API in non-V8 environments as much as possible

Issue #119, opened 2019-11-27
Audit dependencies and remove unused ones. 


janedegtiareva:

vgrichina:
This is something I looked into before

nearlib has only 8 direct dependencies (http://npm.broofa.com/?q=nearlib), they are all used. 
http-errors (and by extension error-polyfill) can potentially be removed – we can just reimplement `createError`, but they also are relatively trivial in size.

PR #598, opened 2021-05-19
Error polyfill deletion

volovyk-s:
@MaximusHaximus we are having some troubles with this dependency, do we still need it in the wallet? Looks like it's not used in `near-api-js` internally.
Details here: https://github.com/near/near-api-js/issues/593

### DIFF
--- a/lib/account.d.ts
+++ b/lib/account.d.ts
@@ -14,7 +14,7 @@ export interface AccountBalance {
 export interface AccountAuthorizedApp {
     contractId: string;
     amount: string;
-    publicKey: PublicKey;
+    publicKey: string;
 }
 /**
  * Options used to initiate sining and sending transactions
--- a/lib/account_multisig.js
+++ b/lib/account_multisig.js
@@ -173,8 +173,8 @@ class Account2FA extends AccountMultisig {
         const confirmOnlyKey = key_pair_1.PublicKey.from((await this.postSignedJson('/2fa/getAccessKey', { accountId })).publicKey);
         const actions = [
             transaction_1.deleteKey(confirmOnlyKey),
-            ...lak2fak.map(({ public_key }) => transaction_1.deleteKey(public_key)),
-            ...lak2fak.map(({ public_key }) => transaction_1.addKey(public_key, null)),
+            ...lak2fak.map(({ public_key }) => transaction_1.deleteKey(key_pair_1.PublicKey.from(public_key))),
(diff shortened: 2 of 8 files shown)


MaximusHaximus:
@volovyk-s thanks for the heads up -- I did some digging through git history and discovered that this polyfill exists due to usage of `error.captureStackStrace()` inside of the `http-errors` package, which is a package made explicitly for nodeJS  usage (thus the captureStackTrace() usage...), used for composing http errors (basically, errors with `status`/`statusCode` properties, but also some other API details around attaching custom properties).

Here's the problem in a nutshell:
https://github.com/jshttp/http-errors/issues/28

All that we currently use from `http-errors` _directly_ in the wallet and `near-api-js` is the `createError()` method, which we could replace with our own implementation with guarding around the `captureStackTrace()` functionality for backwards compat for browsers.  We use the _simplest_ form of `createError()`, so all we are really getting is the equivalent of creating an error and setting `code` and `statusCode` properties on it in one call.

There's even mention of doing this in this (sadly, closed) issue:
https://github.com/near/near-api-js/issues/119

it's worth nothing that both `near-wallet` and `near-api-js` also require `http-errors` dependency _indirectly_, by way of `fetch-send-json` package as well, which is a simple wrapper around fetch that we also own (https://github.com/near/send-json).

Since we want `near-api-js` to be browser-friendly, and it's clear from the issue I linked above that `http-errors` package does not intend to add compatibility for browsers (or react native -- see https://github.com/jshttp/http-errors/issues/74 also), I think it makes sense for us to have a browser-friendly version of `http-errors` that wraps the `captureStackTrace()` logic for older browsers.

However, `error-polyfill` does do some formatting/parsing of stack traces inside of its provided `captureStackTrace()` method, so removing this could negatively effect our stack trace content in older browsers if I am reading the code therein correctly.

**Most importantly, by removing this polyfill, we probably broke old browsers being able to use the `near-api-js` since we still have the dependency on `http-errors` present.**

MaximusHaximus (changes requested):
Comment added by mistake instead of review :)

vgrichina:
@volovyk-s +1 to @MaximusHaximus comments, it's not about wallet, but about browser support

note that it doesn't seem hard to remove http-errors dependency from all of these packages (they are all ours) but it has to happen before removing polyfill

volovyk-s:
@MaximusHaximus thanks for a detailed answer!
@vgrichina ok, sounds like a plan. Meanwhile, I'm converting this PR to `Draft`. 
@mikedotexe  I would say that we should prioritize this issue since it creates problems for developers.

PR #599, opened 2021-05-20
Use TypeScript files for tests

mehtaphysical:
This makes the tests a bit faster to run (you don't have to compile the TS to JS before hand).

Also, this opens the door to unit tests that we can run in watch mode.


volovyk-s:
I'm wondering if it's ok to test TS code in `src` folder instead of `lib` folder that is used by the majority of the developers. Is it common to have JS-specific bugs when we are generating JS code with `tsc`?

thisisjoshford:
Curious about your thoughts on @volovyk-s question @mehtaphysical. Are there any potential bugs we could miss from skipping the compiling process to JS from TS?

thisisjoshford (approved):

volovyk-s (approved):

PR #600, opened 2021-05-20
remove deprecated calls

### LINKED ISSUE #597: Don't use deprecated functions in `near-api-js`
we are still using deprecated versions if `contract.methodName` and `account.signAndSendTransaction` in `near-api-js` code base

mehtaphysical:
Fixes #597

### DIFF
--- a/test/account.test.js
+++ b/test/account.test.js
@@ -135,7 +135,10 @@ describe('with deploy contract', () => {
     });
 
     test('cross-contact assertion and panic', async () => {
-        await expect(contract.crossContract({}, 300000000000000)).rejects.toThrow(/Smart contract panicked: expected to fail./);
+        await expect(contract.crossContract({
+            args: {},
+            gas: 300000000000000
+        })).rejects.toThrow(/Smart contract panicked: expected to fail./);
         expect(logs.length).toEqual(7);
         expect(logs[0]).toMatch(new RegExp('^Receipts: \\w+, \\w+, \\w+$'));
         //	Log [test_contract1591458385248117]: test_contract1591458385248117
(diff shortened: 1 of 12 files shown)


test/account.test.js
```diff
@@ -185,14 +196,17 @@ describe('with deploy contract', () => {
...
         expect(result2).toEqual(setCallValue);
         expect(await contract.getValue()).toEqual(setCallValue);
     });
 
     test('make function calls via contract with gas', async() => {
         const setCallValue = testUtils.generateUniqueString('setCallPrefix');
-        const result2 = await contract.setValue({ value: setCallValue }, 1000000 * 1000000);
+        const result2 = await contract.setValue({
+            args: { value: setCallValue },
+            gas: 1000000 * 1000000
```
thisisjoshford, line 208: Curious why we are doing a calculation here rather than just passing the actual amount.

thisisjoshford (approved):
Tests are passing and everything LGTM. Left one comment on the `account.test.js` file.

test/account.test.js
volovyk-s: It's easier to count zeros this way!
It's even more useful when we are talking about time. Like, `24*60*60` is better than `86400`.

test/promise.test.js
```diff
@@ -63,17 +68,22 @@ describe('with promises', () => {
...
-            args: null,
-            gas: '3000000000000',
-            balance: '0',
-            callback: 'callbackWithName',
-            callbackArgs: null,
-            callbackBalance: '0',
-            callbackGas: '2000000000000', 
-        }}, CONTRACT_CALL_GAS);
+        const realResult = await contract.callPromise({
+            args: {
```
volovyk-s, line 72: `args: { args: {` does not look good :/ but let's not block this change

volovyk-s (approved):

PR #603, opened 2021-05-22
fix: jsonrpc empty string response should success

ailisp:
Right now if a jsonrpc returns this result:
```
{"jsonrpc":"2.0","result":"","id":1}
```
 near-api-js consider it as fail and retry until timeout. This behavor is incorrect. According to jsonrpc spec, Valid success jsonrpc can respond empty result, such as `""` or `null`, for example in sandbox_patch_state rpc and adversarial rpcs. Valid error response should be determined by `result` key not present or `error` key is present.

Also exponentialBackoff won't return if result is "" or null, so make this return entire json rpc response to avoid it keep retry on `{..., result: ""}` response and only retry on real error response and fetch errors

### DIFF
--- a/src/providers/json-rpc-provider.ts
+++ b/src/providers/json-rpc-provider.ts
@@ -379,7 +379,7 @@ export class JsonRpcProvider extends Provider {
                 throw error;
             }
         });
-        if (!result && result !== "") {
+        if (result === undefined) {
             throw new TypedError(
                 `Exceeded ${REQUEST_RETRY_NUMBER} attempts for request to ${method}.`, 'RetriesExceeded');
         }
--- a/src/providers/json-rpc-provider.ts
+++ b/src/providers/json-rpc-provider.ts
@@ -379,7 +379,7 @@ export class JsonRpcProvider extends Provider {
                 throw error;
             }
         });
-        if (!result) {
+        if (!result && result !== "") {
             throw new TypedError(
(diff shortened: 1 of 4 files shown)


ailisp:
This still doesn't work as exponentialBackoff won't return when result is "" or null. I'll think of a better way to do this.

volovyk-s (approved):

src/providers/json-rpc-provider.ts
```diff
@@ -379,7 +380,12 @@ export class JsonRpcProvider extends Provider {
                 throw error;
             }
         });
-        if (!result) {
+        const result = response.result;
```
vgrichina, line 383: nit: `const { result } = response`

vgrichina (approved):

vgrichina:
@ailisp looks good to me, but ideally also needs regression test

ailisp:
> @ailisp looks good to me, but ideally also needs regression test

sounds good, i'll make one

ailisp:
@vgrichina Unfortunately all RPCs enabled for ci node and ci-betanet node do not have any empty string result. RPC `sandbox_*` and `adversarial_*` have some empty result RPC but they're not enabled for ci or testnet node. In code snippet on this doc i tested near-api-js of this branch which shows after this change sendJsonRpc can proceed with empty result.

PR #606, opened 2021-05-25
add quick reference snippets

thisisjoshford:


volovyk-s (approved):
Looks cool!
What do you think about moving `Contribute to this library` down and have the `Examples` section right after the `Docs`?

mehtaphysical:
Instead of an `md` file these should be in `js` files so they are runable. That will allow us to test against them (we can/should add tests that run the examples to make sure they still work).

vgrichina:
> Instead of an `md` file these should be in `js` files so they are runable. That will allow us to test against them (we can/should add tests that run the examples to make sure they still work).

I agree it would be nice to generate such reference from runnable files, but seems like this is a separate sizeable chunk of work and so shouldn't block this PR.

vgrichina (approved):

thisisjoshford:
> What do you think about moving `Contribute to this library` down and have the `Examples` section right after the `Docs`?

Will do!

>Instead of an md file these should be in js files so they are runable. That will allow us to test against them (we can/should add tests that run the examples to make sure they still work).

@mehtaphysical sounds like a good plan... Going to merge this one now and create a new issue to break these up and get to it next sprint.

PR #619, opened 2021-06-08
updat error to NotEnoughAllowance from NotEnoughBalance

mehtaphysical:
Discovered during support in discord. Matches the NEPs: https://github.com/near/NEPs/blob/17af66e32a83e0bfea4e8110b59902daf09e7ab4/specs/RuntimeSpec/Transactions.md#validation-with-state

### DIFF
--- a/src/wallet-account.ts
+++ b/src/wallet-account.ts
@@ -307,7 +307,7 @@ export class ConnectedWalletAccount extends Account {
             try {
                 return await super.signAndSendTransaction({ receiverId, actions });
             } catch (e) {
-                if (e.type === 'NotEnoughBalance') {
+                if (e.type === 'NotEnoughAllowance') {
                     accessKey = await this.accessKeyForTransaction(receiverId, actions);
                 } else {
                     throw e;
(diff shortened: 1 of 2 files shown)


src/wallet-account.ts
```diff
@@ -307,7 +307,7 @@ export class ConnectedWalletAccount extends Account {
             try {
                 return await super.signAndSendTransaction({ receiverId, actions });
             } catch (e) {
-                if (e.type === 'NotEnoughBalance') {
+                if (e.type === 'NotEnoughAllowance') {
```
volovyk-s, line 310: @mehtaphysical do you think we should add a simple explanation of what is happening in this piece of code? It's not obvious why we are checking for `NotEnoughAllowance` or `NotEnoughBallance` here.

mikedotexe: I'm down with adding comments, too. And thanks for jumping on this issue

volovyk-s (approved):

mikedotexe (approved):

PR #620, opened 2021-06-10
Add support for borsh in view functions

mfornet:


src/account.ts
```diff
@@ -499,12 +499,15 @@ export class Account {
         { parse = parseJsonFromRawResponse } = {}
     ): Promise<any> {
         this.validateArgs(args);
+
+        const isUint8Array = args.byteLength !== undefined && args.byteLength === args.length;
```
mfornet, line 503: This way to detect if the type of `isUint8Array` was copied from other place in `near-api-js`. Some thoughts on this:

- We may want to put this function inside `utils` and reuse it elsewhere
- It is not very robust. For example, it detects Buffer (which "inherits" from Uint8Array), but it fails to detect properly `Uint8Array`. So we may as well call it `isBuffer`.

volovyk-s: Agree, let's not copy-paste it and move to a separate util.
I would not call args `serializedArgs` before they actually serialized.

src/account.ts
```diff
@@ -499,12 +499,15 @@ export class Account {
         { parse = parseJsonFromRawResponse } = {}
```
MaximusHaximus, line 499: Rather than have conversion happening implicitly based on duck-typing, I think it would be preferable to have the choice of how for `viewFunction` handles the stringification of `args` be more explicit.

Currently, `args` can be _anything_, but in the end, the only critical, consistent property of `args` when we post it is that it _must_ be a `base64` encoded string.  Whether we should `JSON.stringify()` the thing or not is unknown to us; we can make assumptions about it, but we can be wrong (as we are in the use case for Uint8Array in some cases).

We've added a bit of convenience logic to this method that assumes that we should `JSON.stringify()` whatever we are given _before_ Base64 encoding it, but we've discovered that stringifying a Uint8Array before b64 encoding isn't the behavior we want in some cases.

Rather than assume that a particular set of properties on the data that we're given should disable `JSON.stringify()` of that object, which makes it our responsibility to maintain this logic and adds brittle ducktyping behavior to the internals of this method, I think it would be preferable to allow the caller to directly, explicitly control how `args` is converted to a string.  While the ducktyping approach fixes Uint8Array, what about the next developer who says, "I just have a _raw_ I want to submit as `args` -- it's _already valid JSON_,  but it gets `JSON.stringify()`'d which double escapes everything in the string?!".  Next we'll be adding a typecheck for 'did we receive a string object? And who knows how many other cases we cannot account for (like, what if someone _else_ actually _wants_ a Uint8Array to be sent in JSON for whatever reason?).

I propose that, just as the caller can provide a `parse` option as the last argument here to control the handling of the viewFunction _response_, we add a new options property here called e.g. `stringify`, that lets the caller control how we serialize the args any way they want to.  This new options property could be _defaulted_ to `(args => JSON.stringify(args)` for backwards compatibility, but can be overridden to be any function (like `(args) => args` in cases where we, as a caller, don't want to have our provided args run through `JSON.stringiify()` before being converted to a base64 string.

Here's are some simple examples of how this approach allows a consumer of our API to make simple wrappers for their use cases that can disable all JSON assumptions (both on submitting args and on receiving back the response from the viewFunction call) -- a `raw` submission/response -- or alternatively, can allow them to easily override either behavior by providing an `options` argument on a per-call basis if they wanted just by changing where they spread the options pass into their wrapper.  We could even publish the `viewFunctionRaw()` method below as part of our API if we wanted to make it more discoverable to people, and document an example use cases (using Uint8Array) for when someone might use `viewFunctionRaw()`.

```javascript
// No JSON stringification of args and no expectation that we got JSON back
function viewFunctionRaw (
  account,
  contractId,
  methodName,
  args,
  opts = {}
) {
 return account.viewFunction(
    contractId,
    methodName,
    args,
    {
      ...opts, // provided opts can NOT override `stringify` or `parse`
      stringify: (args) => args,
      parse: (response) => response,
    }
  );
}

// DEFAULT to no JSON stringification of args and no expectation that we got JSON back, but allow caller to override on a per call basis
function viewFunctionDefaultRaw (
  account,
  contractId,
  methodName,
  args,
  opts = {}
) {
 return account.viewFunction(
    contractId,
    methodName,
    args,
    {
      stringify: (args) => args,
      parse: (response) => response,
      ...opts, // Provided opts can override `stringify` and `parse`
    }
  );
}
```

In this way, we have relinquished responsibility to make decisions about the stringification of `args` based on (arbitrary) object shapes, and empowered the user of our code to do whatever they want to without having to worry about our implicit behaviors and assumptions.  If a developer _happens_ to have an object with `byteLength` that matches `args.length` that they _do_ want `JSON.stringify()`'d, they can have that behavior, just like the developer who has the same object but does _not_ want `JSON.stringify()`'d can do.  Developers can provide any stringification method they want to any call, which allows them to easily define custom stringifiers in their codebase that can be composed with any viewFunction calls they use (easily unit tested), etc.

What do you guys think?

PR #623, opened 2021-06-14
feat: prepare tests for the stabilization of `EXPERIMENTAL_tx_status` method

frol:
See this failing test: https://buildkite.com/nearprotocol/nearcore/builds/1547#2c89c279-4fd8-490d-8235-86531597f5d3

image

`tx` method was extended with one more field (`receipts`), which is not a breaking change, but the output of the broadcast_tx_commit does not return `receipts` (only returns `status`, `transaction`, `transaction_outcome`, and `receipts_outcome`), so we need to check if the response from `broadcast_tx_commit` is a strict subset of the `tx` response.


volovyk-s (approved):

MaximusHaximus (approved):
:+1: :shipit:

PR #629, opened 2021-06-29
Shouldn't check if account exists if `contractId` not passed

vgrichina:
This allows app to request login (and get account name of the user) without adding access key for a contract.


volovyk-s (approved):

MaximusHaximus (approved):
:+1: legit bugfix is legit.

PR #631, opened 2021-07-01
Release files whitelisted

volovyk-s:
Our releases on NPM contain config files, tests, and example code snippets. Let's use whitelisting instead of `.npmignore` file.

I'm wondering if we should remain `src` folder. Can we assume that everybody is using `lib` only?
`src` folder is ~180 kByte's, it would be cool to get rid of it and reduce loading time in all our apps.

Also, it's not clear for me how  `dist/near-api-js.min.js` works and how it should be used. There are no mentions of it in our docs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,5 +69,9 @@
             }
         ]
     },
+    "files": [
+        "lib",
+        "browser-exports.js"
+    ],
     "author": "NEAR Inc"
 }
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     },
     "files": [
         "lib",
+        "dist",
         "browser-exports.js"
     ],
     "author": "NEAR Inc"
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,0 @@
-storage
-node_modules
-
-# near-api-js files & coverage
-neardev/
-coverage/


MaximusHaximus:
@volovyk-s

1. The `dist/near-api-js.min.js` exists so that people writing sites that use `near-api-js` can use a static, minified version of the package as a dependency on their pages, rather than making it part of a transpiled bundle and needing to minify it themselves; I think it should remain.

2. IMO, the `src` directory should _not_ be in the NPM package at all, because it is TS code, which is not generally expected to be in the distributed NPM package (NPM package is for consuming, github repo clone is for developing on, and we distribute .d.ts files for TS support). Also, TS server generally does not transpile anything in `node_modules` without custom configurations (not a best practice) and our `package.json`'s `main` entry points into `lib` anyway, so it would take some hacking to even get the code in the `src` directory to run via an NPM dependency.

volovyk-s:
Ok, so `dist` folder belongs here to. @MaximusHaximus Please, review.

volovyk-s:
@MaximusHaximus do we need to keep `dist/near-api-js.js` if we already have `dist/near-api-js.min.js`? Is it used for the debug?

MaximusHaximus:
> @MaximusHaximus do we need to keep `dist/near-api-js.js` if we already have `dist/near-api-js.min.js`? Is it used for the debug?

Good question... debugging is the only reason I can think of to have the un-minified file there, **but**, I think that we could get rid of `dist/near-api-js.js` if we start publishing sourcemaps for `dist/near-api-js.min.js` instead -- easily done by using the `--debug` flag for `browserify` AFAIK (ref: https://github.com/hughsk/uglifyify#source-maps)

mikedotexe:
Restarted one of the tests that had failed. It was for betanet, and I'm going to assume this was failing when betanet was down recently.

mikedotexe:
I do think we'll want to make sure we do a major version release in case folks were relying or using some of the files that'll no longer be there

package.json
```diff
@@ -69,5 +69,10 @@
             }
         ]
     },
+    "files": [
+        "lib",
+        "dist",
```
mikedotexe, line 74: probably `dist/near-api-js.min.js` eh?

mikedotexe (approved):

volovyk-s:
> I do think we'll want to make sure we do a major version release in case folks were relying or using some of the files that'll no longer be there

Do you want to merge it and release 1.0? Or wait until we redesign `naj`, then merge this PR and release 1.0?
My proposition is next:
- keep all files in `dist` folder and release 0.4X.0 version
- create an issue for `dist/near-api-js.js` deletion with `1.0` tag
- exclude `dist/near-api-js.js` when we are ready and release 1.0

volovyk-s:
@mehtaphysical @thisisjoshford @MaximusHaximus I need one more review here.

MaximusHaximus (commented):
:+1: looks good; keeping everything in `dist` for now seems like the safest bet

MaximusHaximus (approved):
:+1:

package.json
volovyk-s: We decided to keep the not-minified version for now.

PR #633, opened 2021-07-02
`Expired` error handled

volovyk-s:


src/account.ts
```diff
@@ -229,6 +229,10 @@ export class Account {
                     delete this.accessKeyByPublicKeyCache[publicKey.toString()];
                     return null;
                 }
+                if (error.type === 'Expired') {
+                    console.warn(`Expired block_hash`);
```
MaximusHaximus, line 233:
```suggestion
                    console.warn(`Retrying transaction ${receiverId}:${baseEncode(txHash)} due to expired block hash`);
```

MaximusHaximus (approved):
A brief textual description on the PR would be nice just for indexing purposes :)

I think the goal here is to retry just like we do for `InvalidNonce` type errors above -- if that's the case, I think it'd be nice from the user's perspective if the logs for both of these case were phrased in basically the same way for consistency -- made one suggestion in-line for that :). 

Otherwise, LGTM!

volovyk-s:
@MaximusHaximus yeah, sorry
All the context is in the issue.

volovyk-s:
@mikedotexe @thisisjoshford @mehtaphysical we still need the second approval here.

mehtaphysical (approved):
Seems like our normal pattern.

We should discuss if we want to `console.warn` on retry though.

PR #638, opened 2021-07-09
add verify signature example

thisisjoshford:
Adds a cookbook example that demonstrates how to verify an account signature.


examples/cookbook/package.json
```diff
@@ -12,6 +12,7 @@
     "chalk": "^4.1.1",
     "homedir": "^0.6.0",
```
MaximusHaximus, line 13: Not related _directly_  to this PR, but just something that I happened to notice since you aren't using it in this example -- can we delete this package from our dependencies entirely since it is part of the node `os` package implementation? I don't think we even use it, so it'd be good to get it out of here as a follow-up PR

examples/cookbook/utils/verify-signature.js
```diff
@@ -0,0 +1,37 @@
...
+const config = {
+    keyStore,
+    networkId: "testnet",
+    nodeUrl: "https://rpc.testnet.near.org",
+};
+
+verifySignature();
+
+async function verifySignature() {
+    const near = await connect({ ...config, keyStore });
```
MaximusHaximus, line 22: `config` already contains `keyStore`, so this spread is a bit confusing

examples/cookbook/utils/verify-signature.js
```diff
@@ -0,0 +1,37 @@
...
+    const account = await near.account(ACCOUNT_ID);
+    const tokenMessage = new TextEncoder().encode(`Hello world, ${ACCOUNT_ID}`);
+    const signatureData = await account.connection.signer.signMessage(
+        tokenMessage,
+        ACCOUNT_ID,
+        "testnet"
```
MaximusHaximus, line 28: I think we should be referencing `config.networkId`, that way when if someone changes the example to try something against `mainnet`, they don't need to find and update this magic string for it to match their connection configuration :)

examples/cookbook/utils/verify-signature.js
```diff
@@ -0,0 +1,37 @@
...
+    );
+    const hash = new Uint8Array(sha256.sha256.array(tokenMessage));
```
MaximusHaximus, line 30: Why use `new Uint8Array()` here, but use `TextEncoder.encode()` above?

Since this is an example, if there is a reason for the difference it'd be nice if there was a comment explaining why here.

However, if they are effectively doing the same thing, maybe rather than use `sha256.array()`, could we make the composition of the Uint8Array here be done consistently by using e.g. `new TextEncoder().encode(sha256.sha256(tokenMessage))` directly, since `encode()` _returns a Uint8Array for us_?

examples/cookbook/utils/verify-signature.js
```diff
@@ -0,0 +1,37 @@
...
+    let isValid = nacl.sign.detached.verify(
+        hash,
+        signatureData.signature,
+        signatureData.publicKey.data
+    );
+    console.log(isValid);
```
MaximusHaximus, line 36: Would be nice to return `isValid` also, assuming `verify()` doesn't `throw` when it is invalid (most people, when converting this example to real-world code, will need the result of the function)

MaximusHaximus (changes requested):
Added comments/questions in-line!

examples/cookbook/package.json
thisisjoshford: Will do!

mehtaphysical:
After some thought, I think this would be best as a `KeyPair` example:

```js
async function verifySignature() {
  const keyPair = await keyStore.getKey("testnet", ACCOUNT_ID);
  const msg = Buffer.from("hi");

  const { signature } = keyPair.sign(msg);

  const isValid = keyPair.verify(msg, signature);

  console.log("Signature Valid?:", isValid);

  return isValid;
}
```

mikedotexe (approved):

examples/cookbook/utils/verify-signature.js
thisisjoshford: Refactored the code based on @mehtaphysical 's suggestion :)

examples/cookbook/package.json
```diff
@@ -12,6 +12,7 @@
...
     "near-api-js": "^0.41.0",
-    "path": "^0.12.7"
+    "path": "^0.12.7",
+    "tweetnacl": "^1.0.3"
```
mehtaphysical, line 16: these can be removed.

- `tweetnacl` is no longer used
- `path` is built into node

mehtaphysical (changes requested):
just some deps changes

mehtaphysical (approved):

Issue #640, opened 2021-07-11
Missing keyStore in connect raise an error whereas it should not 


zurgl:
**Describe the bug**
The claim, from the official [doc](https://near.github.io/near-api-js/modules/connect.html)

> Connect to NEAR using the provided configuration.
> ConnectConfig.networkId and ConnectConfig.nodeUrl are required

is not verified.
 
**To Reproduce**
run this test case:
```javascript
(async () => {
    const near = await connect({
        networkId: "testnet",
        nodeUrl: "https://rpc.testnet.near.org",
    });
    return 'connected'
})()
    .then(console.log)
    .catch(console.error)
```
**Expected behavior**
This test should pass, but in fact we have to provide a keystore otherwise it fail.

**logged error**
```bash
TypeError: Cannot read property 'keyStore' of undefined
    at new Near (xxxxx/test/node_modules/near-api-js/lib/near.js:34:105)
...
```

**Fix**
Applying the following patch to `lib/near.js` at line 34
```javascript
            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || config.deps?.keyStore }
```
precisly `config.deps?.keyStore` replacing `config.deps.keyStore`

Fix the issue

volovyk-s:
Closing this issue since it affects deprecated API's. More info in the linked PR.

PR #641, opened 2021-07-11
fix issue 640

zurgl:
Should fix the issue describe [here](https://github.com/near/near-api-js/issues/640)

### DIFF
--- a/lib/near.js
+++ b/lib/near.js
@@ -31,7 +31,7 @@ class Near {
         this.connection = connection_1.Connection.fromConfig({
             networkId: config.networkId,
             provider: { type: 'JsonRpcProvider', args: { url: config.nodeUrl } },
-            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || config.deps.keyStore }
+            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || config.deps?.keyStore }
         });
         if (config.masterAccount) {
             // TODO: figure out better way of specifiying initial balance.
--- a/lib/near.js
+++ b/lib/near.js
@@ -31,7 +31,7 @@ class Near {
         this.connection = connection_1.Connection.fromConfig({
             networkId: config.networkId,
             provider: { type: 'JsonRpcProvider', args: { url: config.nodeUrl } },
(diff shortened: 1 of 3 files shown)


lib/near.js
```diff
@@ -31,7 +31,7 @@ class Near {
         this.connection = connection_1.Connection.fromConfig({
             networkId: config.networkId,
             provider: { type: 'JsonRpcProvider', args: { url: config.nodeUrl } },
-            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || config.deps.keyStore }
+            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || config.deps?.keyStore }
```
volovyk-s, line 34: @zurgl please, add changes to the `.ts` files. Then run `yarn build`. `near.js` is a build artifact.
Also, can you add a test that is failing without this fix? Thanks!

zurgl:
@volovyk-s I've added the path to `near.ts` and a test case.

btw, I'm just patching a deprecated functionality.

volovyk-s:
> @volovyk-s I've added the path to `near.ts` and a test case.
> 
> btw, I'm just patching a deprecated functionality.

Yeah, you are right. Developers can face this issue only if they are following an old API.
We will completely delte `deps` object soon, so I would say that we should just close this issue and PR and follow the new API's.
But anyway, thank you for pointing this out! We have lot's of other work to do ;-)

tiendq:
> 

I'm sorry but where is your new API? I'm following [official docs](https://docs.near.org/docs/api/naj-quick-reference) and got this error too.

`TypeError: Cannot read property 'keyStore' of undefined`

volovyk-s:
Thanks @zurgl , let's continue in https://github.com/near/near-api-js/issues/740

volovyk-s:
@zurgl reopening this one, the fix is correct, my bad.
Thanks, @tiendq !

volovyk-s (approved):

ailisp (approved):
Looks good. Thank you @zurgl ! Could you please resolve the conflict so we can merge.

volovyk-s:
Copied to https://github.com/near/near-api-js/pull/752 since I can not merge this PR because of conflicts.

Issue #630, opened 2021-06-30
Investigate solution(s) for NodeJS supply chain attacks


mikedotexe:
We recently had Daryl notice that GitHub doesn't do a great job of alerting us to the fact that a lockfile might be pulling in dependencies from other package repositories. It turned out not to be malicious, but made us realize we'd like to have these types of changes stand out more; probably on the CI level.

Investigate how we can identify these.
Let's also consider NEAR CLI in this ticket instead of duplicating it over there.

volovyk-s:
Here is what I have found in `yarn` docs: https://classic.yarnpkg.com/en/docs/yarn-lock/#toc-current-package-only
Only the top-level lock file is used. `naj` is a dependency, so its `lock` file will be ignored. NEAR CLI is not a dependency (most of the time), so we should be especially careful there.
Also, here is an interesting opinion: https://gajus.medium.com/stop-using-package-lock-json-or-yarn-lock-909035e94328

volovyk-s:
@mikedotexe supply chain attack is a multidimensional problem, but to solve this particular problem with `yarn.lock` we can try one of the following:
1. Create a bot that will check if `yarn.lock` is changed and create an alert. I don't think that it should be a failed status on the CI, comment in PR should be enough (@MaximusHaximus mentioned this approach in his comment in Discord).
2. Restrict registries except `https://registry.yarnpkg.com/`
3. Configure projects to use private registry

PR #645, opened 2021-07-20
Danger setup

volovyk-s:
Fixes: https://github.com/near/near-api-js/issues/630
Check the comment below to see an example of how it works.
I will make an additional test with fake PR after the merge.


MaximusHaximus (approved):
Nice! :+1: This is awesome.  I think I'll add it to the wallet codebase too :+1:

chadoh (approved):
💋

PR #663, opened 2021-08-08
Allow accountId to used when a key path is designated

mikedotexe:
One thing we've learned from the NEAR CLI proposal and Hack-it-Fridays is the shortfall we have when it comes to storing multiple Unencrypted File Store keys for a single account.

As a user, I want to have a function-call access key and a full-access key and be able to specify which account I'm calling from and where to find the key file.

This will largely help NEAR CLI and does not change any other behavior unless the user explicitly goes out of their way to designate an `accountId` in the config that's passed to `connect`.

This change is backwards-compatible and only adds a feature that can be built upon, particularly for folks who want to have multiple keys for the same account and use NEAR CLI.

Imagine this scenario:
A user wants to add a function-call access key to their account using NEAR CLI. At the time of this writing, the new NEAR CLI is in the early stages of progress and all our instructions point to the CLI that uses near-api-js. In this version of the CLI, when someone specifies an `accountId` flag (which is required to "call" contracts) it will look for a key in the user's home directory like `~/.near-credentials/testnet/mike.testnet.json` where each JSON file only contains ONE key, not an array of keys. To get around this, I think it's reasonable for someone to generate a key that will not collide like so:

    near generate-key mike.testnet-fcak-01

The Terminal spits out the public key and this can be added as a function-call access key (fcak) let's say.

The `generate-key` command would create the file `~/.near-credentials/testnet/mike.testnet-fcak-01.json` where the contents start like this:

>{"account_id":"delme.testnet-fcak-01","public_key":"ed25519:…

Currently, if this key is going to be useful, we must change the value of the `account_id` key in the JSON above to be `mike.testnet`. 

**Only then** can we use a command like this, to make sure that a contract is called using the key file we intend:

    near call mike.testnet my_method --accountId mike.testnet --keyPath ~/.near-credentials/testnet/mike.testnet-fcak-01.json

Note that in the above command we have explicitly stated (with the required flag `accountId`) what account we are calling from.

This pull request basically allows a user to skip the step where they have to modify that unencrypted file store JSON file, changing it to match the name of the account. We can trust that when a user is interacting with the API and says, "I'm using this account ID and please use the keys from this file" that we follow their expected behavior, and not suddenly change the account ID to what the key file has.

This could be a much larger conversation that I'm prepared to discuss if that's helpful.

### DIFF
--- a/lib/account.js
+++ b/lib/account.js
@@ -3,6 +3,7 @@ var __importDefault = (this && this.__importDefault) || function (mod) {
     return (mod && mod.__esModule) ? mod : { "default": mod };
 };
 Object.defineProperty(exports, "__esModule", { value: true });
+exports.Account = void 0;
 const bn_js_1 = __importDefault(require("bn.js"));
 const depd_1 = __importDefault(require("depd"));
 const transaction_1 = require("./transaction");
--- a/lib/account.js
+++ b/lib/account.js
@@ -3,7 +3,6 @@ var __importDefault = (this && this.__importDefault) || function (mod) {
     return (mod && mod.__esModule) ? mod : { "default": mod };
 };
 Object.defineProperty(exports, "__esModule", { value: true });
-exports.Account = void 0;
 const bn_js_1 = __importDefault(require("bn.js"));
 const depd_1 = __importDefault(require("depd"));
 const transaction_1 = require("./transaction");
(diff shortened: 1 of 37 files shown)


volovyk-s (commented):
@mikedotexe can you please run `yarn build` and push changes? Looks like there are a lot of redundant changes.
Adding @MaximusHaximus since he agreed to work on multiple keys support issue.

PR #664, opened 2021-08-09
tweaks for near-runner / near-sandbox

chadoh:
As we build [near-runner](github.com/near/runner), we're finding lots of things we want to improve. This branch will contain our scratchpad with all of these improvements at once. Later, we will split individual improvements into separate PRs.


volovyk-s:
@chadoh should we reopen this?

PR #668, opened 2021-08-18
Borsh in view

mfornet:
Add support for custom way to stringify arguments.

The motivation was to support borsh both in view & call functions.

The implementation is based on @MaximusHaximus suggestion at https://github.com/near/near-api-js/pull/620#discussion_r656438592

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -503,12 +503,15 @@ export class Account {
         { parse = parseJsonFromRawResponse } = {}
     ): Promise<any> {
         this.validateArgs(args);
+
+        const isUint8Array = args.byteLength !== undefined && args.byteLength === args.length;
+        const serializedArgs = isUint8Array ? args : Buffer.from(JSON.stringify(args));
         
         const result = await this.connection.provider.query<CodeResult>({
             request_type: 'call_function',
             account_id: contractId,
             method_name: methodName,
-            args_base64: Buffer.from(JSON.stringify(args)).toString('base64'),
+            args_base64: serializedArgs.toString('base64'),
             finality: 'optimistic'
         });
 
--- a/src/account.ts
+++ b/src/account.ts
@@ -94,6 +94,10 @@ export interface FunctionCallOptions {
      * @see {@link RequestSignTransactionsOptions}
      */
     walletCallbackUrl?: string;
+    /**
+     * Convert input arguments into bytes array.
+     */
+    stringify?: (input: any) => Buffer;
 }
 
 interface ReceiptLogWithFailure {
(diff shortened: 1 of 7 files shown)


src/transaction.ts
```diff
@@ -51,6 +51,12 @@ export function deployContract(code: Uint8Array): Action {
     return new Action({ deployContract: new DeployContract({code}) });
 }
 
+function stringifyJsonOrBytes(args: any): Buffer {
+    const isUint8Array = args.byteLength !== undefined && args.byteLength === args.length;
+    const serializedArgs = isUint8Array ? args : Buffer.from(JSON.stringify(args));
+    return serializedArgs;
+}
+
```
mfornet, line 59: I created this default function for being backward compatible.

src/account.ts
```diff
@@ -414,11 +422,12 @@ export class Account {
         });
     }
 
-    private functionCallV2({ contractId, methodName, args = {}, gas = DEFAULT_FUNCTION_CALL_GAS, attachedDeposit, walletMeta, walletCallbackUrl }: FunctionCallOptions): Promise<FinalExecutionOutcome> {
+    private functionCallV2({ contractId, methodName, args = {}, gas = DEFAULT_FUNCTION_CALL_GAS, attachedDeposit, walletMeta, walletCallbackUrl, stringify }: FunctionCallOptions): Promise<FinalExecutionOutcome> {
         this.validateArgs(args);
+        const opts = stringify === undefined ? {} : {stringify};
```
volovyk-s, line 427: let's use a more specific name instead of `opts`

volovyk-s (commented):
Great work @mfornet ! Can you please add a simple test that will cover this functionality?

src/transaction.ts
```diff
@@ -59,12 +65,10 @@ export function deployContract(code: Uint8Array): Action {
...
+ * @param stringify Convert input arguments into bytes array.
  */
-export function functionCall(methodName: string, args: Uint8Array | object, gas: BN, deposit: BN): Action {
-    const anyArgs = args as any;
-    const isUint8Array = anyArgs.byteLength !== undefined && anyArgs.byteLength === anyArgs.length;
-    const serializedArgs = isUint8Array ? args : Buffer.from(JSON.stringify(args));
-    return new Action({functionCall: new FunctionCall({methodName, args: serializedArgs, gas, deposit }) });
+export function functionCall(methodName: string, args: Uint8Array | object, gas: BN, deposit: BN, {stringify = stringifyJsonOrBytes} = {}): Action {
+    return new Action({functionCall: new FunctionCall({methodName, args: stringify(args), gas, deposit }) });
 }
```
MaximusHaximus, line 72: There's actually a destructured new object argument aded here that has a single property called `stringify` on that object, which makes the JSDoc here inaccurate.  For the JSDoc to be accurate, function signature should be:
```suggestion
export function functionCall(methodName: string, args: Uint8Array | object, gas: BN, deposit: BN, stringify = stringifyJsonOrBytes): Action {
    return new Action({functionCall: new FunctionCall({methodName, args: stringify(args), gas, deposit }) });
}
```
Or we could document `opts` in the JSDoc

MaximusHaximus (changes requested):
Just one thing about the change to `functionCall()` signature

src/account.ts
mfornet: is `options` specific enough? 

Feel free to suggest a new name.

mfornet: Removed this altogether.

src/transaction.ts
mfornet: Applied suggestion instead (use argument stringify)

volovyk-s (approved):

MaximusHaximus (approved):
:+1: LGTM :), thanks @mfornet !

PR #669, opened 2021-08-18
batch-transaction fix

volovyk-s:


volovyk-s:
@thisisjoshford looks like there is no contract file that is needed by this example. Can you please point me to the contract that you used here?

thisisjoshford:
> @thisisjoshford looks like there is no contract file that is needed by this example. Can you please point me to the contract that you used here?

This uses the staking pool factory contract located [here](https://github.com/near/near-api-js/tree/master/examples/cookbook/utils/wasm-files) as a example contract that it deploys on [line 34](https://github.com/near/near-api-js/blob/26b3dd2f55ef97107c429fdfa704de566617c9b3/examples/cookbook/transactions/batch-transactions.js#L34) as part of the batch of transactions @volovyk-s.

mikedotexe:
I modified this a bit because this is a pretty odd example to run. Before it would create an account that had no access keys and be lost forever. Removed the create account and transfer actions to make this a little more usable.

mikedotexe (approved):

thisisjoshford (approved):

PR #671, opened 2021-08-18
build: cookbook uses root NAJ

chadoh:
Rather than using version published to NPM, the `examples/cookbook` folder can use the version of NAJ from the root of the project.


thisisjoshford (approved):
Oh, awesome! This makes way more sense. Thanks @chadoh !

volovyk-s (approved):
Awesome! And this is a required change for automated cookbook testing.

PR #690, opened 2021-09-22
retry json rpc request when expired

danielwpz:
We have seen more expired errors than ever. And these errors could usually be resolved by retrying the request.


danielwpz:
already been fixed, just need to release

PR #695, opened 2021-09-28
fix: disable internal logging with env variable

willemneal:
'NEAR_NO_LOGS ' is used to disable printing internal logs  and errors.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,3 @@
-process.env.NEAR_PRINT_LOGS = true;
-
 module.exports = {
     preset: 'ts-jest',
     testEnvironment: 'node',
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,3 +1,5 @@
+process.env.NEAR_PRINT_LOGS = true;
+
 module.exports = {
     preset: 'ts-jest',
     testEnvironment: 'node',
--- a/lib/account.js
+++ b/lib/account.js
@@ -63,18 +63,22 @@ class Account {
     }
     /** @hidden */
     printLogsAndFailures(contractId, results) {
-        for (const result of results) {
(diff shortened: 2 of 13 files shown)


willemneal:
@volovyk-s It's not running CI.  Do I need a different role to start the CI?

volovyk-s:
@willemneal also noticed that, trying to find the reason

chadoh (approved):
This works in NodeJS, but does not give browser environments an easy way to opt-in (at least not on-the-fly, without rebuilding & redeploying). Maybe that's a good thing, though.

volovyk-s:
@willemneal looks like it's a problem with GitHub: 
image
I will double-check if it will not be solved by tomorrow.

willemneal:
> This works in NodeJS, but does not give browser environments an easy way to opt-in (at least not on-the-fly, without rebuilding & redeploying). Maybe that's a good thing, though.

Yeah I figure that console logs in browser aren't a huge deal.  Also they can mock `process` with a bundler.

willemneal:
@volovyk-s no rush, just wanted to get this in before a release since it will make our test output much cleaner.

chadoh:
Might want to check with strongly opinionated folks who might prefer that
we keep this change backwards-compatible. We could make it so verbose
logging is on by default, with an environment variable flag to turn it off.

willemneal:
@chadoh  It should be backwards compatible.  I didn't have to change the test setup, which uses the logs.

chadoh:
It’s not backwards compatible. Logs now need to be opted into by dependent projects. Previously they were always on.

We could make the environment variable something like `NODE_NO_LOGS`, like your description says, instead of `NEAR_PRINT_LOGS`, which is currently used in the code here.

willemneal:
@chadoh My bad I hadn't pushed my changes that did exactly as you mention.  Also adding more now to suppress warnings too.

chadoh (approved):

volovyk-s (approved):

PR #639, opened 2021-07-10
use block_id in rpc query

mehtaphysical:
Fixes #635

Issue #635, opened 2021-07-07
RpcQueryRequest expects and sends `blockId` but RPC node expects `block_id`


encody:
**Describe the bug**

This code conforms to the TypeScript typings, and the library has no problem sending it off:

```ts
await near.connection.provider.query({
  request_type: 'view_account',
  account_id: 'nectar.near',
  blockId: 40258947,
});
```

But the RPC node expects `block_id` instead of `blockId` and breaks.

[Possibly related](https://github.com/near/near-api-js/blob/c5762ea30119a2966505afaec132a39f19d623a1/lib/providers/provider.d.ts#L29)

**To Reproduce**

```ts
const near = await connect({
  networkId: 'mainnet',
  keyStore: new keyStores.BrowserLocalStorageKeyStore(),
  nodeUrl: 'https://archival-rpc.mainnet.near.org', // using archival RPC so that I can request a specific block ID
  walletUrl: 'https://wallet.mainnet.near.org',
  helperUrl: 'https://helper.mainnet.near.org',
});

await near.connection.provider.query({
  request_type: 'view_account',
  account_id: 'nectar.near',
  blockId: 40258947,
});
```

Generated request:

```json
{
    "method": "query",
    "params": {
        "request_type": "view_account",
        "account_id": "nectar.near",
        "blockId": 40258947
    },
    "id": 125,
    "jsonrpc": "2.0"
}
```

Response:

```json
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32602,
        "message": "Invalid params",
        "data": "Failed parsing args: no variant of enum BlockReference found in flattened data"
    },
    "id": 125
}
```

**Expected behavior**

Either the library expects `block_id` or it remaps `blockId` to `block_id` such that generated request is:

```json
{
    "method": "query",
    "params": {
        "request_type": "view_account",
        "account_id": "nectar.near",
        "block_id": 40258947
    },
    "id": 125,
    "jsonrpc": "2.0"
}
```

Response:
```json
{
    "jsonrpc": "2.0",
    "result": {
        "amount": "19916070150000000000000",
        "locked": "0",
        "code_hash": "11111111111111111111111111111111",
        "storage_usage": 346,
        "storage_paid_at": 0,
        "block_height": 40258947,
        "block_hash": "HxdYhKTs44Huyg7gZkMoUxVW6xHP2j1BfQeBroHtGeTx"
    },
    "id": 125
}
```

**Screenshots**

**Desktop (please complete the following information):**
[...]

PR #698, opened 2021-09-30
block_id fix

volovyk-s:
We have a small breaking change that I want to fix in this PR.
Somebody requested a change in this issue: https://github.com/near/near-api-js/issues/635
And it was fixed in this PR: https://github.com/near/near-api-js/pull/639
Now, when we are passing `block_id` instead of the suggested `blockId`, it's becoming `undefined`.
I think that our API should be consistent and we should support only `block_id`, so I'm basically reverting the PR mentioned above.


willemneal (approved):

chadoh (approved):
Agree with your logic here

PR #700, opened 2021-09-30
build artifacts updated

volovyk-s:


volovyk-s:
@willemneal agree
https://github.com/near/near-api-js/issues/701

chadoh (approved):

PR #702, opened 2021-10-01
fix(near.account): rm pointless async interface

willemneal:


volovyk-s (commented):
Looks like a breaking change to me:
```JavaScript
account("acc.near").then((data) => { console.log(data); });
```

willemneal:
@volovyk-s Good point.  Should we start a new branch to collect breaking changes for the next release?  This definitely feels the breaking change is still in the favor of users since it removes the async call.

volovyk-s:
@willemneal yes, let's do that. Probably we will have more breaking changes during the provider refactoring.

volovyk-s (approved):

Issue #703, opened 2021-10-01
Refactoring JSONRpcProvider


willemneal:
At its core JSONRpcProvider should be a stateless set of functions to make RPC requests.  However, in its current form it caches access keys and parses error data immediately. 

The access keys management should be handled by a key store, however, until that work is started this functionality will remain here.

As we have found with the runner it is advantageous to return the complete data from the RPC request in a class that provides parsing.

Rishabhraghwendra18:
Hi @willemneal . I think you may close this issue now

Rishabhraghwendra18:
@willemneal I am sorry . I overlooked about it . I am new to this repo , the issue description is little bit unclear to me . May you explain me what changes/implementation I have to do regarding this issue .

PR #704, opened 2021-10-01
feat: add `returnError` to transaction options

willemneal:
Currently this is the first part of #703.


volovyk-s (approved):
Agree with the logic described in the issue. We will need to deprecate `returnError = False` with time.

PR #710, opened 2021-10-12
feat: Introduced `verify` method on `PublicKey` type to verify signature correctness

### LINKED ISSUE #339: Missing Signature Verification Method
Do we need a method to verify signatures?  We can sign, but how does one verify?

frol:
*Resolves #339*

Before this PR, one could verify a message/signature only with KeyPair, which in its turn required to have a secret key, which is definitely completely irrelevant in the context of the signature verification process.


frol:
@volovyk-s done

PR #712, opened 2021-10-15
Esm artifacts

volovyk-s:
`package.json` is modified to generate the `ESM` module alongside with `CJS`.

`fix_compile.sh` is used to specify the type of the module (`CJS`/`ESM`). `esm/package.json` and `lib/package.json` are generated by `fix_compile.sh` script.
The idea for the script with motivation is taken from this article: https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html

`exports` keyword is added to `package.json`. It is a breaking change right now. I'm not sure if we can reexport all of the files.
`require` usage similar to this one is not working anymore: https://github.com/near/near-cli/blob/master/middleware/seed-phrase.js#L3
Dirty fix for it: https://github.com/near/near-api-js/pull/712/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R20
But it is not solving the problem overall. Need some fresh thoughts here.

The list of `dead` files that do not exist in TS code anymore/redundant and were deleted (not related to ESM).
- `lib/res/error_messages.d.ts`
- `lib/utils/network.d.ts`
- `lib/utils/network.js`


itegulov:
So, I have tried to make this PR work with https://github.com/tonborzenko/svelte-kit-near, but there seems to be a fundamental problem at root here: all dependencies of `near-api-js` need to be available as esm modules as well, which is not the case for many of them. For instance, we are depending on `bn.js` which uses `UMD`; funnily enough, it seems like `ethers.js` is having the same [issue](https://github.com/ethers-io/ethers.js/issues/1784) with it. I have made my best attempt at getting rid of all dependencies that cause this issue in this [branch](https://github.com/near/near-api-js/tree/pure-esm-artifacts), but in the end I was unable to reach a working state because the remaining dependencies are pretty essential. On the bright side of things, I tried to migrate [borsh-js](https://github.com/near/borsh-js) to ESM and it went very smoothly (see [this](https://github.com/near/borsh-js/tree/pure-esm-artifacts) branch), I was even able to import and use it from svelte-kit-near. The weird thing is that `borsh-js` also depends on `bn.js`, so I am not sure what is so different between them...

I think [rollupjs](https://rollupjs.org/guide/en/) is a decent candidate for a potential solution, but from what I have read in various blogposts/GH issues, it does not seem to play nicely with vite...

volovyk-s:
@itegulov here is a few updates (thanks for the help, @tonborzenko):
- looks like `next.js` is using `csj` modules even if it's imported as an `esm` module
- we will need to start refactoring from other dependencies like `borsh-js` (we should do it ourselves since it's our packet) and `tweetnacl` (esm version is available) to make it work.
- local storage is also an issue...
- usage of `tweetnacl` `esm` breaks cjs support (we need to investigate this)
- check PR from Anton: https://github.com/tonborzenko/near-api-js/commit/9052e2e8acee8bb135599d1c02e55aa7804c2406
- and this npm packet: https://www.npmjs.com/package/near-api-esm

tonborzenko:
By the way, for `tweetnacl` there are some errors with variable _self_. If somebody make `borsh-js` and `tweetnacl` into cjs+esm, I will try to help with `near-api-js`, It's not so urgent, but I think better to do it.

I am not sure, but maybe fixing `borsh-js` and `tweetnacl` will be enought for working `near-api-js` as well, becouse node.js working good with cjs, we need to make esm for browsers, and save cjs for node.js.

jessupjn:
Any updates here?

Issue #677, opened 2021-08-27
I think it should be `than` rather than `then`


Rishabhraghwendra18:
**Describe the bug**
I was sending some tokens using `account.sendMoney()` in `near-api-js` library but due to some reasons it outputs this error:
![Screenshot from 2021-08-27 16-42-53](https://user-images.githubusercontent.com/43534227/131119130-f0ec08e5-a327-41fb-b4c0-0b46b71417ce.png)

The error says that `The transaction contains more then one action ...` . I see `than` as more appropriate in place of `then` here . What do you say ?

**Browser:** Google Chrome

Rishabhraghwendra18:
@volovyk-s yup sure

PR #715, opened 2021-11-02
replaced then with than in the error msg

Rishabhraghwendra18:
issue: #677


volovyk-s (commented):
Hm... we do not have strong typing-in errors and sometimes people are checking them as substrings. It means that this is a breaking change and I would rather not merge it.
Sorry for misleading you.

Rishabhraghwendra18:
oh...no problem . I found the grammatical error in the error message so I thought of making a PR for this and sorry for such a late PR due to some health issues I was not able to contribute .

PR #722, opened 2021-11-09
Add cookbook item to check for existence of account

mikedotexe:
This has come up a few times and I think is a good cookbook example

### DIFF
--- a/examples/cookbook/utils/check-account-existence.js
+++ b/examples/cookbook/utils/check-account-existence.js
@@ -1,0 +1,28 @@
+// Demonstrates checking if an account exists
+const { providers } = require("near-api-js");
+
+const provider = new providers.JsonRpcProvider(
+    "https://archival-rpc.testnet.near.org"
+);
+
+async function accountExists() {
+    let rawResult;
+
+    for (const account_id of ["does-not-exist.mike.testnet", "mike.testnet"]) {
+        let succeeded = true;
+        try {
+            rawResult = await provider.query({
+                request_type: "view_account",
+                account_id: account_id,
+                finality: "final",
+            });
+        } catch (e) {
+            if (e.message.includes(`Server error: account ${account_id} does not exist while viewing`)) {
+                succeeded = false;
+            }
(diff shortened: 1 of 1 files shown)


examples/cookbook/utils/check-account-existence.js
```diff
@@ -0,0 +1,28 @@
...
+    for (const account_id of ["does-not-exist.mike.testnet", "mike.testnet"]) {
+        let succeeded = true;
+        try {
+            rawResult = await provider.query({
+                request_type: "view_account",
+                account_id: account_id,
+                finality: "final",
+            });
+        } catch (e) {
+            if (e.message.includes(`Server error: account ${account_id} does not exist while viewing`)) {
```
volovyk-s, line 20: You can check for `AccountDoesNotExist` error type now.

volovyk-s (commented):
Should it be in `examples/cookbook/accounts/ `?

examples/cookbook/utils/check-account-existence.js
volovyk-s, line 20:
```suggestion
            if (e.type === 'AccountDoesNotExist') {
```

volovyk-s (approved):

examples/cookbook/utils/check-account-existence.js
```diff
@@ -0,0 +1,28 @@
...
+            rawResult = await provider.query({
+                request_type: "view_account",
+                account_id: account_id,
+                finality: "final",
+            });
+        } catch (e) {
+            if (e.type === 'AccountDoesNotExist') {
+                succeeded = false;
+            }
+        }
```
itegulov, line 23: I am not sure if `view_account` can throw anything else, but shouldn't we rethrow an exception is its of an unexpected type? Otherwise we might report false-positive results in this function.

mikedotexe:
I've made this pull request because people who are building on NEAR are asking these questions and we need to be able to point to an example.
It has taken a month, we need to move way faster.

PR #724, opened 2021-11-12
Fix regex for error messages in getErrorTypeFromErrorMessage

mikedotexe:
I'm pretty sure there's no way `getErrorTypeFromErrorMessage` will work in practice with these carets (`^`), so I removed them. I would challenge anyone to provide a short JavaScript file where this works.

This has come about by working on this PR:
https://github.com/near/near-api-js/pull/722

### DIFF
--- a/src/utils/rpc_errors.ts
+++ b/src/utils/rpc_errors.ts
@@ -88,11 +88,11 @@ function walkSubtype(errorObj, schema, result, typeName) {
 export function getErrorTypeFromErrorMessage(errorMessage) {
     // This function should be removed when JSON RPC starts returning typed errors.
     switch (true) {
-    case /^account .*? does not exist while viewing$/.test(errorMessage):
+    case /account .*? does not exist while viewing$/.test(errorMessage):
         return 'AccountDoesNotExist';
-    case /^Account .*? doesn't exist$/.test(errorMessage):
+    case /Account .*? doesn't exist$/.test(errorMessage):
         return 'AccountDoesNotExist';
-    case /^access key .*? does not exist while viewing$/.test(errorMessage):
+    case /access key .*? does not exist while viewing$/.test(errorMessage):
         return 'AccessKeyDoesNotExist';
     case /wasm execution failed with error: FunctionCallError\(CompilationError\(CodeDoesNotExist/.test(errorMessage):
         return 'CodeDoesNotExist';


mikedotexe:
> @mikedotexe https://github.com/near/near-api-js/blob/master/test/utils/rpc-errors.test.js#L88 Please, provide a test if some edge cases are not covered.

Those tests that are linked simply test JavaScript's ability to perform regex and nothing to do with our library. :) I don't know how we would write tests for this unless we have the ability to use Sandbox testing, as we'd need to hit various endpoints

volovyk-s:
@mikedotexe the provided test is a unit test, and it does its job. I would be happy to add an integration test when workspaces-js will be ready for it.
If you think that regex in users code is better than regex in `naj` - feel free to use it.

volovyk-s:
@mikedotexe can you explain why this PR was created?
https://regex101.com/
Tested the existing regex here and it works just fine.

mikedotexe:
> @mikedotexe can you explain why this PR was created? Tested the existing regex [here](https://regex101.com/) and in unit tests and it works just fine.

Yes, the regex works fine, but will never work in practice and has never worked.

You can see by running this somewhere where near-api-js exists:

```js
// Demonstrates checking if an account exists
const nearAPI = require("near-api-js");
const {
    getErrorTypeFromErrorMessage,
} = nearAPI.utils.rpc_errors;

const provider = new nearAPI.providers.JsonRpcProvider(
    "https://archival-rpc.testnet.near.org"
);

function currentGetErrorTypeFromErrorMessage(errorMessage) {
    switch (true) {
        case /^account .*? does not exist while viewing$/.test(errorMessage):
            return 'AccountDoesNotExist';
        default:
            return 'UntypedError';
    }
}

function newGetErrorTypeFromErrorMessage(errorMessage) {
    switch (true) {
        case /account .*? does not exist while viewing$/.test(errorMessage):
            return 'AccountDoesNotExist';
        default:
            return 'UntypedError';
    }
}

async function accountExists() {
    let rawResult;
    try {
        rawResult = await provider.query({
            request_type: "view_account",
            account_id: "does-not-exist.mike.testnet",
            finality: "final",
        });
    } catch (e) {
        const currentErrorType = currentGetErrorTypeFromErrorMessage(e.message);
        const newErrorType = newGetErrorTypeFromErrorMessage(e.message);
        console.log('debug currentErrorType', currentErrorType);
        console.log('debug newErrorType', newErrorType);
    }
}

accountExists();
```

You'll see it prints out:

```
debug currentErrorType UntypedError
debug newErrorType AccountDoesNotExist
```

That's because the current implementation uses regex to say that the error message is going to start with "account…" but it doesn't and never will. It will be:

`Error: [-32000] Server error: account does-not-exist.mike.testnet does not exist while viewing`

This part `Error: [-32000] Server error: ` will always mean that the regex will fail.

volovyk-s:
@mikedotexe Ok, so, now I understand what is the confusion here. Error handling is a bit tricky in this part.
So, this function is not intended to be used by the end-users, it's not exported. It reminds me that we need to restrict access only to the exported functions.

This function is used only in Provider to get the error type form response and set it to the `type` field.
Check this line:
https://github.com/near/near-api-js/blob/master/src/providers/json-rpc-provider.ts#L369
The error message that you are getting is not original, it's modified here:
https://github.com/near/near-api-js/blob/master/src/providers/json-rpc-provider.ts#L361
 
Since error messages are a part of the JSON RPC API and we can expect that they will never be changed, I'm using `^` to be more strict. And it works, check this snippet:
```javascript
// Demonstrates checking if an account exists
const nearAPI = require("near-api-js");
const {
    getErrorTypeFromErrorMessage,
} = nearAPI.utils.rpc_errors;

const provider = new nearAPI.providers.JsonRpcProvider(
    "https://archival-rpc.testnet.near.org"
);

async function accountExists() {
    let rawResult;
    try {
        rawResult = await provider.query({
            request_type: "view_account",
            account_id: "does-not-exist.mike.testnet",
            finality: "final",
        });
        console.log(rawResult);
    } catch (e) {
        console.log(e.type);
    }
}

accountExists();
```

mikedotexe:
>So, this function is not intended to be used by the end-users, it's not exported. It reminds me that we need to restrict access only to the exported functions.

If it's not intended to be used by the end users, why is this PR still blocked:
https://github.com/near/near-api-js/pull/722

And how are people supposed to know to use the providers if the other way works.

volovyk-s:
@mikedotexe `getErrorTypeFromErrorMessage()` function is not intended to be used by developers. But the whole point of it is to set a `type` field to the `TypedError` object that is thrown.
I have added a suggestion in that PR: https://github.com/near/near-api-js/pull/722/files#r761791320
Do you still want to merge this one?

mikedotexe:
thanks, I think doing the .type thing will work for now. I don't think I'll be the first person to try to use the function, but it's okay

PR #725, opened 2021-11-12
fix: add check to suppress retry log

willemneal:
Now when trying a transaction the warning can be ignore.  It was very log and made the output of workspace tests very noisy.

### DIFF
--- a/lib/wallet-account.js
+++ b/lib/wallet-account.js
@@ -164,6 +164,8 @@ class WalletConnection {
         currentUrl.searchParams.delete('public_key');
         currentUrl.searchParams.delete('all_keys');
         currentUrl.searchParams.delete('account_id');
+        currentUrl.searchParams.delete('meta');
+        currentUrl.searchParams.delete('transactionHashes');
         window.history.replaceState({}, document.title, currentUrl.toString());
     }
     /**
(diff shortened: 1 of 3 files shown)


willemneal:
@volovyk-s Any thoughts?

volovyk-s (approved):
Thanks, @willemneal !

lib/wallet-account.js
```diff
@@ -164,6 +164,8 @@ class WalletConnection {
         currentUrl.searchParams.delete('public_key');
         currentUrl.searchParams.delete('all_keys');
         currentUrl.searchParams.delete('account_id');
+        currentUrl.searchParams.delete('meta');
+        currentUrl.searchParams.delete('transactionHashes');
```
mpeterdev, line 168: just double checking, is this change intentionally included in this PR? didn't see it mentioned in commit message or comments

willemneal: This is because the last PR didn't commit the JS artifacts.  So this adds them, but generally you can ignore the JS files since they get built before a release anyway.

mpeterdev: gotcha. thanks!

mpeterdev (approved):

willemneal:
@volovyk-s Any other blockers?

PR #729, opened 2021-11-17
API Keys support

volovyk-s:
- Added support for RPC Server API Keys
- Added two examples to the `cookbook` (a good place to start code review)
- fixed cookbook dependencies


examples/cookbook/api-keys/near-connection.js
```diff
@@ -0,0 +1,29 @@
+// demonstrates how to use API-KEY with 'connect' function. 
+const { connect, keyStores } = require("near-api-js");
+const path = require("path");
+
+const homedir = require("os").homedir();
+const CREDENTIALS_DIR = ".near-credentials";
+const credentialsPath = path.join(homedir, CREDENTIALS_DIR);
+const keyStore = new keyStores.UnencryptedFileSystemKeyStore(credentialsPath);
+
+const RPC_API_ENDPOINT = 'http://34.141.205.230:3030/';
```
mpeterdev, line 10: does this need to be swapped out with a URL or was this value provided by SRE for the time being?

examples/cookbook/api-keys/near-connection.js
```diff
@@ -0,0 +1,29 @@
...
+const API_KEY = 'ZmowOTMyNHUwMnUzNDA5MnUzMDk0dTIzeA==';
```
mpeterdev, line 11: I believe API keys are going to be in UUID v4 format, so giving an example with that might make it clearer for users. This looks like it also might be Base64 encoded, will users be required to do that?

src/utils/web.ts
```diff
@@ -17,39 +17,39 @@ export interface ConnectionInfo {
...
 }
 
-export async function fetchJson(connection: string | ConnectionInfo, json?: string): Promise<any> {
-    let url: string = null;
-    if (typeof(connection) === 'string') {
-        url = connection;
+export async function fetchJson(connectionInfoOrUrl: string | ConnectionInfo, json?: string): Promise<any> {
+    let connectionInfo: ConnectionInfo = null;
+    if (typeof (connectionInfoOrUrl) === 'string') {
+        connectionInfo.url = connectionInfoOrUrl;
```
MaximusHaximus, line 23: Sadly, although `typeof null` is `object`, it isn't actually an object and will throw an exception if you try to use like as one -- as far as I am aware, referencing a property of `null` will throw a TypeError (Cannot set property x of null).  I'm actually pretty surprised that Typescript didn't error on compiling this, unless Typescript transpile is converting this weirdly to make the property access safe...

**We should have a test case for this since it is the backwards compatibility branch that all existing consumers of this code will be running when we release it.**

src/providers/json-rpc-provider.ts
```diff
@@ -56,9 +56,14 @@ export class JsonRpcProvider extends Provider {
     /**
      * @param url RPC API endpoint URL
      */
-    constructor(url?: string) {
+    constructor(connectionInfoOrUrl?: string | ConnectionInfo) {
         super();
-        this.connection = { url };
+        if (!connectionInfoOrUrl) { return; }
```
MaximusHaximus, line 61: This represents a behavioural change that would need to be classified as a breaking change, given that consumers can reference `JsonRpcProvider.connection` directly.

Previously, `this.connection` would **always** at least be an object, even if it only had 1 property called `url` that might be `undefined` or an empty string.  

Now, `this.connection` will be `undefined` in this case (`url` is undefined or empty string), which could represent a breaking change for anyone who references `JsonRpcProvider.connection.xxxx` in their code when they have constructed the provider with an empty string for URL, or no url -- they would start suddenly getting errors thrown `cannot read xxx of undefined`.

examples/cookbook/api-keys/near-connection.js
MaximusHaximus: If we haven't settled on a final format for API keys, I think it would be pretty nice if our API keys were formatted in such a way that they don't need URL escaping when provided as an URL query string. What do you folks think?

examples/cookbook/package.json
```diff
@@ -11,6 +11,6 @@
   "dependencies": {
     "chalk": "^4.1.1",
     "homedir": "^0.6.0",
-    "near-api-js": "file:../../lib/index.js"
+    "near-api-js": "file:../../"
```
MaximusHaximus, line 14: Just curious why we had to change this here?

src/connection.ts
```diff
@@ -9,7 +9,10 @@ function getProvider(config: any): Provider {
     switch (config.type) {
         case undefined:
             return config;
-        case 'JsonRpcProvider': return new JsonRpcProvider(config.args.url);
+        case 'JsonRpcProvider': return new JsonRpcProvider({
```
MaximusHaximus, line 12: 1. Can we configure this with the entire `config.args` object rather than have to manually translate sub-keys under that object?  In the future it will be less code if we can add new properties to the `config.args` type and know that they are passed 1:1 into the Provider constructor
2.  It would be nice if consumers of this object could add any arbitrary headers they want, without us needing to be involved in adding code cases to support all of them explicitly.  e.g. something like `const { url, headers } = config.args;`.
We could still document that `headers` can contain any number of optional keys, but it would be really nice on the consumer side if the shape of `headers` here would be a 1:1 match with the _actual header keys to be sent to the RPC node_; that way anyone creating a provider instance could intuitively configure their headers without needing to identify which keys in `args` correspond to which keys in `connection.headers`

MaximusHaximus (changes requested):
Just a few questions/comments added in-line :)

examples/cookbook/api-keys/near-connection.js
volovyk-s: @mpeterdev I'm not sure in what format API Keys will be distributed. @chefsale what is the design for it? Is it finalized?
We can encode it to `base64` ourselves if needed.
@MaximusHaximus agree, I'm in for that.

examples/cookbook/package.json
volovyk-s: Local `naj` files were not used. `yarn` was giving me this error:
```txt
error An unexpected error occurred: "ENOTDIR: not a directory, scandir '/home/serhii/.cache/yarn/v6/npm-near-api-js-0.0.0-6b3d283e-f71e-46ce-84aa-ea1ce6432df4-1637087145174/node_modules/near-api-js'".
```
And this format is encouraged in [npm docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies).

src/utils/web.ts
volovyk-s: Nice catch. And yes, the compiler was silent about it. Fixed, + created test.

src/providers/json-rpc-provider.ts
volovyk-s: Fixed, test added. Thanks!

src/connection.ts
volovyk-s: Agree, now instead of the key itself we will pass the headers object. (example included)

examples/cookbook/api-keys/near-connection.js
volovyk-s: It was provided by our SRE team. Replaced with a placeholder, since we will not post a real working key here anyway.

mpeterdev:
I like the updates to have callouts for users to replace values. Especially
`const RPC_API_ENDPOINT = '<Replace this string with your RPC server URL>';`
since they may use third party RPCs in the future.

Last question from me, naj will still function as before if the user does not provide an API key, correct? Just want to make sure since there will be a grace period before we start requiring the keys

volovyk-s:
> I like the updates to have callouts for users to replace values. Especially `const RPC_API_ENDPOINT = '<Replace this string with your RPC server URL>';` since they may use third party RPCs in the future.
> 
> Last question from me, naj will still function as before if the user does not provide an API key, correct? Just want to make sure since there will be a grace period before we start requiring the keys

Yes, everything is made in a none breaking way.

mpeterdev (approved):

examples/cookbook/api-keys/near-connection.js
volovyk-s: Just to close this discussion:
- UUID v4 is the final format
- base64 encoding is not required

volovyk-s:
@MaximusHaximus , please, take a look at this PR.

src/providers/json-rpc-provider.ts
MaximusHaximus: 1. We can't use the same logic here as we do in `fetchJson()`, because  connectionInfoOrUrl is optional.  To ensure that this behaviour remains the same for anyone calling this method without an URL, we would actually need to invert this check and only use the new logic if passed an **object**...

2. The check on L62 is a weak equality check which will coerce falsy values, one of which is an empty string.  If we were given an empty string previously, then that empty string would become `this.connection.url` -- but with the early exit on L62, now `this.connection.url` will be `null` which will break any existing strict equality checks.

I think this would retain existing behaviour while adding support for `url` to be a `ConnectionInfo` object.
```javascript
    constructor(connectionInfoOrUrl?: string | ConnectionInfo) {
        super();

        if (value != null && typeof value == 'object') {
            this.connection = connectionInfoOrUrl as ConnectionInfo;
        } else {
            this.connection = { url: connectionInfoOrUrl };            
        }
    }
```
What do you think?

3. Should we mark the previous function signature as deprecated like we did with other functions that we migrated from positional arguments to taking a single configuration object?

4. JSDoc for this function needs to be updated or our generated near-api-js API docs will be incorrect.

MaximusHaximus (changes requested):
I think we've still got a behavioural change in the logic in the JsonRpcProvider constructor, and I noticed the JSDoc wasn't updated for that function signature -- comment added in-line :)

src/providers/json-rpc-provider.ts
volovyk-s: 1, 2, 4 -> fixed/added. Thanks :)
3. It's a constructor, so I just added a deprecation message.

MaximusHaximus (approved):
:+1: Ready to rock! <3

PR #736, opened 2021-11-25
deployMultisig: check that there's no existing contract

Patrick1904:
We deploy an empty contract when we disable 2FA (main.wasm), so we must check for that contract's hash OR no code_hash to be sure it's safe to deploy multisig

Reference issue: https://github.com/near/near-wallet/issues/2261

### DIFF
--- a/src/account_multisig.ts
+++ b/src/account_multisig.ts
@@ -43,7 +43,7 @@ export class AccountMultisig extends Account {
     }
 
     protected signAndSendTransaction(...args: any[]): Promise<FinalExecutionOutcome> {
-        if(typeof args[0] === 'string') {
+        if (typeof args[0] === 'string') {
             return this._signAndSendTransaction({ receiverId: args[0], actions: args[1] });
         }
 
@@ -68,7 +68,7 @@ export class AccountMultisig extends Account {
                     functionCall('add_request_and_confirm', args, MULTISIG_GAS, MULTISIG_DEPOSIT)
                 ]
             });
-        } catch(e) {
+        } catch (e) {
             if (e.toString().includes('Account has too many active requests. Confirm or delete some')) {
                 await this.deleteUnconfirmedRequests();
                 return await this.signAndSendTransaction(receiverId, actions);
@@ -101,7 +101,7 @@ export class AccountMultisig extends Account {
         return result;
     }
 
-    async deleteUnconfirmedRequests () {
+    async deleteUnconfirmedRequests() {
         // TODO: Delete in batch, don't delete unexpired
         // TODO: Delete in batch, don't delete unexpired (can reduce gas usage dramatically)
         const request_ids = await this.getRequestIds();
@@ -112,10 +112,10 @@ export class AccountMultisig extends Account {
             }
             try {
                 await super.signAndSendTransaction({
-                    receiverId: this.accountId, 
+                    receiverId: this.accountId,
                     actions: [functionCall('delete_request', { request_id: requestIdToDelete }, MULTISIG_GAS, MULTISIG_DEPOSIT)]
                 });
-            } catch(e) {
+            } catch (e) {
                 console.warn('Attempt to delete an earlier request before 15 minutes failed. Will try again.');
             }
         }
@@ -135,7 +135,7 @@ export class AccountMultisig extends Account {
         }
         return storageFallback[MULTISIG_STORAGE_KEY];
     }
-    
+
     setRequest(data) {
         if (this.storage) {
             return this.storage.setItem(MULTISIG_STORAGE_KEY, JSON.stringify(data));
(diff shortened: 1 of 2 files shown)


volovyk-s:
@MaximusHaximus please, take a look. I don't have the context here.

src/account_multisig.ts
```diff
@@ -19,6 +19,13 @@ export const MULTISIG_DEPOSIT = new BN('0');
 export const MULTISIG_CHANGE_METHODS = ['add_request', 'add_request_and_confirm', 'delete_request', 'confirm'];
 export const MULTISIG_CONFIRM_METHODS = ['confirm'];
 
+// We deploy an empty contract when we disable 2FA (main.wasm), so we must check for that
+// contract's hash OR no code_hash to be sure it's safe
+const ALLOW_2FA_ENABLE_HASHES = [
+    'E8jZ1giWcVrps8PcV75ATauu6gFRkcwjNtKp7NKmipZG',
```
vgrichina, line 25: @Patrick1904 most likely issue is that we have some other multisig hashes (see MULTISIG_CONTRACT_HASHES) in wallet

ideally this should be passable through config / environment variable / etc

src/account_multisig.ts
```diff
@@ -205,32 +212,33 @@ export class Account2FA extends AccountMultisig {
...
+            const actions = [
+                ...fak2lak.map((pk) => deleteKey(pk)),
+                ...fak2lak.map((pk) => addKey(pk, functionCallAccessKey(accountId, MULTISIG_CHANGE_METHODS, null))),
+                addKey(confirmOnlyKey, functionCallAccessKey(accountId, MULTISIG_CONFIRM_METHODS, null)),
+                deployContract(contractBytes),
+                functionCall('new', newArgs, MULTISIG_GAS, MULTISIG_DEPOSIT)
+            ];
+            console.log('deploying multisig contract for', accountId);
+            return await super.signAndSendTransactionWithAccount(accountId, actions);
+        } else {
```
vgrichina, line 239: nit: might be cleaner to do early exit when checking `if (!ALLOW_2FA_ENABLE_HASHES.includes((await this.state()).code_hash)) {`, so that no need to put everything inside extra block

src/account_multisig.ts
Patrick1904: @vgrichina AFAIU `'11111111111111111111111111111111'` is the default and whenever we disable multisig in the wallet we deploy the same `'E8jZ1giWcVrps8PcV75ATauu6gFRkcwjNtKp7NKmipZG'` hash so we check to make sure that the account has is either of those two before deploying a multisig contract here.

Patrick1904: > ideally this should be passable through config / environment variable / etc
Good point. Added `process.env.ALLOW_2FA_ENABLE_HASHES` 👍

src/account_multisig.ts
Patrick1904: Good one:)

src/account_multisig.ts
vgrichina: I think it is likely safer to pass it as a part of config / options in constructor / etc

because depending on bundler you won't get environment variable substituted in libraries and it could be confusing and hard to work around

vgrichina: > @vgrichina AFAIU `'11111111111111111111111111111111'` is the default and whenever we disable multisig in the wallet we deploy the same `'E8jZ1giWcVrps8PcV75ATauu6gFRkcwjNtKp7NKmipZG'` hash so we check to make sure that the account has is either of those two before deploying a multisig contract here.

we historically have deployed different multisig contract versions

Patrick1904: @vgrichina all the accounts that I tested had the same hash deployed after disabling the multisig. Is there any way to query for additional potential hashes?

MaximusHaximus:
So it turns out that _only_ making sure that we keep people from deploying this over an already currently deployed multisig contract is not enough to really handle this problem.  This is because deploying a new contract to an existing account so does not delete the on-chain state that the prior (multisig) smart contract has previously persisted on-chain.

Historically, this behaviour has been convenient in some ways, because it allowed someone to disable and re-enable 2FA  without invaliding any existing scheduled transaction(s), but it has two drawbacks:
1. Even if someone has disabled 2FA a long time ago, the next time that they deploy a new contract on the account it will likely crash due to the unexpected remaining multisig contract state still on the account.
2. It also means the account w/ disabled 2FA may still have extra staked NEAR to cover storage even though they don't care about the storage that the multisig contract was using.

In order to properly handle on-chain stored state being left behind in an inconsistent state that will likely cause problems if someone deploys another smart contract in the future, we will need to deploy a state cleanup contract, call a method on it to clean up the state, and then deploy our 'empty' contract over _that_.

Reference:
https://github.com/near/core-contracts/pull/171

PR #745, opened 2021-12-08
New findSeatPrice implementation

volovyk-s:
Appeared that it can be done in a none breaking way, but we need to provide extra parameters.
As an option, I can add a completely new function that will get that data for us.
Something like:
```javascript
export async function getSeatPrice(validators);
``` 
Or even:
```javascript
export async function getSeatPrice(CURRENT/NEXT);
```
But I'm not sure if it's flexible enough.

### DIFF
--- a/src/validators.ts
+++ b/src/validators.ts
@@ -1,6 +1,7 @@
 'use strict';
 
 import BN from 'bn.js';
+import depd from 'depd';
 import { CurrentEpochValidatorInfo, NextEpochValidatorInfo } from './providers/provider';
 
 /** Finds seat price given validators stakes and number of seats.
@@ -10,9 +11,14 @@ import { CurrentEpochValidatorInfo, NextEpochValidatorInfo } from './providers/p
  * @params minimumStakeRatio: minimum stake ratio 
  * @params protocolVersion: version of the protocol from genesis config
  */
-export function findSeatPrice(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], maxNumberOfSeats: number, minimumStakeRatio?: number, protocolVersion?: number): BN {
-    if (!minimumStakeRatio || !protocolVersion || protocolVersion < 49) {
+export function findSeatPrice(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], maxNumberOfSeats: number, minimumStakeRatio: number, protocolVersion?: number): BN {
+    if (protocolVersion && protocolVersion < 49) {
         return findSeatPriceForProtocolBefore49(validators, maxNumberOfSeats);
+    }
+    if (!minimumStakeRatio) {
+        const deprecate = depd('findSeatPrice(validators, maxNumberOfSeats)');
+        deprecate('`use `findSeatPrice(validators, maxNumberOfSeats, minimumStakeRatio)` instead');
+        minimumStakeRatio = 6250; // harcoded minimumStakeRation from 12/7/21
     }
     return findSeatPriceForProtocolAfter49(validators, maxNumberOfSeats, minimumStakeRatio);
 }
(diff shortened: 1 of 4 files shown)


nikurt (approved):

src/validators.ts
```diff
@@ -6,9 +6,18 @@ import { CurrentEpochValidatorInfo, NextEpochValidatorInfo } from './providers/p
...
+ * @params maxNumberOfSeats: maximum number of seats in the network.
+ * @params minimumStakeRatio: minimum stake ratio 
+ * @params protocolVersion: version of the protocol from genesis config
  */
-export function findSeatPrice(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], numSeats: number): BN {
+export function findSeatPrice(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], maxNumberOfSeats: number, minimumStakeRatio?: number, protocolVersion?: number): BN {
+    if (!minimumStakeRatio || !protocolVersion || protocolVersion < 49) {
+        return findSeatPriceForProtocolBefore49(validators, maxNumberOfSeats);
+    }
+    return findSeatPriceForProtocolAfter49(validators, maxNumberOfSeats, minimumStakeRatio);
```
frol, line 17: I feel the default should actually be "After49" version. I feel that given that one will only receive a new version near-api-js after an explicit upgrade, it is better to throw an exception that the old version of the function does not work without the extra arguments rather than letting developers down with the wrong seat price number.

P.S. We need to have a reference where developers can read more about the change of the seat allocation algorithm. @nikurt Is there any post/changelog about the new seat allocation?

src/validators.ts
```diff
@@ -36,6 +45,16 @@ export function findSeatPrice(validators: (CurrentEpochValidatorInfo | NextEpoch
...
 
+function findSeatPriceForProtocolAfter49(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], maxNumberOfSeats: number, minimumStakeRatio: number): BN {
+    const stakes = validators.map(v => new BN(v.stake, 10)).sort((a, b) => a.cmp(b));
+    const stakesSum = stakes.reduce((a, b) => a.add(b));
+    if (validators.length < maxNumberOfSeats) {
+        return stakesSum.div(new BN(minimumStakeRatio));
+    } else {
+        return stakes[0].add(new BN(1));
+    }
+}
```
frol, line 56: LGTM, but let's have a link to nearcore implementation for the future reference: https://github.com/near/nearcore/blob/5a8ae263ec07930cd34d0dcf5bcee250c67c02aa/chain/epoch_manager/src/validator_selection.rs#L308;L315

frol (approved):

src/validators.ts
volovyk-s: @frol we do not know `minimumStakeRatio` to use a new function as default.
We can throw deprecation warning like this and hardcode a value:
```javascript
export function findSeatPrice(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], maxNumberOfSeats: number, minimumStakeRatio?: number, protocolVersion?: number): BN {
    if (protocolVersion && protocolVersion < 49) {
        return findSeatPriceForProtocolBefore49(validators, maxNumberOfSeats);
    }
    if (!minimumStakeRatio) {
        //TODO: deprecation warning
        minimumStakeRatio =  6250;
    }
    return findSeatPriceForProtocolAfter49(validators, maxNumberOfSeats, minimumStakeRatio);
}
```
Or we can change the function signature and make a breaking change like this:
```javascript
export function findSeatPrice(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], maxNumberOfSeats: number, minimumStakeRatio: number, protocolVersion?: number): BN {
    if (protocolVersion && protocolVersion < 49) {
        return findSeatPriceForProtocolBefore49(validators, maxNumberOfSeats);
    }
    return findSeatPriceForProtocolAfter49(validators, maxNumberOfSeats, minimumStakeRatio);
}
```
I like the second option better.

frol: I would mix the two to make sure that if someone uses near-api-js from JS, they don't get garbage in return.

volovyk-s: Ok, merged.

Issue #747, opened 2021-12-09
treat `window` and `Buffer` more gracefully so that using Gatsby is possible


ryancwalsh:
**Is your feature request related to a problem? Please describe.**
I'm trying to deploy a frontend for NEAR using Gatsby, but I'm blocked because near-api-js assumes `window` will always be defined.

```
failed Building static HTML for pages - 1.293s

 ERROR #95312 

"window" is not available during server side rendering.

See our docs page for more info on this error: https://gatsby.dev/debug-html


  32 |      * @param prefix defaults to `near-api-js:keystore:`
  33 |      */
> 34 |     constructor(localStorage = window.localStorage, prefix = LOCAL_STORAGE_KEY_PREFIX) {
     | ^
  35 |         super();
  36 |         this.localStorage = localStorage;
  37 |         this.prefix = prefix;


  WebpackError: ReferenceError: window is not defined
  
  - browser_local_storage_key_store.js:34 
    [gatsby-starter-default]/[near-api-js]/lib/key_stores/browser_local_storage_key_store.js:34:1
  
  - near.ts:23 
    gatsby-starter-default/src/services/near.ts:23:13

```

This breaks the Gatsby build. It's probably simple to edit near-api-js to use `window` only when it exists.

A similar problem (with this [hacky workaround](https://stackoverflow.com/a/70296962/470749)) relates to Buffer:

```
Uncaught (in promise) ReferenceError: Buffer is not defined
    at bytesJsonStringify (webpack-internal:///./node_modules/near-api-js/lib/account.js:27)
    at ConnectedWalletAccount.viewFunction (webpack-internal:///./node_modules/near-api-js/lib/account.js:365)
    at eval (webpack-internal:///./node_modules/near-api-js/lib/contract.js:73)
    at Contract.getCommitments (webpack-internal:///./node_modules/near-api-js/lib/contract.js:15)
    at _callee$ (webpack-internal:///./src/services/near.ts:80)
```

near-api-js probably could similarly be edited to handle Buffer more gracefully. 

**Describe the solution you'd like**
See above.

**Describe alternatives you've considered**

**Additional context**

volovyk-s:
Thanks, @ryancwalsh ! That needs to be fixed. Do you think you can create a PR?

ryancwalsh:
Hey @volovyk-s , I'm curious whether anyone has made progress in the direction that @MaximusHaximus described at https://github.com/near/near-api-js/pull/748. Where can I track that?

[...]

PR #748, opened 2021-12-10
Gracefully handle `window` global variable

ryancwalsh:
See https://github.com/near/near-api-js/issues/747

All 19 tests in `yarn test` pass.

And my Gatsby project that uses near-api-js now builds too.

### DIFF
--- a/lib/key_stores/browser_local_storage_key_store.d.ts
+++ b/lib/key_stores/browser_local_storage_key_store.d.ts
@@ -1,7 +1,7 @@
 import { KeyStore } from './keystore';
 import { KeyPair } from '../utils/key_pair';
 /**
- * This class is used to store keys in the browsers local storage.
+ * This class is used to store keys in the browser's local storage.
  *
  * @example {@link https://docs.near.org/docs/develop/front-end/naj-quick-reference#key-store}
  * @example
--- a/lib/key_stores/browser_local_storage_key_store.js
+++ b/lib/key_stores/browser_local_storage_key_store.js
@@ -4,8 +4,9 @@ exports.BrowserLocalStorageKeyStore = void 0;
 const keystore_1 = require("./keystore");
 const key_pair_1 = require("../utils/key_pair");
 const LOCAL_STORAGE_KEY_PREFIX = 'near-api-js:keystore:';
+const windowLocalStorage = (typeof window === 'undefined') ? null : window.localStorage;
 /**
- * This class is used to store keys in the browsers local storage.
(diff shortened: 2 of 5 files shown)


volovyk-s:
@MaximusHaximus can it be a breaking change for any web app?

MaximusHaximus:
> @MaximusHaximus can it be a breaking change for any web app?

I don't think so, but although this changeset may make the code execute without throwing a hard error due to the invalid reference, it also makes the code fail to work in unexpected ways rather than throwing a hard error if someone were to try to use it in the wrong context.  The code that relies on `window` cannot function correctly without a browser context, and the code that relies on `Buffer` cannot function correctly without a real Buffer object implementation existing -- working around the error and providing 'fake' values for the browser's local storage and window objects just means the code will not throw a hard error when you try to use it -- the logic that exists in this PR doesn't make the object _actually function correctly_ in the wrong context.

The key thing to understand is that _you can't use a `WalletConnectedAccount` in the backend/nodeJS context_ -- the _entire purpose_ of this object is to interact with the Wallet _by way of being run inside a web browser_.  'Fixing' this problem actually just hides from the user that they are trying to use an object in a context where it will never work.  The confusing part here is that NAJ contains code that is designed to run in totally different contexts.

To fix this problem _properly_, we need to give the user the ability to know which code will run in which context based on the package that it is part of, and give them an easy way to import only the code that is relevant in the context that they are bundling it for.  

At a high level, this means that we will need to put all code that is **not** context-agnostic into a package named e.g. `near-api-js-core`.  If someone imports this package into _any_ JS environment, the code will work there.  However, any code that is designed to only work in the browser should be shipped as `near-api-js-browser` and code that is designed to only work in the nodeJS context should be shipped as `near-api-js-node`.  Both the `browser` and `node` packages would depend on  the `core` package.  I'm working on a detailed proposal for this style of refactoring now, actually, as this type of confusion about where code is designed to be executed is a _really_ common thing I see reported, and has lead to a series of monkey-patches and hacks.  Note that none of these changes will ever make a `WalletConnectedAccount` object work in the nodejs context, because the _only way this object can function is by changing `window.location` in a browser context_.

ryancwalsh:
@MaximusHaximus Great points! Thanks!

ryancwalsh:
Hey @MaximusHaximus , I'm curious whether you or anyone has made progress in the direction that you described above.  Where can I track that?

If not, can you please reopen https://github.com/near/near-api-js/issues/747 because I keep running into this problem.

Thanks!

PR #752, opened 2021-12-15
near.deps check added

volovyk-s:
Closing #640

### DIFF
--- a/lib/near.js
+++ b/lib/near.js
@@ -31,7 +31,7 @@ class Near {
         this.connection = connection_1.Connection.fromConfig({
             networkId: config.networkId,
             provider: { type: 'JsonRpcProvider', args: { url: config.nodeUrl, headers: config.headers } },
-            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || config.deps.keyStore }
+            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || (config.deps && config.deps.keyStore) }
         });
         if (config.masterAccount) {
             // TODO: figure out better way of specifiying initial balance.
--- a/src/near.ts
+++ b/src/near.ts
@@ -85,7 +85,7 @@ export class Near {
         this.connection = Connection.fromConfig({
             networkId: config.networkId,
             provider: { type: 'JsonRpcProvider', args: { url: config.nodeUrl, headers: config.headers } },
-            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || config.deps.keyStore }
+            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || (config.deps && config.deps.keyStore) }
         });
         if (config.masterAccount) {
             // TODO: figure out better way of specifiying initial balance.
--- a/test/providers.test.js
+++ b/test/providers.test.js
@@ -329,3 +329,10 @@ test('JsonRpc connection url is not null on empty string', async () => {
     const provider = new nearApi.providers.JsonRpcProvider('');
     expect(provider.connection).toStrictEqual({ url: '' });
 });
+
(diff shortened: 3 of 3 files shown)


lib/near.js
```diff
@@ -31,7 +31,7 @@ class Near {
         this.connection = connection_1.Connection.fromConfig({
             networkId: config.networkId,
             provider: { type: 'JsonRpcProvider', args: { url: config.nodeUrl, headers: config.headers } },
-            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || config.deps.keyStore }
+            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || (config.deps && config.deps.keyStore) }
```
ChaoticTempest, line 34: Wouldn't `keyStore` evaluate to `false` if `config.deps` is false? Or is there another check later to see if `config.keyStore` is a valid object?

volovyk-s: So... if `config.keyStore` exists it will be returned. If it does not exist - `config.deps.keyStore` will be returned. Or `undefined`, if they are both `undefined`.
Try to play with this example to understand it better:
```javascript
let var1;
let var2 = {"test": {"test2": "test3"}};
let testValue = var1 || (var2.test && var2.test.test2);
console.log(testValue);
```

ChaoticTempest: Got it, yeah I was just wondering if we handled the undefined case properly, but it looks like we do 👍

test/providers.test.js
```diff
@@ -328,4 +328,11 @@ test('JsonRpc connection object exist without url provided', async () => {
 test('JsonRpc connection url is not null on empty string', async () => {
     const provider = new nearApi.providers.JsonRpcProvider('');
     expect(provider.connection).toStrictEqual({ url: '' });
+});
+
+test('near json rpc fetch node status', async () => {
+    const config = Object.assign(require('./config')(process.env.NODE_ENV || 'test'));
```
ChaoticTempest, line 334: What does the `Object.assign` do here? From what I've seen from the docs, it should take a target and some sources, but it looks like we're only supplying the target. Wouldn't this:
```suggestion
    const config = require('./config')(process.env.NODE_ENV || 'test');
```
be the same? Or is there some magic going on I'm not seeing here

volovyk-s: Yes, it will work the same way, thanks!

ChaoticTempest (approved):
LGTM

volovyk-s:
@itegulov please, take a look at this PR.

src/near.ts
```diff
@@ -85,7 +85,7 @@ export class Near {
         this.connection = Connection.fromConfig({
             networkId: config.networkId,
             provider: { type: 'JsonRpcProvider', args: { url: config.nodeUrl, headers: config.headers } },
-            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || config.deps.keyStore }
+            signer: config.signer || { type: 'InMemorySigner', keyStore: config.keyStore || (config.deps && config.deps.keyStore) }
```
itegulov, line 88: Hmm, I have just taken a look and it seems like `NearConfig.deps` is deprecated and we should use `NearConfig.keyStore` to access key store anyway. Am I missing something here?

volovyk-s: Yes, it's deprecated, but the keyStore is marked optional in our docs. And it was optional all the time. But now it works only if you are providing `{}`. If you will provide `undefined`, we will have a crash on `config.deps.keyStore` because `config.deps` is undefined.

itegulov: Oh, I see. I was confused about where the `config.keyStore` was coming from. Since it is supplied by the user, your change makes sense to me.

itegulov (approved):

ailisp (approved):

PR #755, opened 2021-12-18
Update borsh to 0.7.0

### LINKED ISSUE #743: When use as ES module GLOBAL not defined
I fixed importing cjs in my vite.js app, but when I try working wit SSR (it working on cjs) it falls on `global.fetch` inside connect.ts. Is it possible to rewrite it? The problem is `global` is not defined.
 [https://github.com/near/near-api-js/blob/ea4d3624c87518ab4661f850f9752986dbe78fca/esm/connect.js#L29](https://github.com/near/near-api-js/blob/ea4d3624c87518ab4661f850f9752986dbe78fca/esm/connect.js#L29)

rhythnic:
This fixes the issue of `global` not defined as it occurred in `borsh`, but it doesn't remove usage of `global` in connect.ts.
I'm working on fixing the other usage of global now.  I can do a separate PR or add to this one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "types": "lib/index.d.ts",
     "dependencies": {
         "bn.js": "5.2.0",
-        "borsh": "^0.6.0",
+        "borsh": "^0.7.0",
         "bs58": "^4.0.0",
         "depd": "^2.0.0",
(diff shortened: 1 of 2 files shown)


rhythnic:
Noticed that this PR is a duplicate of #741.  Closing.

Issue #733, opened 2021-11-22
RPC Redundancy/Failover Configuration


TrevorJTClarke:
***Is your feature request related to a problem? Please describe.***
 Current RPC providers can have downtime, temporary connectivity issues, or rate limits that make clients transactions fail. Over the past year we have observed several windows of RPC failure, which could have been mitigated if near-api-js had configurations for multiple RPC providers.

***Describe the solution you'd like***
 Similar to issues #703 and #717, the core JSON RPC provider should be refactored to not only have retries, but allow for retries against multiple Providers. The default provider list can configure both near foundation and openshards RPC services. Provider list can be a single node string for backward compatibility OR an array of node strings.
 Up for discussion but needed: Create a failover threshold of retries for each provider, and a threshold for provider failures before defaulting to a different priority.

***Configuration Example:***
 ```bash
 RPC_MAINNET_PROVIDERS="https://rpc.mainnet.near.org,[https://mainnet-rpc.openshards.io](https://mainnet-rpc.openshards.io)"

RPC_GUILDNET_PROVIDERS="https://guildnet-rpc.openshards.io,[https://rpc.guildnet.near.org](https://rpc.guildnet.near.org)"
 ```

NOTE: Because near-api-js is used in many dapps and repos, this functionality is very key toward providing the easiest way to allow clients to decentralize their RPC access. This is critical for community attacks against public resources.

***Describe alternatives you've considered***
 Users must create multiple instances of the Near module with different providers configured and detect TXN failures. Not idea at all.

***Additional context***
 There is an ongoing effort to create a decentralized RPC for mainnet & guildnet using many of the openshards.io nodes with a redundant load-balancer.

volovyk-s:
I believe our strategy for the decentralization of RPC Servers is a bit different, but some of these ideas can be implemented. On `near-api-js` level we can provide support for multiple servers and fallback logic, but it will add an additional level of complexity and source of petensioal bugs.
@frol @MaximusHaximus , any thoughts?

volovyk-s:
Similar suggestion from @artob: https://github.com/near/near-api-js/issues/735

frol:
@volovyk-s What is on your mind in terms of a different strategy? I feel near-api-js is the right abstraction layer to deal with the pool of RPC servers to enable true decentralization.

volovyk-s:
[...]

PR #769, opened 2022-01-25
Multiple nodeUrls configuration

SteveYuOWO:
Multiple nodeUrls configuration #733 
- Support multiple nodeUrls & compatible with nodeUrl
- Failover node when `fetchJson` error

Based on the failover retry mechanism, the program only needs to switch the next node when fetch error. The user had configured `n` nodes. When the `kth` node fetch error, failover to `( k + 1 ) % n` nodes.

### DIFF
--- a/lib/near.d.ts
+++ b/lib/near.d.ts
@@ -49,6 +49,11 @@ export interface NearConfig {
      */
     nodeUrl: string;
     /**
+     * NEAR RPC API urls. Compatibility with existing nodeUrl, multiple URLs for JSON RPC calls.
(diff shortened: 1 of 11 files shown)


TrevorJTClarke:
wow, just saw this - seems exactly what I was hoping for - sucks things get set to stale and not reviewed. Can it be re-opened?

SteveYuOWO:
> wow, just saw this - seems exactly what I was hoping for - sucks things get set to stale and not reviewed. Can it be re-opened?

Glad to reopen it. 😃

PR #772, opened 2022-01-26
fix: improve typings

chadoh:
* The `headers` attribute for `NearConfig` seems to be optional, but is currently typed as required
* The `appKeyPrefix` for `WalletConnection` doesn't even seem to be used, but its type (`string | null`) makes it required; better to get rid of the `| null` and add the `?` instead


chadoh:
Not stale. @volovyk-s could you review?

volovyk-s:
Hey @chadoh, agree about the headers, will merge it later.
Currently, I'm refactoring WalletConnection, I think it's better to add this change after that. And it's actually used in WwalletConection.

chadoh:
Is this stale? It's a quick and useful fix that could be merged while the larger fix is still in-progress.

PR #777, opened 2022-01-28
allow key_pair encryption/decryption

RoyTimes:
Our dapps require the near-wallet to decrypt or encrypt messages with Tweetnacl. Therefore, we are adding such capacities in the base `key_pair` class in this repo. 

Please kindly review.


RoyTimes:
flush stale status

RoyTimes:
flush

PR #780, opened 2022-01-31
added secp256k1 algo for key generation

artur-torosian-4irelabs:
Added algo secp256k1 for key generation process. 
It uses https://www.npmjs.com/package/secp256k1 npm module; private key generates from random bytes using core nodejs lib 'crypto', public key generation, signing and verifying provided by secp256k1 module;
This feature will add an option to generate keys using new algorithm

### DIFF
--- a/src/utils/key_pair.ts
+++ b/src/utils/key_pair.ts
@@ -1,6 +1,8 @@
 import nacl from 'tweetnacl';
 import { base_encode, base_decode } from './serialize';
 import { Assignable } from './enums';
+import { randomBytes } from 'crypto';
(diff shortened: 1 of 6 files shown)


src/utils/key_pair.ts
```diff
@@ -44,6 +49,7 @@ export class PublicKey extends Assignable {
 
     static fromString(encodedKey: string): PublicKey {
         const parts = encodedKey.split(':');
+        //TODO: ask what to do in this case
```
artur-torosian-4irelabs, line 52: Need advice on how to deal with this

PR #783, opened 2022-02-01
Account, WalletConnection and ConnectedWalletAccount interfaces

volovyk-s:
Playing with separation of Account logic. It should help to be more flexible and pass signing to the wallet/walletSelector.  More info in comments.


volovyk-s:
> _No description provided._

PR #785, opened 2022-02-02
WalletConnection refactoring

volovyks:
Refactoring of the WalletConnection. WIP.
Should be merged after #828


willemneal (approved):
My comments are just nits, but otherwise I approve this!

willemneal (changes requested):
If an argument is optional it should be after all non-optional arguments.

src/wallet/injected/wallet-injected-interface.ts
```diff
@@ -0,0 +1,196 @@
...
+    /* This function is redundant, since it can be acomplished with signAndSendTransaction.
+    Also, there is high level API in near-api-js Account class */
+    sendMoney: (params: SendMoneyParams) => Promise<unknown>;
+
+    /* Can be usefull, but since it's imposible to do it in NEAR Wallet,
+     * I'm not sure that this should be part of the standart.
+     * Even with Sender wallet, dApp is not always open,
+     * so it should not be a part of the lifecycle.
+     */
+    onAccountChanged: (callback: AccountChangedCallback) => void;
```
think-in-universe, line 94: The [`on(eventName, callback)`](https://docs.senderwallet.io/api-reference/near-provider-api#events) interface will be supported, which mostly comes from the design of MetaMask and can be useful when user switches between their accounts, or any other event happens. This doesn't make sense in web wallet though.

src/wallet/injected/wallet-injected-interface.ts
```diff
@@ -0,0 +1,196 @@
...
+    /** Creates a new Function call key that is stored locally alongside with accountId*/
+    requestSignIn: ({ contractId, methodNames }: SignInOptions) => Promise<boolean>;
+
+    /** Returns true if function call key was created or false otherwise */
+    isSignedIn: () => boolean;
+
+    // TODO: Sender Wallet returns Promise<boolean>, is it possible to return boolean?
+    // TODO: Should we make signOut reurn boolean in walletconnection interface? 
+    /** Cleares function call key and accountId from local storage*/
+    signOut: () => boolean;
```
think-in-universe, line 21: `signOut` can be synchronous without Promise I believe. (In this mode, no popup will be shown in Sender as a reminder).

src/wallet/injected/wallet-connection-sender.ts
```diff
@@ -0,0 +1,54 @@
...
+            if (res.result !== "success") {
+                throw new Error("Failed to sign out");
+            } else {
+                console.log("Signed out");
+            }
+        });
+        return true;
+    }
+    account(): Account {
+        throw new Error('Method not implemented.');
```
think-in-universe, line 52: I also suggest to implement this in Sender, but it may also require some change in `near-api-js`.

src/wallet/injected/wallet-connection-sender.ts
```diff
@@ -0,0 +1,54 @@
...
+                console.log('requestSignTransactions res:', res);
+                return res;
+            });
+    }
+    async requestSignIn({ contractId, methodNames, successUrl, failureUrl }: SignInOptions): Promise<void> {
+        const { accessKey } = await this.senderWallet.requestSignIn({
+            contractId: contractId,
+        });
+        //TODO: why do we need accessKey here?
+        console.log("New function call key added:", accessKey);
```
think-in-universe, line 33: I think the purpose is to make the function call key accessible to DApp that wants to sign transaction using `near-api-js`, but the way to make this work could be improved.

src/wallet/injected/wallet-connection-injected.ts
```diff
@@ -0,0 +1,69 @@
...
+        return this.injecteWallet.signOut();
+    }
+
+    account(): Account {
+        if (!this._connectedAccount) {
+            this._connectedAccount = new ConnectedWalletAccountInjected(this, this._near.connection, this.getAccountId());
+        }
+        return this._connectedAccount;
+    }
+}
```
think-in-universe, line 59: Another common use case is to `signData()` (messages, data), which can be supported in both Redirect and Injected Wallet.

src/wallet/wallet-connection.ts
```diff
@@ -0,0 +1,56 @@
...
+    _wallet: Wallet;
+
+    /** @hidden */
+    _near: Near;
+
+    /**
+    * @param {Near} _near Near object
+    * @param {Walle} _wallet Wallet object
+    */
+    constructor(near: Near, wallet: Wallet) {
```
willemneal, line 18: why not just `private near: Near, private wallet: Wallet`?, they should be hidden by default.

willemneal: And why `Near` and not `connection`?

src/wallet/injected/wallet-connection-sender.ts
volovyk-s: In the current design, the wallet should not return the Account object. First of all, it is not possible for Redirect (NEAR) Wallet, so the experience will not be consistent.

think-in-universe: why? I thought we use reuse the [`account()`](https://github.com/near/near-api-js/blob/e92725352a5ad33db7ba97f4d181da44af88267f/src/wallet-account.ts#L264-L272) method in `WalletConnection`. but I may misunderstand the latest design

vikinatora:
Hey @volovyks, do you think this PR is still beneficial to NAJ? If so our team can take over and finish it, so that it can get merged. If not, we should close the PR.

volovyks:
@vikinatora I think it's too old to waste time on it. But I'm happy if it can be an inspiration for somebody. Closing for now.

PR #792, opened 2022-02-11
feat: reexport near-units

willemneal:
Now users can import `near-units` to get access to `NEAR` and `Gas` classes, which have a `parse` method to parse strings with units and has a `toHuman` method to convert the units to a human readable string.

e.g.

```ts
NEAR.parse("10.19 N");
Gas.parse("100 Tgas");
```

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -3,3 +3,4 @@ export * as keyStores from './key_stores/index';
 export * from './common-index';
 export * from './connect';
 export * from './constants';
+export { NEAR, Gas, parse as parseUnits } from 'near-units';
--- a/lib/index.js
+++ b/lib/index.js
@@ -27,3 +27,7 @@ exports.keyStores = __importStar(require("./key_stores/index"));
 __exportStar(require("./common-index"), exports);
 __exportStar(require("./connect"), exports);
 __exportStar(require("./constants"), exports);
+var near_units_1 = require("near-units");
+Object.defineProperty(exports, "NEAR", { enumerable: true, get: function () { return near_units_1.NEAR; } });
+Object.defineProperty(exports, "Gas", { enumerable: true, get: function () { return near_units_1.Gas; } });
+Object.defineProperty(exports, "parseUnits", { enumerable: true, get: function () { return near_units_1.parse; } });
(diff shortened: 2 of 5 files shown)


willemneal:
Not stale

willemneal:
Bump

volovyk-s (approved):

volovyk-s:
@MaximusHaximus please, take a look. `.lock` file is ok.
@willemneal @chadoh are you planning to pass `near-units` npm package to Pagoda (Near Inc)?

chadoh:
Not stale

willemneal:
ping

MaximusHaximus:
@willemneal sorry for the delayed response; we're making some changes to processes in relation to NAJ, so you should start seeing more regular responses regarding this project as we move into Q2.  Thanks so much for your patience this quarter and your ongoing contributions to the project 💖 

We're trying to reduce the all-inclusive scope of `near-api-js` and slim down overall scope/dependencies. In the NEAR future, `near-api-js` will be split into smaller packages with more distinct responsibilities, which will allow people to use only the portions of the package that are relevant in the context they are using it (for example, specific packages for code that depends on the browser context.)

Given the intentional shift away from NAJ's current 'kitchen sink' approach, we want to avoid adding `near-units` as an explicit dependency of NAJ, and leave it to folks to import and use `near-units` explicitly if they want/need to use it.

At a high level:
1. `near-api-js` core methods will continue to expect that it will receive values that have _already_ been cast to the correct format (BN instances).
2. `near-api-js` will export only basic formatting/parsing methods like `parseNearAmount`, `formatNearAmount`.  Methods may be added explicitly to NAJ for extremely common use cases for specific amounts (`parseTGas`, `parseGgas`, for example).

I noticed that #793 depends on this PR.  It would be great to get the transaction builder interface from #793 into NAJ, but with the APIs for the transaction builder's wrapper functions updated to match the APIs of our existing action classes in NAJ, requiring that BN instances are passed in rather than `BN | string` (see `functionCall` action definition), and that plain _BN_ instances are _returned_ from any functions that return NEAR amounts.  

In short, we want any conversions to be done explicitly, at the cost of a little verbosity, and to keep our lowest-level code in NAJ working using BN instances that always represent values using the highest level of granularity/detail. We don't want 'fuzzy' string matching/base conversions to be happening _inside_ of NAJ methods -- we want callers to be explicitly responsible for conversions.

PR #793, opened 2022-02-11
feat: add Transaction Builder class and add contract check to Account

willemneal:
Depends on #792.

```
bob.createTransaction("alice.near").transfer("1000000000000000000").signAndSend();
```


willemneal:
@MaximusHaximus addressed feedback from #792

willemneal:
Not stale

PR #801, opened 2022-02-15
fix: don't use window in `new WalletConnection`

chadoh:
Quick workaround to address https://github.com/near/near-api-js/issues/747 until the larger https://github.com/near/near-api-js/pull/748 is dealt with

### DIFF
--- a/lib/wallet-account.js
+++ b/lib/wallet-account.js
@@ -38,16 +38,17 @@ const PENDING_ACCESS_KEY_PREFIX = 'pending_key'; // browser storage key for a pe
  */
 class WalletConnection {
     constructor(near, appKeyPrefix) {
+        const isBrowser = typeof window !== 'undefined';
         this._near = near;
         const authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
-        const authData = JSON.parse(window.localStorage.getItem(authDataKey));
+        const authData = isBrowser && JSON.parse(window.localStorage.getItem(authDataKey));
         this._networkId = near.config.networkId;
         this._walletBaseUrl = near.config.walletUrl;
         appKeyPrefix = appKeyPrefix || near.config.contractName || 'default';
         this._keyStore = near.connection.signer.keyStore;
         this._authData = authData || { allKeys: [] };
         this._authDataKey = authDataKey;
-        if (!this.isSignedIn()) {
+        if (isBrowser && !this.isSignedIn()) {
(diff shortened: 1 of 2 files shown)


chadoh:
You can see that this works here: https://github.com/AhaLabs/tenk-template-gatsby/commit/bdfd70d148bb9826e734f141cd964a89617c9284

This can be used now by updating your `package.json` with:

    "near-api-js": "AhaLabs/near-api-js#1.0.1"

chadoh:
Not stale!

volovyk-s (approved):

volovyk-s:
@MaximusHaximus I think we can merge such a fix since we are not working on a better resolution now.

chadoh:
Not stale

chadoh:
this stale bot is a pain

chadoh:
🥇

chadoh:
Nooooooo

agileurbanite:
@gutsyphilip @andy-haynes can you take a look at this PR?

esaminu:
@agileurbanite @MaximusHaximus can this wait until we publish a `near-api-nodejs` package shortly that can be used in the server context? I think the user needs to be alerted of these issues as early as possible (i.e. when importing) rather than when using something. Also, making the current package 'safe' to import on the server context will involve applying this pattern in many other places including `BrowserLocalStorageKeyStore` among others. 

For the purposes of SSR, like in Gatsby and Next.js I believe these issues can be addressed using dynamic imports like suggested [here](https://www.gatsbyjs.com/docs/using-client-side-only-packages/#workaround-3-use-reactlazy-and-suspense-on-client-side-only) for example so that parts of the app that require use of the current package or the future `near-api-browser` are only invoked on the client side. Users implementing SSR will have to keep dynamically importing the browser package after we create a server side package if they would like to request login or signing of a transaction from a wallet.

chadoh:
I understand this approach is not perfect but it's small and safe and fixes some important use cases. Why not merge it.

chadoh:
Like, you can merge this now and do the bigger better thing later.

ryancwalsh:
I agree. There have been so many people reporting this problem since Dec 2021 or earlier, and near-api-js has never been updated to address it.  cc @agileurbanite

esaminu:
Created a proposed solution in #896 . I think there's a genuine complaint here about devX in SSR stacks, we just have to balance that out with the behavior in Node.js as well

esaminu:
Fixed in #896 Thanks guys! 🙂

PR #802, opened 2022-02-15
fix!: avoid deploying a contract to an account with existing state

esaminu:
This PR addresses near/near-wallet#2261 and avoids bricking accounts with arbitrary contract state left over by doing the following:

- The disable 2fa procedure will clear all contract state after itself in 2 atomic steps: 
     1.    Clear all requests by calling `delete_request` to ensure that the rpc `view_state` call needed to use the state cleanup contract's [clean](https://github.com/near/core-contracts/blob/master/state-cleanup/src/lib.rs#L17) method does not throw a [TOO_LARGE_CONTRACT_STATE](https://docs.near.org/docs/api/rpc/contracts#view-contract-state) error due to the state size limit.
     2.    Deploy of the state cleanup contract passed as an argument and call the clean method in the same transaction as the standard key deletion and conversion steps and finally deploy of the ['empty' contract](https://github.com/near/near-wallet/blob/master/packages/frontend/src/wasm/main.wasm) that is passed as an argument.

- The enable 2fa procedure will submit a transaction that will always fail designed to identify the current state of the contract it is interacting with. The transaction has 2 actions, one that deploys the multisig contract and the second calls `delete_request` with the `request_id` argument set to `u32::MAX`.
  - If the transaction throws a 'No such request' error, we know that the state in the contract is valid multisig state and we proceed with the enable 2fa operation.
  - If the transaction throws a 'Multisig contract should be initialized' error, we know that we are dealing with a contract that has no state so we proceed with the enable 2fa operation adding a `new` function call at the end of the action batch.
  - If the transaction throws a 'Cannot deserialize the contract state' error then we know the contract we are dealing with has arbitrary contract state and we abort the operation throwing a `contractHasExistingStateError`.
  - If the transaction throws any other error, then we abort, throwing that error.

This was tested with different combinations of contract states and `code_hash`es [here](https://nearinc.atlassian.net/wiki/spaces/WA/pages/24969217/Log+of+responses+by+different+contract+state+code+combinations+when+a+2fa+transaction+is+sent.)

### DIFF
--- a/src/account_multisig.ts
+++ b/src/account_multisig.ts
@@ -18,6 +18,7 @@ export const MULTISIG_GAS = new BN('100000000000000');
 export const MULTISIG_DEPOSIT = new BN('0');
 export const MULTISIG_CHANGE_METHODS = ['add_request', 'add_request_and_confirm', 'delete_request', 'confirm'];
 export const MULTISIG_CONFIRM_METHODS = ['confirm'];
+const SHOULD_BE_INITIALIZED_ERROR_REGEX = /Smart contract panicked: Multisig contract should be initialized before usage/;
 
 type sendCodeFunction = () => Promise<any>;
 type getCodeFunction = (method: any) => Promise<string>;
@@ -216,16 +217,6 @@ export class Account2FA extends AccountMultisig {
     // default helpers for CH deployments of multisig
 
     async deployMultisig(contractBytes: Uint8Array) {
-        const currentAccountState = await this.viewState('').catch(error => {
-            const cause = error.cause && error.cause.name;
-            if(cause == 'NO_CONTRACT_CODE') {
-                return [];
-            }
-            throw cause == 'TOO_LARGE_CONTRACT_STATE' ? this.contractHasExistingStateError : error;
-        });
-        if(currentAccountState.length) { // TODO add check that tries to deserialize existing state and passes if it can
-            throw this.contractHasExistingStateError;
-        }
         const { accountId } = this;
 
         const seedOrLedgerKey = (await this.getRecoveryMethods()).data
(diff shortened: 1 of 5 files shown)


frol:
> This is currently blocked due to [view_state](https://docs.near.org/docs/api/rpc/contracts#view-contract-state) requests having a limit and not being paginated.

@bowenwang1996 Is there a chance that fetching a list of state keys can be faster than fetching the key-values?

bowenwang1996:
> > This is currently blocked due to [view_state](https://docs.near.org/docs/api/rpc/contracts#view-contract-state) requests having a limit and not being paginated.
> 
> @bowenwang1996 Is there a chance that fetching a list of state keys can be faster than fetching the key-values?

I don't think so, at least with the current implementation

src/account_multisig.ts
```diff
@@ -245,7 +293,21 @@ export class Account2FA extends AccountMultisig {
...
+            const cause = error.cause && error.cause.name;
+            if (cause == 'NO_CONTRACT_CODE') {
+                return [];
+            }
+            throw cause == 'TOO_LARGE_CONTRACT_STATE' ? this.contractHasExistingStateError : error;
+        });
+        const currentAccountStateKeys = currentAccountState.map(({ key }) => key.toString('base64'))
+        const cleanupActions = currentAccountState.length ? [
+            deployContract(cleanupContractBytes),
+            functionCall('clean', { keys: currentAccountStateKeys }, MULTISIG_GAS, new BN('0'))
```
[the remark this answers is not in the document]
esaminu: It's about 101 Tgas based on my testing with 100 Tgas being attached to the `functionCall` . I'm not sure if this is significant given the 300 Tgas limit and whether it would make sense to attach less gas to the `clean` function call. [here](https://explorer.testnet.near.org/transactions/9wzLLWdQHozXL6WPe66A3gPW6R1gz89ej5cQeb1SeMPV)'s a basic disable transaction with no requests on the multisig.

src/account_multisig.ts
```diff
@@ -101,6 +113,37 @@ export class AccountMultisig extends Account {
...
+                throw e;
+            })
+    }
+
+    async deleteAllRequests() {
+        const request_ids = await this.getRequestIds();
+        if(request_ids.length) {
+            await super.signAndSendTransaction({
+                receiverId: this.accountId,
+                actions: request_ids.map(requestIdToDelete => functionCall('delete_request', { request_id: requestIdToDelete }, MULTISIG_GAS, MULTISIG_DEPOSIT))
```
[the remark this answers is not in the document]
esaminu: It did indeed exceed 300 Tgas in my testing with more than 3 requests. Changed to a `Promise.all` thanks for catching 👍

src/account_multisig.ts
```diff
@@ -226,14 +298,23 @@ export class Account2FA extends AccountMultisig {
...
+              return await super.signAndSendTransactionWithAccount(accountId, actions);
+            case MultisigStateStatus.INVALID_STATE:
+              throw new TypedError(`Can not deploy a contract to account ${this.accountId} on network ${this.connection.networkId}, the account has existing state.`, 'ContractHasExistingState');
+            default:
+              throw new TypedError(`Can not deploy a contract to account ${this.accountId} on network ${this.connection.networkId}, the account state could not be verified.`, 'ContractStateUnknown');
+        }
     }
 
-    async disable(contractBytes: Uint8Array) {
+    async disable(contractBytes: Uint8Array, cleanupContractBytes: Uint8Array) {
```
MaximusHaximus, line 317: Please add a JSDoc on this function that references the compiled source bytes for the clean-up contract since this (and so the clean-up contract itself) is now a required parameter

src/account_multisig.ts
```diff
@@ -101,6 +122,57 @@ export class AccountMultisig extends Account {
...
+        } catch (e) {
+            if (new RegExp(MultisigDeleteRequestRejectionError.CANNOT_DESERIALIZE_STATE).test(e?.kind?.ExecutionError)) {
+                return { codeStatus: validCodeStatusIfNoDeploy, stateStatus: MultisigStateStatus.INVALID_STATE };
+            } else if (new RegExp(MultisigDeleteRequestRejectionError.MULTISIG_NOT_INITIALIZED).test(e?.kind?.ExecutionError)) {
+                return { codeStatus: validCodeStatusIfNoDeploy, stateStatus: MultisigStateStatus.STATE_NOT_INITIALIZED };
+            } else if (new RegExp(MultisigDeleteRequestRejectionError.NO_SUCH_REQUEST).test(e?.kind?.ExecutionError)) {
+                return { codeStatus: validCodeStatusIfNoDeploy, stateStatus: MultisigStateStatus.VALID_STATE };
+            } else if (new RegExp(MultisigDeleteRequestRejectionError.METHOD_NOT_FOUND).test(e?.message)) {
+                // not reachable if transaction included a deploy
+                return { codeStatus: MultisigCodeStatus.UNKNOWN_CODE, stateStatus: MultisigStateStatus.UNKNOWN_STATE };
```
MaximusHaximus, line 156: Shouldn't this case be returning `MultisigCodeStatus.INVALID_CODE`?

MaximusHaximus (changes requested):
We're just about there - just a few things in-line and then we can get this landed :+1: <3

src/account_multisig.ts
esaminu: Thanks for catching!

MaximusHaximus (approved):
:+1: nice work! Excited to get this landed :)

PR #806, opened 2022-02-19
docs: remove sender property from `new nearAPI.Contract`

### LINKED ISSUE #799: possible typo in quick reference docs
***Describe the bug***
 The quick reference docs includes a section about <span class="error">[*loading a contract*]</span>([https://github.com/near/near-api-js/blob/master/examples/quick-reference.md#load-contract](https://github.com/near/near-api-js/blob/master/examples/quick-reference.md#load-contract)). In the sub-section about loading a contract from a wallet there appears to be a typo. 

The current code is as follows,
 ```js
 const contract = new nearAPI.Contract(
 wallet.account(), // the account object that is connecting
 "example-contract.testnet",

 { // name of contract you're connecting to viewMethods: ["getMessages"], // view methods do not change state but usually return a value changeMethods: ["addMessage"], // change methods modify state sender: wallet.Account(), // account object to initialize and sign transactions. } 

);
 ```

***Expected behavior***
 I believe it should be `wallet.account` with a smaller case.
 ```diff

*   sender: wallet.Account(), // account object to initialize and sign transactions.
     + sender: wallet.account(), // account object to initialize and sign transactions.
     ```

***Additional context***
 I recently started developing on Near. I maybe wrong but I am using `wallet.account()` in my code and it seems to work.

il3ven:
Fixes #799

### DIFF
--- a/examples/quick-reference.md
+++ b/examples/quick-reference.md
@@ -355,7 +355,6 @@ const contract = new nearAPI.Contract(
     // name of contract you're connecting to
     viewMethods: ["getMessages"], // view methods do not change state but usually return a value
     changeMethods: ["addMessage"], // change methods modify state
-    sender: account, // account object to initialize and sign transactions.
   }
 );
 ```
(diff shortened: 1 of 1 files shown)


willemneal:
Bump

PR #810, opened 2022-02-23
Add secp256k1 support

samingle:
This PR adds support for the secp256k1 curve. It's based off #780, which gets pretty close. @mikedotexe asked us to take a look at the js implementation after adding secp support in the swift library  [here](https://github.com/near/near-api-swift/pull/3).


samingle:
Thanks Mike, looks like my master rebase messed up the imports 😊

mikedotexe:
Bumping a bit, as this is a pretty important feature and without it, folks are not able to use one of the two key types available in NEAR.

morgsmccauley:
Closing in favor of https://github.com/near/near-api-js/pull/985

PR #823, opened 2022-03-17
feat: add method to disable 2FA with FAK

esaminu:
This PR adds a `disableWithFAK` method to the `Account2FA` class. This method will allow 2fa disable to be performed without calling the multisig contract if a FAK is loaded into the keystore. It will optionally clear the account state using the [state cleanup contract](https://github.com/near/core-contracts/tree/master/state-cleanup) if the bytes are provided.

If no FAK is loaded into the keystore, the method will throw a `NoFAKFound` error.

`disableWithFAK` will be used on a script that is intended to restore accounts with bricked multisig with a seedphrase.

### DIFF
--- a/src/account_multisig.ts
+++ b/src/account_multisig.ts
@@ -6,11 +6,10 @@ import { Account, SignAndSendTransactionOptions } from './account';
 import { Connection } from './connection';
 import { parseNearAmount } from './utils/format';
 import { PublicKey } from './utils/key_pair';
-import { Action, addKey, deleteKey, deployContract, fullAccessKey, functionCall, functionCallAccessKey, signTransaction } from './transaction';
+import { Action, addKey, deleteKey, deployContract, fullAccessKey, functionCall, functionCallAccessKey } from './transaction';
 import { FinalExecutionOutcome, TypedError } from './providers';
 import { fetchJson } from './utils/web';
-import { AccessKeyView, FunctionCallPermissionView } from './providers/provider';
-import { baseDecode } from 'borsh';
+import { FunctionCallPermissionView } from './providers/provider';
 
 export const MULTISIG_STORAGE_KEY = '__multisigRequest';
 export const MULTISIG_ALLOWANCE = new BN(parseNearAmount('1'));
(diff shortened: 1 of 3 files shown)


src/account_multisig.ts
```diff
@@ -314,7 +314,54 @@ export class Account2FA extends AccountMultisig {
         }
     }
 
-    async disable(contractBytes: Uint8Array, cleanupContractBytes: Uint8Array) {
+    async disableWithFAK({ contractBytes, cleanupContractBytes }: { contractBytes: Uint8Array, cleanupContractBytes?: Uint8Array }) {
+        let cleanupActions = [];
+        const keyConversionActions = await this.get2faDisableKeyConversionActions();
```
MaximusHaximus, line 319: 1. Let's move this call to after the `if(cleanupContractBytes){...}` block so that we don't hit RPC until we're sure that we haven't encountered a fatal error
2. Since this method will be used in cases where the multisig contract is unhappy/unusable, we _don't_ want to throw an error just because `deleteAllRequests()` failed; in this case it is acceptable for that attempt to fail, and if it does fail, we want to _ignore_ the failure and continue with removing the smart contract and re-converting their keys from LAKs to FAKs -- otherwise this method won't allow someone to disable 2FA for an account where the multisig contract is unusable for whatever reason (underlying state etc.).

src/account_multisig.ts
```diff
@@ -314,7 +314,54 @@ export class Account2FA extends AccountMultisig {
...
+
+    async get2faDisableCleanupActions(cleanupContractBytes: Uint8Array) {
+        const currentAccountState: { key: Buffer, value: Buffer }[] = await this.viewState('').catch(error => {
+            const cause = error.cause && error.cause.name;
+            if (cause == 'NO_CONTRACT_CODE') {
+                return [];
+            }
+            throw cause == 'TOO_LARGE_CONTRACT_STATE' ?
+                new TypedError(`Can not deploy a contract to account ${this.accountId} on network ${this.connection.networkId}, the account has existing state.`, 'ContractHasExistingState') :
+                error;
```
MaximusHaximus, line 354: Suggestion for a bit easier reading:

```suggestion
            throw cause == 'TOO_LARGE_CONTRACT_STATE'
                ? new TypedError(`Can not deploy a contract to account ${this.accountId} on network ${this.connection.networkId}, the account has existing state.`, 'ContractHasExistingState')
                : error;
```

MaximusHaximus (changes requested):
Just 2 things added in-line and a nitpick about formatting -- then I think we're ready to rock :)

src/account_multisig.ts
esaminu: It's currently swallowing any `deleteAllRequests` errors and throwing them if `get2faDisableCleanupActions` throws but updated so it swallows all `deleteAllRequests` errors and throws any `get2faDisableCleanupActions` errors directly 👍

MaximusHaximus (approved):
:+1: Nice work! :)

PR #830, opened 2022-03-28
fix: types for block_id in query provider

AZbang:
I spent some time to figure out why my code was not working until I realized that the interface `BlockReference`  was not correct for RPC calls...

https://docs.near.org/docs/api/rpc#using-block_id-param


AZbang:
@vgrichina

Issue #817, opened 2022-03-14
WalletConnection calling asynchronous method from synchronous constructor


miancadonc:
_**Describe the bug**_ Constructor for `WalletConnection` calls `this._completeSignInWithAccessKey()` which is marked as asynchronous. If we initialize a `WalletConnection` in the recommended way,

 near = await connect(nearConfig);
 wallet = new WalletConnection(near);

`wallet` is not ready for use. Specifically, `_completeSignInWithAccessKey()` modifies browser history state and creates a race condition if client code tries to do the same.

 _**To Reproduce**_ Steps to reproduce the behavior:

 near = await connect(nearConfig);
 wallet = new WalletConnection(near);
 window.history.pushState(null, '', '/');

 _**Expected behavior**_ Expected result was to see the url modified according to `pushState()`, but we actually see unpredictable url results

 _**Desktop:**_ 

*   OS: macOS
*   Version 12.2.1

aka-toxa:
hey 

Just created this PR https://github.com/near/near-api-js/pull/842

janewang:
Thanks for the PR @aka-toxa!

PR #842, opened 2022-04-18
Be able to await until sign in process done completely

### LINKED ISSUE #857: WalletConnection initialisation race condition
**Describe the bug** async method `WalletConnection._completeSignInWithAccessKey` called from constructor cause `keyStore.getKey` to fail to fetch the key

**To Reproduce** after wallet login and return back to `?account_id=test.testnet&public_key=ed25519%..` in the code

const keyStore = new nearApi.keyStores.BrowserLocalStorageKeyStore(); const near = await nearApi.connect(

 {keyStore, ...nearConfig} 

); const walletConnection = new nearApi.WalletConnection(near); const keyPair = await keyStore.getKey(nearConfig.networkId, walletConnection.getAccountId())

keyPair is null

**Expected behavior** keyPair should have a valid key

**Additional context** looks like the reason is: async method `WalletConnection._completeSignInWithAccessKey` called from constructor [https://github.com/near/near-api-js/blob/master/src/wallet-account.ts#L93](https://github.com/near/near-api-js/blob/master/src/wallet-account.ts#L93) might not finish before `keyStore.getKey` tries to fetch the key

aka-toxa:
This should fix https://github.com/near/near-api-js/issues/817 and closes #857 this problem 

also it will help apps understand that keys are ready and can be used after login callback

### DIFF
--- a/src/wallet-account.ts
+++ b/src/wallet-account.ts
@@ -92,6 +92,7 @@ export class WalletConnection {
         this._keyStore = (near.connection.signer as InMemorySigner).keyStore;
         this._authData = authData || { allKeys: [] };
         this._authDataKey = authDataKey;
+        console.log(this.isSignedIn(), appKeyPrefix);
         if (!this.isSignedIn()) {
             this._completeSignInPromise = this._completeSignInWithAccessKey();
         }
--- a/src/wallet-account.ts
+++ b/src/wallet-account.ts
@@ -110,22 +110,22 @@ export class WalletConnection {
     }
 
     /**
-     * Returns promise of completing singing in
+     * Returns promise of completing signing in after redirecting from wallet
      * @example
      * ```js
      * // on login callback page
      * const wallet = new WalletConnection(near, 'my-app');
      * wallet.isSignedIn(); // false
-     * await wallet.promiseSignIn();
-     * wallet.isSignedIn(); // true
+     * await wallet.isSignedInAsync(); // true
      * ```
      */
-    async promiseSignIn() {
+    async isSignedInAsync() {
         if (!this._completeSignInPromise) {
-            return {};
+            return this.isSignedIn();
         }
 
-        return this._completeSignInPromise;
+        await this._completeSignInPromise;
+        return this.isSignedIn();
     }
 
     /**
--- a/src/wallet-account.ts
+++ b/src/wallet-account.ts
@@ -92,7 +92,6 @@ export class WalletConnection {
         this._keyStore = (near.connection.signer as InMemorySigner).keyStore;
         this._authData = authData || { allKeys: [] };
         this._authDataKey = authDataKey;
-        console.log(this.isSignedIn(), appKeyPrefix);
         if (!this.isSignedIn()) {
             this._completeSignInPromise = this._completeSignInWithAccessKey();
         }
--- a/src/wallet-account.ts
+++ b/src/wallet-account.ts
@@ -79,6 +79,9 @@ export class WalletConnection {
     /** @hidden */
     _connectedAccount: ConnectedWalletAccount;
 
+    /** @hidden */
+    _completeSignInPromise: Promise<void>;
+
     constructor(near: Near, appKeyPrefix: string | null) {
         this._near = near;
         const authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
@@ -90,7 +93,7 @@ export class WalletConnection {
         this._authData = authData || { allKeys: [] };
         this._authDataKey = authDataKey;
         if (!this.isSignedIn()) {
-            this._completeSignInWithAccessKey();
+            this._completeSignInPromise = this._completeSignInWithAccessKey();
         }
     }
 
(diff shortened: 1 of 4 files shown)


aka-toxa:
hey @vgrichina not sure I can do anything to fix the build, it looks like there are some internal issues in the build scripts

aka-toxa:
@MaximusHaximus can you assist with this PR please?

aka-toxa:
Anyone?

willemneal:
@aka-toxa It seems that there is no one tasked with maintaining this package.  They have plans to hire someone, but until then you can expect no PRs to be reviewed.  Thank you for your continued efforts and I hope for your and the ecosystem's sake that they hire someone soon!

janewang:
Reopening this PR.

@esaminu This PR was closed by the stalebot and it addresses a race condition in the contructor. See https://github.com/near/near-api-js/issues/817 Looks a valid issue and could you please review. Thanks

src/wallet-account.ts
```diff
@@ -106,6 +109,25 @@ export class WalletConnection {
...
+     * wallet.isSignedIn(); // true
+     * ```
+     */
+    async promiseSignIn() {
+        if (!this._completeSignInPromise) {
+            return {};
+        }
+
+        return this._completeSignInPromise;
+    }
```
esaminu, line 129: Do we need an additional method here? Could we just await the value we assigned the promise to in the constructor? e.g. from test:

```javascript
const newWalletConn = new nearApi.WalletConnection(nearFake, 'promise_on_complete_signin');

await newWalletConn._completeSignInPromise;
expect(await keyStore.getKey('networkId', 'near2.account')).toEqual(keyPair);
...
```

src/wallet-account.ts
```diff
@@ -222,14 +244,15 @@ export class WalletConnection {
...
+            const authData = {
                 accountId,
                 allKeys
             };
-            window.localStorage.setItem(this._authDataKey, JSON.stringify(this._authData));
+            window.localStorage.setItem(this._authDataKey, JSON.stringify(authData));
             if (publicKey) {
                 await this._moveKeyFromTempToPermanent(accountId, publicKey);
             }
+            this._authData = authData;
```
esaminu, line 255: Any reason in particular we need to make these changes?

esaminu (commented):
Thanks for the PR @aka-toxa ! This is a needed addition. I left a comment regarding how we await

src/wallet-account.ts
zcstarr: I don't think you'd want to expose the internals that way. Like if you were to change the impl, you'd be locking people in to that.

src/wallet-account.ts
```diff
@@ -106,6 +109,25 @@ export class WalletConnection {
...
+     * @example
+     * ```js
+     * // on login callback page
+     * const wallet = new WalletConnection(near, 'my-app');
+     * wallet.isSignedIn(); // false
+     * await wallet.promiseSignIn();
+     * wallet.isSignedIn(); // true
+     * ```
+     */
+    async promiseSignIn() {
```
zcstarr, line 123: would isSignInComplete or something be a better name? Also just putting 2 cents in as a community member here.

zcstarr: also I'm just a community member that is also hitting this issue :D so 🥳  for this PR

src/wallet-account.ts
esaminu: Good point. To address that we can change this to a getter, there's no need for the check and object return

src/wallet-account.ts
zcstarr: Actually thinking about this again, it seems like the check for promise should be  contained with isSignedIn(), which seems like you could have an internal variable where the promise is set up if you don't block the execution context and just use a then from the constructor. 🤔

zcstarr: updated it to callbackSignInComplete(). I think it's more descriptive of when this would occur with the connection object, so different context know when they should be blocking for url params to be set in local storage

src/wallet-account.ts
esaminu: We want this to be a predicate that resolves to whether sign in was completed on this load of the instance. Therefore the return value should not be the same in both cases. We don't imply failure by returning false if we get the naming right. 

We can actually create a second method here called `isSignedInAsync` to complement the already existing `isSignedIn`:

```javascript
async isSignedInAsync() {
       if (!this._completeSignInPromise) {
           return this.isSignedIn();
       }

       await this._completeSignInPromise;
       return this.isSignedIn();
}
```

src/wallet-account.ts
esaminu: The intention here is to only change `this.authData` on the instance once sign in is actually finished and the localStorage changes have completed successfully. Agreed ✅

esaminu: The intention here is to only modify `_authData` on the instance once sign in has _actually_ been completed successfully and the required localStorage changes have been made ✅

esaminu (approved):
Thanks for the PR @aka-toxa and thanks for the review and brainstorm @zcstarr ! 🎉

PR #848, opened 2022-05-05
Adding calbackUrl and meta to Call Contract

sandoche:
Updating the Call Contract method with a callbackUrl using the example from https://github.com/near/near-api-js/blob/master/src/contract.ts

### DIFF
--- a/examples/quick-reference.md
+++ b/examples/quick-reference.md
@@ -394,6 +394,22 @@ await contract.method_name(
 );
 ```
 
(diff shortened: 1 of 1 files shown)


sandoche:
I did the same PR here https://github.com/near/docs/pull/1057

sandoche:
Just commenting, so this PR doesn't get closed by the stale bot.

esaminu (approved):

esaminu:
Thanks for the contribution!

Issue #860, opened 2022-05-23
RPC api method txStatusReceipts does not accept String for TxHash


urieltan:
**Describe the bug** txStatus() accepts both uint8[] and string while txStatusReceipts only accept uint8[].

[<span class="image-wrap" style="">![](https://user-images.githubusercontent.com/9346641/169740398-ed50c201-73d7-4004-9488-df1472a9f182.png)</span>](https://user-images.githubusercontent.com/9346641/169740398-ed50c201-73d7-4004-9488-df1472a9f182.png)

**To Reproduce** Steps to reproduce the behavior:

1.  Run [https://github.com/near/near-api-js/blob/master/examples/cookbook/transactions/get-tx-status.js](https://github.com/near/near-api-js/blob/master/examples/cookbook/transactions/get-tx-status.js) 2\. This should work well.
2.  Change `txStatus` to `txStatusReceipts`3\. ERROR `(node:44485) UnhandledPromiseRejectionWarning: Error: <span class="error">[-32700]</span> Parse error: Failed parsing args: invalid length 2, expected fewer elements in array`

**Desktop (please complete the following information):**

*   OS: M1 Mac Big Sur 11.6

PR #861, opened 2022-05-23
fix: allow rpc txStatusReceipts to accept String

urieltan:
Reference issue #860


willemneal:
It seems you updated the JS but not the TS.  You should update the same file in `./src` and then rebuild.  Having artifacts of `js` in this lib was probably initially so that you could depend on the repo as an NPM dep.  However, ideally, it wouldn't be a part of the repo and only be built before publishing.

urieltan:
Thanks for reviewing, this is solved by PR #866 by @BatiGencho

PR #863, opened 2022-05-25
Support blockId in account viewFunction and fix viewState

evgenykuzyakov:
- Bump to version `0.45.0`
- Support passing `blockQuery` to `viewFunction` in account, so you can make view calls at different height. This is pre-work for near-cli to support passing `--blockId=123` for `near view`. Added test
- Fix `viewState` method to parse `blockQuery` correctly when `blockId` is given.

### DIFF
--- a/src/providers/json-rpc-provider.ts
+++ b/src/providers/json-rpc-provider.ts
@@ -348,6 +348,10 @@ export class JsonRpcProvider extends Provider {
     async sendJsonRpc<T>(method: string, params: object): Promise<T> {
         const response = await exponentialBackoff(REQUEST_RETRY_WAIT, REQUEST_RETRY_NUMBER, REQUEST_RETRY_WAIT_BACKOFF, async () => {
             try {
+                if ('blockId' in params) {
+                    params['block_id'] = params['blockId'];
+                    delete params['blockId'];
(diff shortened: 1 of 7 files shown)


src/providers/json-rpc-provider.ts
```diff
@@ -348,6 +348,10 @@ export class JsonRpcProvider extends Provider {
     async sendJsonRpc<T>(method: string, params: object): Promise<T> {
         const response = await exponentialBackoff(REQUEST_RETRY_WAIT, REQUEST_RETRY_NUMBER, REQUEST_RETRY_WAIT_BACKOFF, async () => {
             try {
+                if ('blockId' in params) {
```
vgrichina, line 351: seems like this logic should be in `query`

also probably can be smth like:
```
const { blockId: block_id, ...otherParams } = params;
params = { ...otherParams, block_id };
```

normally you won't use object fields as dictionary

vgrichina (commented):
minor change might be needed

vgrichina (approved):

volovyk-s (approved):

volovyk-s:
@MaximusHaximus hey, let's merge it and release a new version.

PR #870, opened 2022-06-03
Wal 338 abstractions if using js sdk

### LINKED ISSUE #867: Abstractions If Using JS SDK
# Overview

Currently if you are using the javascript SDK, you're forced into an isolated environment similar to aurora. This isolated environment is called an enclave. The differences between the enclave and WASM (Rust and AS) approaches are outlined in the following table (as outlined in [this PR](https://github.com/near/near-sdk-js/pull/49):

| Areas                     | WebAssembly Version                                    | Enclave Version                                                                                                                                                     |
|---------------------------|--------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Initial Deployment Cost   | **~5x** more than Rust or AssemblyScript               | - no difference                                                                                                                                                     |
| Cross-contract calls      | - no difference                                        | - synchronous and can only interact with contracts **within the jsvm**.                                                                                             |
| Integrates standards      | - no difference                                        | - does not **currently** support any standards such as NFTs, FTs, events, etc.. See [this](https://github.com/near/near-sdk-js/discussions/35) discussion for more details.                                                                                     |
| Using near-api-js         | - no difference                                        | - you **must** encode your arguments and make calls to the `jsvm` account rather than the JS contract                                                               |
| GAS                       | - no difference                                        | - 30% GAS burnt payed out to the contract not **yet** implemented.                                                                                                  |
| External packages         | - supports external npm packages written in javascript | - supports external npm packages written in javascript                                                                                                              |
| Clearing State            | - no difference                                        | - you **must** deploy a contract to your account and call methods to clear state. You cannot delete your account and recreate it.                                   |
| Function Call Access Keys | - no difference                                        | - Any user with a function call access key can call any method from any contract. Logic must be implemented in contract to allow/disallow certain calls can proceed |

> **NOTE:** these differences are noted as of right now. Some of them will be changed in future. An example of this is the 30% gas burnt payed out to the contracts will be implemented in the enclave version.

If you'd like to check out my WIP quickstart guide for using the enclave, checkout [this draft PR](https://github.com/near/docs/pull/1073). In addition, I've made a game using the JS SDK (which I'm working on a frontend for). This game can be found [here](https://github.com/BenKurrek/near-sdk-js-game/tree/main).

## Abstractions made with the NEAR-CLI

The current development cycle when creating a javascript is as follows:
- Create a contract written in JS
- Compile the contract down to bytecode (i think) and then encode it using base64. This is in contrast with Rust and AS where the contracts are compiled to wasm.

If you want to deploy the contract, you need to call a method on the enclave contract (`jsvm.testnet`) and pass the base64 as an argument. Using the NEAR-CLI, this looks as such:

```bash
near call jsvm.testnet deploy_js_contract --accountId YOUR_ACCOUNT --base64 --args $(cat nim-game/build/nim.base64) --deposit 1
```
This is contrast with Rust and AS as to deploy a contract, you would simply run: 

```bash
near deploy --wasmFile YOUR_WASM.wasm --accountId YOUR_ACCOUNT
```
The differences here are that using the JSVM, the contract is never actually "deployed" and simply lives in the state of the JSVM contract. Your account ID will have no contract deployed to it. When using Rust or AS, the contract is deployed to your account and lives on your account. 

This deploy functionality has been abstracted to look more like the AS and Rust deployment command in [this PR](https://github.com/near/near-cli/pull/975) and there is an ongoing discussion found [here](https://github.com/near/near-cli/discussions/978). The new command is: 

```bash
near js deploy --accountId <my-account> --base64File <path> --deposit 0.1
```

This makes it easier for developers to interact with their contracts and they don't need to know about the enclave. This sort of abstractions has been made for calling, viewing, and deploying smart contracts to the JSVM enclave.

# Using NAJ Right Now

The conventional way to interact with a deployed smart contract on NEAR is to do something as follows:

```js
await contract.account.functionCall({
      contractId: accountName,
      methodName: "newGame",
      args: {
        playerOne: "benjiman.testnet",
        playerTwo: "patrick.testnet"
      },
      gas: "300000000000000",
    });
```
You're signing a transaction whereby the receiver is the actual contract ID and you're passing in the arguments as key-value pairs. If you'd like to interact with a javascript contract that's deployed to the enclave, the way you use NAJ is different. The first thing you need to do is to introduce an encoding function like such:

```js
function encodeCall(contract, method, args) {
  return Buffer.concat([Buffer.from(contract), Buffer.from([0]), Buffer.from(method), Buffer.from([0]), Buffer.from(args)])
}
```
Once you have the encode function, you need to encode your arguments and pass them in like so:
```js
let args = encodeCall(contractId, 'getState', '');
const result = await wallet.account().functionCall({
    contractId: "jsvm.testnet", 
    methodName: 'call_js_contract', 
    args,
    gas: "300000000000000"
});
```
If you want to perform a view call, you would need to do this:
```js
let args = encodeCall(contractId, 'getState', '');
const stateView = await wallet.account().viewFunction("jsvm.testnet", 'view_js_contract', args, {
    stringify: (val) => val,
});
```

# The Ask

If we can attempt to mirror what is being done in the NEAR-CLI and abstract some of this encoding and strange behaviour away from the user, it would provide a better UX. We can introduce an optional flag for whether or not users have an enclave JS contract and if they do, the encoding can happen behind the scenes and the receiver ID can be set to the JSVM account behind the scenes. 

We could have: 

```js
await contract.account.functionCall({
      contractId: "contract.benji.testnet,
      methodName: "newGame",
      args: {
        playerOne: "benjiman.testnet",
        playerTwo: "patrick.testnet"
      },
      gas: "300000000000000",
      jsContract: true
    });
```
which would do this behind the scenes:

```js
let args = encodeCall("contract.benji.testnet", 'newGame', '["benjiman.testnet", "patrick.testnet"]');
const result = await wallet.account().functionCall({
    contractId: "jsvm.testnet", 
    methodName: 'call_js_contract', 
    args,
    gas: "300000000000000"
});
```

# Additional discussions

If you want to read more about some of the discussions going on surrounding the JS SDK, please check [this](https://github.com/near/near-sdk-js/discussions) out.

gutsyphilip:
Closes #867 

In this PR:
- Adds support for passing a `jsContract` boolean to the functionCall payload which indicates that the call came from the JS SDK. Hence, args gets encoded accordingly.

### DIFF
--- a/lib/account.js
+++ b/lib/account.js
@@ -288,7 +288,7 @@ class Account {
                 return Buffer.concat([Buffer.from(contractId), Buffer.from([0]), Buffer.from(method), Buffer.from([0]), Buffer.from(args)]);
             }
             const encodedArgs = encodeCall(contractId, methodName, JSON.stringify(Object.values(args)));
-            functionCallArgs = ['call_js_contract', encodedArgs, gas, attachedDeposit, null, null, true];
+            functionCallArgs = ['call_js_contract', encodedArgs, gas, attachedDeposit, null, true];
         }
         else {
             const stringifyArg = stringify === undefined ? transaction_1.stringifyJsonOrBytes : stringify;
--- a/src/account.ts
+++ b/src/account.ts
@@ -442,7 +442,7 @@ export class Account {
                 return Buffer.concat([Buffer.from(contractId), Buffer.from([0]), Buffer.from(method), Buffer.from([0]), Buffer.from(args)]);
             }
             const encodedArgs = encodeCall( contractId, methodName, JSON.stringify(Object.values(args)) );
-            functionCallArgs =  ['call_js_contract', encodedArgs, gas, attachedDeposit, null, null, true ];
+            functionCallArgs =  ['call_js_contract', encodedArgs, gas, attachedDeposit, null, true ];
         } else{
             const stringifyArg = stringify === undefined ? stringifyJsonOrBytes : stringify;
             functionCallArgs = [methodName, args, gas, attachedDeposit, stringifyArg, false];


lib/account.js
```diff
@@ -280,12 +280,23 @@ class Account {
             actions: [transaction_1.functionCall(methodName, args, gas || constants_1.DEFAULT_FUNCTION_CALL_GAS, amount)]
         });
     }
-    functionCallV2({ contractId, methodName, args = {}, gas = constants_1.DEFAULT_FUNCTION_CALL_GAS, attachedDeposit, walletMeta, walletCallbackUrl, stringify }) {
+    functionCallV2({ contractId, methodName, args = {}, gas = constants_1.DEFAULT_FUNCTION_CALL_GAS, attachedDeposit, walletMeta, walletCallbackUrl, stringify, jsContract }) {
         this.validateArgs(args);
-        const stringifyArg = stringify === undefined ? transaction_1.stringifyJsonOrBytes : stringify;
+        let functionCallArgs;
+        if (jsContract) {
+            function encodeCall(contract, method, args) {
```
BenKurrek, line 287: If this is exported in `src.transaction.ts`, do we need to redefine it here?

lib/account.js
```diff
@@ -280,12 +280,23 @@ class Account {
...
+                return Buffer.concat([Buffer.from(contract), Buffer.from([0]), Buffer.from(method), Buffer.from([0]), Buffer.from(args)]);
+            }
+            let encodedArgs = encodeCall(contractId, methodName, JSON.stringify(Object.values(args)));
```
BenKurrek, line 290: The arguments need to be passed in like:

`'["arg1_is_string", "arg2_is_string", arg3_is_value, ..., "argn_is_string"]'`

lib/account.js
```diff
@@ -280,12 +280,23 @@ class Account {
...
+            functionCallArgs = ['call_js_contract', encodedArgs, gas, null, null, true];
+        }
+        else {
+            const stringifyArg = stringify === undefined ? transaction_1.stringifyJsonOrBytes : stringify;
+            functionCallArgs = [methodName, args, gas, attachedDeposit, stringifyArg, false];
+        }
         return this.signAndSendTransaction({
-            receiverId: contractId,
-            actions: [transaction_1.functionCall(methodName, args, gas, attachedDeposit, stringifyArg)],
+            receiverId: jsContract ? 'call_js_contract' : contractId,
```
BenKurrek, line 298: shouldn't this be `jsContract ? jsvm.testnet : contractId`?

src/account.ts
```diff
@@ -428,12 +433,24 @@ export class Account {
...
     }
 
-    private functionCallV2({ contractId, methodName, args = {}, gas = DEFAULT_FUNCTION_CALL_GAS, attachedDeposit, walletMeta, walletCallbackUrl, stringify }: FunctionCallOptions): Promise<FinalExecutionOutcome> {
+    private functionCallV2({ contractId, methodName, args = {}, gas = DEFAULT_FUNCTION_CALL_GAS, attachedDeposit, walletMeta, walletCallbackUrl, stringify, jsContract }: FunctionCallOptions): Promise<FinalExecutionOutcome> {
         this.validateArgs(args);
-        const stringifyArg = stringify === undefined ? stringifyJsonOrBytes : stringify;
+        let functionCallArgs;
+
+        if(jsContract){
+            function encodeCall(contract:string, method:string, args) {
```
BenKurrek, line 441: See my comments for `lib/account.js`

BenKurrek (changes requested):

lib/account.js
gutsyphilip: It's not redefined here. `src.transaction.ts` exports the default method for handling(stringifying) `args`. This only applies it if theres no custom handler passed.

lib/account.js
gutsyphilip: I'm a little unclear on what you mean @BenKurrek

BenKurrek: This actually was changed back so that arguments can be passed in using key value pairs as you have it now. Disregard my previous comment. The way it was before is that you needed to pass in arguments in an array (just the value and not key value pairs) so if I had a function that had a parameter accountId I would pass in '["benji.testnet"]' instead of '{"accountId": "benji.testnet"}' and if it's not a string, you just wouldn't use the quotes.

lib/account.js
BenKurrek: Ah ok thanks - Im not amazing at typescript

agileurbanite (approved):
LGTM

BenKurrek (approved):

BenKurrek (approved):

PR #788, opened 2022-02-02
add: returnRawResult option

june07:
Currently only the SuccessValue is returned from calling contracts with this api...

https://github.com/near/near-api-js/blob/a05667361ddab208423f5a5c0e3dd0ce182b9880/lib/providers/provider.js#L30

This option provides a means to access the full rawResult object.

PR #873, opened 2022-06-09
Return full raw transaction details

adhbh:
Initially, for contract method interaction success message is `''` empty string. This is of no use to us. We require to fetch full transaction details ie `FinalExecutionOutcome `. From this, the client can fetch everything about transaction including receipt id and transaction hash.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
-    "name": "near-api-js",
+    "name": "near-api-js-sky",
     "description": "JavaScript library to interact with NEAR Protocol via RPC API",
-    "version": "0.44.2",
+    "version": "0.44.3",
     "repository": {
         "type": "git",
         "url": "git+https://github.com/near/near-api-js.git"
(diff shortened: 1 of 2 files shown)


gagdiez:
Most of the users are actually only interested in getting the final result. For example, if you have a change method that returns an object, the users are interested in the object, and not the whole transaction information. This means that, if the change method returns nothing, the result will be '', which is correct.

adheeshdh:
@gagdiez in our case we require to get tx hash. is there anyway to do that currently using the sdk? what do you suggest? or does it makes sense to add an option for this?

package.json
```diff
@@ -1,7 +1,7 @@
 {
-    "name": "near-api-js",
+    "name": "near-api-js-sky",
```
willemneal, line 2: ?

willemneal:
This is a duplicate of #788, which addresses @gagdiez 's concerns.  However, like that PR shows this library is not being actively maintained unless they need to push out a new feature like the JSVM.

adhbh:
@gagdiez @willemneal Yes the solution is #788 looks good to me. @gagdiez will you be merging that same 788 PR? or do you want me to make the similar change and raise a new one?

PR #878, opened 2022-06-10
Validate account ID

JesseKuntz:
While working on a NEAR integration, couldn't find any re-usable account ID validator. This will handle validating both implicit and named accounts, and the named account regex was re-used from the NEAR wallet: https://github.com/near/near-wallet/blob/565b9e05471098c27134caa1d26f671cf9bc60e8/packages/frontend/src/utils/wallet.js#L69

Please let me know if there's a better place for this or if I've missed anything, thanks!

### DIFF
--- a/lib/utils/account.d.ts
+++ b/lib/utils/account.d.ts
@@ -1,0 +1,2 @@
+export declare const isValidId: (address: string) => boolean;
+export declare const isImplicitAccount: (address: string) => boolean;
--- a/lib/utils/account.js
+++ b/lib/utils/account.js
@@ -1,0 +1,17 @@
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
+exports.isImplicitAccount = exports.isValidId = void 0;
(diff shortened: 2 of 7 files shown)


esaminu (commented):
Thanks for the PR @JesseKuntz ! This is a good addition but we don't need the concept of implicit accounts here. Especially since NEAR TLAs also don't include the `.` character and this code would infer that it's an implicit account. Also, we need more test cases here and some JSDoc annotations / comments preferably.

I suggest we take advantage of [the existing implementation in near-sdk-rs](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/environment/env.rs#L815). We can even re-use the [test cases](https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/environment/env.rs#L848-L930) they've implemented for this.

JesseKuntz:
Hey @esaminu thanks for the review! Sorry for the delay - I've been OOO but will work on making those changes early next week.

MaximusHaximus:
@JesseKuntz We're going to close this for now, please do re-open once you have addressed feedback.  Thanks for your contribution!

PR #881, opened 2022-06-16
Delete stale.yml

judith-Near:


agileurbanite (approved):
Yeah let's get rid of this

PR #883, opened 2022-06-20
feat: warning that deleting an account does not transfer assets besides near

judith-Near:


BenKurrek (approved):

src/account.ts
```diff
@@ -384,6 +384,7 @@ export class Account {
      * @param beneficiaryId The NEAR account that will receive the remaining Ⓝ balance from the account being deleted
      */
     async deleteAccount(beneficiaryId: string) {
+        console.log("Deleting an account does not automatically transfer NFTs and FTs to the beneficiary address. Ensure to transfer assets before deleting.")
```
willemneal, line 387: Isn't this warning too late?

BenKurrek: The hope is that they see it and just recreate the account if they messed up. Not sure how to prevent this other than putting it in docs (which is the next step) as well.

judith-Near: this is just a last resort, applications that have a delete functionality should incorporate this copy as well into the UI.

willemneal: Ah well I think this isn't the place for that because users of NAJ want to control the logging output.  So I would at least wrap this with:
```
if (!process.env["NEAR_NO_LOGS"]) {
}
```

This way it can be disabled by the consumer of NAJ.

Patrick1904 (approved):

PR #890, opened 2022-07-12
ci: Run tests on pull requests

morgsmccauley:
This PR adds a new Github Action to run tests on submitted pull requests.


.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,30 @@
+name: Pull Request
+
+on:
+  pull_request:
```
esaminu, line 4: Would this also trigger the workflow for a push after opening the PR?

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,30 @@
...
+
+  workflow_dispatch:
+
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
esaminu, line 10: Would this only cancel previous runs of the same HEAD commit or all previous runs on the PR? I think we want the latter behavior. I think `github.ref` may work for that and there's also `github.event.pull_request.number`

esaminu (commented):
LGTM but added some questions to confirm workflow trigger and cancel behavior 😃

.github/workflows/pull-request.yml
esaminu: Thanks for clarifying! 😃

andy-haynes: Of course! I knew it worked from experience but I didn't think much about it until I saw your question so I went digging 😅

.github/workflows/pull-request.yml
morgsmccauley: The `group` here is what determines which workflows should be cancelled once a new one is triggered. Using `github.ref` (the branch name) will cancel all previous runs for the given PR, and combining with `github.workflow` (the name of this workflow) will ensure it only cancels _this_ workflow and not others triggered via the same PR.

morgsmccauley:
Also I have intentionally left out eslint and TypeScript checks because there are lots of errors 😅 We can add those in once we get them to a stable state.

.github/workflows/pull-request.yml
esaminu: Nice 👍 was just not sure that different commits on the same branch have the same `github.ref`

esaminu:
> Also I have intentionally left out eslint and TypeScript checks because there are lots of errors 😅 We can add those in once we get them to a stable state.

Cool, opinions of fuzz tests in this workflow?

morgsmccauley:
> Cool, opinions on fuzz tests in this workflow?

@esaminu Do we have fuzz tests already in place for `near-api-js`?

esaminu (approved):
LGTM 💯

PR #891, opened 2022-07-14
feat: create transaction builder class

willemneal:
Reoppening of #793


esaminu:
Thanks for this @willemneal! We should add tests of the new Builder class. Also, would this fit better as it's own module i.e. `src/transactionBuilder.ts`?

willemneal:
The circular dependency with Account lead me to keep them together.  Perhaps when refactoring NAJ we can use an interface for signer and `{ accountId: string}`. 

As a note, I've been using this personally and can vouch for it's correctness, but will add some tests.

esaminu:
Thanks! we'll keep this refactor in mind for the upcoming package separation work 👍 

Yeah I think tests would be great here to ensure future change sets to either class don't break this 🙂

willemneal:
@esaminu Added example and test

esaminu:
I rebased and fixed lint errors in [feat/tx_class](https://github.com/near/near-api-js/tree/feat/tx_class). We should push that to this PR and re review

esaminu:
Closing in favor of #998

PR #895, opened 2022-07-16
refactor(near-api-js): getAccountId() returns a string.

serranoarevalo:
According to this (https://docs.near.org/docs/tutorials/front-end/naj-examples#the-wallet-interface) getAccountId() should return a string.

### DIFF
--- a/packages/near-api-js/src/wallet-account.ts
+++ b/packages/near-api-js/src/wallet-account.ts
@@ -154,7 +154,7 @@ export class WalletConnection {
      * wallet.getAccountId();
      * ```
      */
-    getAccountId() {
+    getAccountId() : string {
         return this._authData.accountId || '';
     }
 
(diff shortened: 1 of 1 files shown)


willemneal:
You need to edit the corresponding ts source as this file is a build artifact.

src/wallet-account.ts
```diff
@@ -136,7 +136,7 @@ export class WalletConnection {
      * wallet.getAccountId();
      * ```
      */
-    getAccountId() {
+    getAccountId() : string {
         return this._authData.accountId || '';
```
willemneal, line 140: I think the real issue here is that `_authData` is an `any` type.  If it included that `_authData.accountId` was a string you wouldn't need to add this annotation.

esaminu (approved):

esaminu:
Thanks @willemneal @serranoarevalo !

PR #896, opened 2022-07-18
feat: explicit walletconnection errors on server

esaminu:
This PR attempts to address server side devX for the `WalletConnection` class per #801 by returning a proxy that intercepts gets on the instance and throws an explicit error if it's a function call or returns predictable values preventing unpredictable errors and behaviors on SSR stacks and Node.js.

### DIFF
--- a/packages/near-api-js/src/wallet-account.ts
+++ b/packages/near-api-js/src/wallet-account.ts
@@ -44,6 +44,7 @@ interface RequestSignTransactionsOptions {
 /**
  * This class is used in conjunction with the {@link BrowserLocalStorageKeyStore}.
  * It redirects users to {@link https://docs.near.org/docs/tools/near-wallet | NEAR Wallet} for key management.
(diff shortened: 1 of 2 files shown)


roshaans:
image

I was able to test this in a vanilla-js project and it seems to work! I was not aware that you could use a Proxy object like that. Really cool!

src/wallet-account.ts
```diff
@@ -80,6 +80,24 @@ export class WalletConnection {
     _connectedAccount: ConnectedWalletAccount;
 
     constructor(near: Near, appKeyPrefix: string | null) {
+        if(typeof window === 'undefined') {
+            return new Proxy(this, {
```
morgsmccauley, line 84: Should we update the JSDoc to include this new behavior?

esaminu: Great point 💯 will do 👍

morgsmccauley (approved):

Issue #753, opened 2021-12-16
Investigate possibility of automated release process


volovyk-s:
_**Is your feature request related to a problem? Please describe.**_ Releases of NAJ and other tooling take a lot of time and effort. It would be great to automate this process.

PR #897, opened 2022-07-19
ci: Publish lerna packages on push to `master`

morgsmccauley:
This PR adds a Github Action for publishing Lerna packages to NPM. With the proposed Lerna configuration the publishing pipeline will determine version bumps and generate `CHANGELOG.md` files per package using the [conventionalcommits](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-conventionalcommits) specification.

Note that this pipeline only supports publishing packages to a **single** account. With the intention to split `near-api-js` in to multiple scoped packages, which will need to published to a different account, we'll need to update this pipeline to support publishing to multiple accounts.

Closes: #753

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -1,0 +1,34 @@
+name: Publish Packages
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Setup Node
+      uses: actions/setup-node@v3
+      with:
+        node-version: 16.x
+        cache: yarn
+        registry-url: https://registry.npmjs.org
+
+    - name: Install dependencies
+      run: lerna bootstrap
+
+    - name: Publish packages
+      run: |
+        git config user.email "wallet@near.org"
+        git config user.name "CI"
+        lerna publish
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -6,7 +6,11 @@ on:
       - master
 
 jobs:
+  checks:
+    uses: ./.github/workflows/checks.yml
+
   publish:
+    needs: checks
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -35,6 +35,9 @@ jobs:
     - name: Install dependencies
       run: yarn --frozen-lockfile
 
+    - name: Build packages
+      run: lerna build
+
     - name: Publish packages
       run: lerna publish
       env:
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -12,6 +12,7 @@ jobs:
   publish:
     needs: checks
     runs-on: ubuntu-latest
+    environment: NPM
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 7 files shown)


.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Install dependencies
+      run: yarn --frozen-lockfile
+
+    - name: Build packages
+      run: lerna build
+
+    - name: Publish packages
+      run: lerna publish
+      env:
+        NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
morgsmccauley, line 44: TODO: Add `near-api-js` CI token to `NPM` Github environment.

commitlint.config.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
     extends: [
-        '@commitlint/config-conventional',
-        '@commitlint/config-lerna-scopes',
```
morgsmccauley, line 4: This only lets you use packages names (`near-api-js`) as commit scopes. @esaminu I found this a bit restrictive, i.e. I can't use `chore(release): ` for the Lerna release.

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,44 @@
...
+        cache: yarn
+        registry-url: https://registry.npmjs.org
+
+    - name: Configure Git user
+      run: |
+        git config user.email "<>"
+        git config user.name "Github Actions Bot"
+
+    - name: Install dependencies
+      run: yarn --frozen-lockfile
```
morgsmccauley, line 36: Since we have set `useWorkspaces: true` Lerna will delegate bootstrapping to yarn workspaces, so this is essentially no different to `lerna bootstrap`. I'm using `yarn` directly here to avoid installing `lerna` first.

.github/workflows/pull-request.yml
```diff
@@ -11,23 +11,4 @@ concurrency:
...
-
-    - name: Install dependencies
-      run: yarn --frozen-lockfile
-
-    - name: Run lint
-      run: yarn lint
-
-    - name: Run tests
-      run: yarn test
+    uses: ./.github/workflows/checks.yml
```
morgsmccauley, line 14: TODO: required PR checks will need to be updated since this has changed.

package.json
```diff
@@ -18,7 +18,6 @@
     "clean": "yarn lerna run clean",
     "lint": "yarn lerna run lint --stream",
     "test": "yarn lerna run test --stream --concurrency 1",
-    "browserify": "yarn lerna run browserify",
```
morgsmccauley, line 21: `browserify` is specific to `near-api-js` and _may_ not be used in future packages, I feel it is better to encapsulate the entire build under `yarn build` and have all packages implement their custom build steps under that.

commitlint.config.js
esaminu: hmm any way we can have defined scopes? I think scopes will be important as the mono repo scales for version resolution based on PR titles

package.json
esaminu: Just double checking - this isn't used by any existing scripts right? I may have included it here for that purpose

morgsmccauley: This is used in `.travis.yml` but looks like Travis isn't connected to the project anymore, we can probably remove this file.

commitlint.config.js
morgsmccauley: Scopes have no effect on versioning of packages - Lerna looks at the commit history and which files have been touched.

package.json
esaminu: Do we need to check this with previous maintainers? If not I say let's remove it 👍  cc @volovyks

morgsmccauley:
Superseded by https://github.com/near/near-api-js/pull/973

morgsmccauley:
Superseded by https://github.com/near/near-api-js/pull/973

Issue #844, opened 2022-04-22
Wrong blockId field in BlockReference type for Provider.query method


shelegdmitriy:
***Describe the bug***
 Wrong `blockId` field in <span class="error">[BlockReference]</span>([https://github.com/near/near-api-js/blob/master/src/providers/provider.ts#L35](https://github.com/near/near-api-js/blob/master/src/providers/provider.ts#L35)) type for <span class="error">[Provider.query]</span>([https://github.com/near/near-api-js/blob/master/src/providers/provider.ts#L416](https://github.com/near/near-api-js/blob/master/src/providers/provider.ts#L416)) method.

***To Reproduce***
 1\. Write some `query` method, and pass `blockId` and execute it:
 ```
 await near.connection.provider.query<ViewStateResult>({
 request_type: "view_state",

 { blockId: 59976515} 

,
 account_id: "some-account",
 prefix_base64: Buffer.from("STATE", "utf-8").toString("base64"),
 });
 ```
 2\. See this error:
 ```
 TypedError: <span class="error">[-32700]</span> Parse error: Failed parsing args: no variant of enum BlockReference found in flattened data
 at /Users/shelegdmitriy/node_modules/near-api-js/lib/providers/json-rpc-provider.js:329:31
 at processTicksAndRejections (node:internal/process/task_queues:96:5)
 at async Object.exponentialBackoff <span class="error">[as default]</span> (/Users/shelegdmitriy/node_modules/near-api-js/lib/utils/exponential-backoff.js:7:24)
 at async JsonRpcProvider.sendJsonRpc (/Users/shelegdmitriy/node_modules/near-api-js/lib/providers/json-rpc-provider.js:304:26)
 at async JsonRpcProvider.query (/Users/shelegdmitriy/node_modules/near-api-js/lib/providers/json-rpc-provider.js:116:22)
 ```
 3\. Then use the same query but change `blockId` => `block_id`.
 4\. The method will complete successfully and without errors.

***Expected behavior***
 <span class="error">[BlockReference]</span>([https://github.com/near/near-api-js/blob/master/src/providers/provider.ts#L35](https://github.com/near/near-api-js/blob/master/src/providers/provider.ts#L35)) type provide `blockId` field in snake_case.

***Screenshots***
 Failed:
 <img width="1183" alt="Снимок экрана 2022-04-22 в 17 36 58" src="https://user-images.githubusercontent.com/34593263/164736655-62632a33-d52e-4e20-a746-086c5793b929.png">

[...]

PR #898, opened 2022-07-20
Fixed wrong blockId field in BlockReference type

erditkurteshiSQA:
Closes issue #844

### DIFF
--- a/lib/providers/provider.d.ts
+++ b/lib/providers/provider.d.ts
@@ -26,7 +26,7 @@ declare type BlockHeight = number;
 export declare type BlockId = BlockHash | BlockHeight;
 export declare type Finality = 'optimistic' | 'near-final' | 'final';
 export declare type BlockReference = {
-    blockId: BlockId;
+    block_id: BlockId;
 } | {
     finality: Finality;
 } | {
--- a/src/providers/provider.ts
+++ b/src/providers/provider.ts
@@ -32,7 +32,7 @@ export type BlockId = BlockHash | BlockHeight;
 
 export type Finality = 'optimistic' | 'near-final' | 'final'
 
-export type BlockReference = { blockId: BlockId } | { finality: Finality } | { sync_checkpoint: 'genesis' | 'earliest_available' }
+export type BlockReference = { block_id: BlockId } | { finality: Finality } | { sync_checkpoint: 'genesis' | 'earliest_available' }
 
 export enum ExecutionStatusBasic {
     Unknown = 'Unknown',


esaminu:
@erditkurteshiSQA @janewang is there a reason to make this change? This is a non backwards compatible change and we have [tests](https://github.com/near/near-api-js/blob/master/test/providers.test.js#L89) that indicate this works just fine with the snake case parameter. Also it seems there are other places in the code that rely on this parameter being `block_id`

erditkurteshiSQA:
@esaminu Thank you for reviewing the PR, 
I tested it again with the latest version and I am getting the same error with this code.
``` ts
   const viewState = await provider.query<ViewStateResult>({
      request_type: "view_state",
      blockId: 95874000,
      account_id: "kurteshierdit.testnet",
      prefix_base64: Buffer.from("STATE", "utf-8").toString("base64"),
    });
```
Can you confirm that this code is working for you?

esaminu:
AFAICT these methods depend on the param being `blockId` :

- JsonRpcProvider.block
- JsonRpcProvider.blockChanges
- JsonRpcProvider.accessKeyChanges
- JsonRpcProvider.singleAccessKeyChanges
- JsonRpcProvider.accountChanges
- JsonRpcProvider.contractStateChanges
- JsonRpcProvider.contractCodeChanges

These methods depend on the param being `block_id`:

- JsonRpcProvider.experimental_protocolConfig
- JsonRpcProvider.query

Decision was to change the implementation for the last two and mark the next release as a breaking change 👍 

cc @janewang

erditkurteshiSQA:
The error in the `JsonRpcProvider.query` method was fixed by these 2 commits (PRs):
The logic for accepting `block_id` or `blockId` in the query method was fixed here
https://github.com/near/near-api-js/commit/91d9b764c968f05e87b0b230f8b389035885f97f

The reason this error **"Parse error: Failed parsing args: no variant of enum BlockReference found in flattened data"** was
showing is because the errors were not handled correctly at the time this issue was reported but they were fixed here:
https://github.com/near/near-api-js/pull/910/commits/3ad619d44f360a33275b997b80397d9b74465b6e

There's no need to change the type for `JsonRpcProvider.experimental_protocolConfig` to `blockId` because it already accepts `blockId` as a param, I think the issue here is because the `sendJsonRpc` method will always wait for a `block_id` param.

The fix for this without including a breaking change can be made by adding a workaround which is present in all other methods.
Something like:
```ts
const { blockId, ...otherParams } = blockReference as any;
```
And then when we call the `sendJsonRpc` we do something like this:
```ts
return await this.sendJsonRpc('EXPERIMENTAL_protocol_config', { ...otherParams, block_id:  blockId });
```

This PR will be closed to another PR because there are too many conflicts in the current branch: https://github.com/near/near-api-js/pull/969

esaminu:
@erditkurteshiSQA Yes, the TS `BlockReference` type is enforcing that the `blockId` field should be camel case so the user would get typescript errors if they try to pass in the snake case `block_id`. The only place that currently accepts both in the implementation is the `query` method and it has an `any` type. We need to update `experimental_protocolConfig` implementation to take in `blockId` instead of `block_id` as you outlined. This is a breaking change though because people that were calling `experimental_protocolConfig` with `block_id` will now not have that forwarded to `sendJsonRpc`. We could consider this a fix since people passing in `block_id` were doing so despite the type signature being incorrect.

esaminu:
@erditkurteshiSQA Yes, the TS `BlockReference` type is enforcing that the `blockId` field should be camel case so the user would get typescript errors if they try to pass in the snake case `block_id`. The only place that currently accepts both in the implementation is the `query` method and it has an `any` type. We need to update `experimental_protocolConfig` implementation to take in `blockId` instead of `block_id` as you outlined. This is a breaking change though because people that were calling `experimental_protocolConfig` with `block_id` will now not have that forwarded to `sendJsonRpc`. We have a test that [passes in `block_id`](https://github.com/near/near-api-js/blob/master/packages/near-api-js/test/providers.test.js#L302) that will fail after this change. 

We could consider this a fix since people passing in `block_id` were doing so despite the type signature being incorrect.

esaminu:
Closed in favor of #969

PR #906, opened 2022-07-28
fix: Fix `eslint` script/errors and run in PR builds

morgsmccauley:
## Motivation
`eslint` is not enforced anywhere within the code base so having it run within the PR builds helps maintain consistency.

## Description
This PR fixes the `eslint` yarn scripts and adds it to the `pull_request` Github action

## Checklist
- [ ] Added automated tests


package.json
```diff
@@ -59,8 +60,12 @@
         "doc": "typedoc src && touch docs/.nojekyll",
         "build": "yarn compile",
         "test": "jest test",
-        "lint": "eslint test && eslint src/**/*.ts",
-        "fix": "eslint test --fix && eslint src/**/*.ts --fix",
```
morgsmccauley, line 63: Prior commands used in `&&` will short-circuit subsequent commands, i.e. if `eslint test` failed, it wouldn't even run `eslint src`.

Using `concurrently` here guarantees that both will be run.

src/.eslintrc.yml
```diff
@@ -8,6 +8,10 @@ extends:
 parser: '@typescript-eslint/parser'
 rules:
   no-inner-declarations: off
+  indent:
+    - error
+    - 4
+    - SwitchCase: 1
```
morgsmccauley, line 14: Without this, `eslint` will enfore:
```js
switch (condition) {
case 'hello':
  return 'world';
default:
  return 'unknown';
}
```
But I feel this is a bit nicer:
```js
switch (condition) {
    case 'hello':
        return 'world';
    default:
        return 'unknown';
}
```

Any oppositions?

esaminu:
Opinions on enforcing these pre-commit / pre-push like on wallet? I assume they're currently not enforced

src/.eslintrc.yml
esaminu: I prefer the latter as well

morgsmccauley:
> Opinions on enforcing these pre-commit / pre-push like on wallet? I assume they're currently not enforced

I find pre-commit a bit invasive at times, so was thinking to add to pre-push but since pushing is usually associated with a PR, and the workflow will now run lint checks I'm wondering if it's even necessary 🤔

esaminu:
My preference is pre commit but seems we did it pre push in the wallet. I think we should do at least one since we want to avoid having to tell external contributors that the CI is failing on their PRs and minimize communication time

morgsmccauley:
@esaminu good point, forgot that actions won't even be run for external contributors initially, so would be good for them to run themselves.

I've added to pre-push 👍🏽

esaminu (approved):
🚀

PR #908, opened 2022-07-28
Naj 6 move current near api js to be a lerna package lernify the repo

esaminu:
## Motivation
This PR adds yarn workspaces and lerna to the project and begins the transition to a multi package monorepo.

## Description

- Added 1 package `@near-api-js/near-api-js`
- Upgraded eslint, typescript and eslint parsers to latest
- Changed ts target to es2015
- Added commitlint with support for lerna scopes and husky hooks to enforce conventional commits 
- Added lerna scripts for build, test, lint, browserify, clean, bundlewatch and release

## Checklist
- [x] Added automated tests


esaminu:
@morgsmccauley for CI integration review and @maxhr for documentation work review 😃

lerna.json
```diff
@@ -0,0 +1,6 @@
+{
+  "packages": ["packages/*"],
```
morgsmccauley, line 2: This isn't needed with `useWorkspaces: true`, `package.json` `workspaces` are used instead.

package.json
```diff
@@ -1,82 +1,35 @@
...
+  "name": "@near-api-js/monorepo",
+  "private": true,
+  "workspaces": {
+    "packages": [
+      "packages/*"
     ],
-    "author": "NEAR Inc"
+    "nohoist": [
+      "**/near-hello",
+      "**/near-hello/**"
```
morgsmccauley, line 10: What's this for?

packages/near-api-js/.eslintrc.yml
```diff
@@ -0,0 +1 @@
+extends: '../../.eslintrc.yml'
```
morgsmccauley, line 1: If all we are doing is extending the base config, I don't think this is needed, the base config will just be used by default?

packages/near-api-js/package.json
```diff
@@ -0,0 +1,80 @@
+{
+    "name": "@near-api-js/near-api-js",
```
morgsmccauley, line 2:
```suggestion
    "name": "near-api-js",
```
This will change where are publishing to so let's keep this as `near-api-js`. We can scope all other (new) packages.

packages/near-api-js/yarn.lock
```diff
@@ -0,0 +1,6199 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
morgsmccauley, line 1: I don't think this should exist under individual packages. Yarn workspaces/Lerna will maintain a single `yarn.lock` at the root of the directory. You'll need to delete this and probably run `lerna bootstrap` or `yarn` from the root to regenerate.

packages/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+    "files": [],
+    "references": [{ "path": "near-api-js" }]
```
morgsmccauley, line 3: What's this for? Is it so typescript can resolve imports across different packages (typesciprt projects)? If so, I don't think this is the right way to go about it. Yarn workspaces/Lerna will (sym)link packages together within `node_modules` which allows us to reference other packages in the monorepo.

yarn.lock
```diff
@@ -2,170 +2,158 @@
 # yarn lockfile v1
```
morgsmccauley, line 2: Ah I see we still have the root lockfile, we can just delete the `near-api-js` one then.

.eslintignore
```diff
@@ -1,3 +1,3 @@
-dist/
-nearcore/
-protos.js
+packages/near-api-js/dist/
```
morgsmccauley, line 1: I think we can create `.eslintignore` in `packages/near-api-js` to scope it to that directory.

morgsmccauley: Do any of these actually exist though? I couldn't find them. Is this still needed at all?

.ignore
```diff
@@ -1,3 +1,3 @@
-lib/
-dist/
-docs/
+packages/near-api-js/lib/
```
morgsmccauley, line 1: What does `.ignore` actually ignore?

.husky/pre-commit
```diff
@@ -2,4 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 yarn build
-git add lib
+git add packages/near-api-js/lib
```
morgsmccauley, line 5: This is fine for now but may become a problem as we add more packages, i.e. if we update a package other than `near-api-js` we don't really want to be running this commit hook. Especially since `yarn build` will build every package, it could become time consuming. 

Is there a way to scope this just to `near-api-js`?

package.json
```diff
@@ -1,82 +1,35 @@
...
+    ]
+  },
+  "scripts": {
+    "build": "yarn lerna run build",
+    "clean": "yarn lerna run clean --concurrency 4",
+    "lint": "yarn lerna run lint --stream --concurrency 1",
```
morgsmccauley, line 16: Why do we need to enforce any concurrency limit at all? Why not just use the default `10` for 🚀 🚀

packages/near-api-js/.gitignore
```diff
@@ -0,0 +1,21 @@
...
+coverage/
+.nyc_output
+src\.ts/protos\.d\.ts
+package\-lock\.json
+
+# OS X
+.DS_Store
+
+# Generated typedoc files
+docs/
```
morgsmccauley, line 21: I can see we still have the root `.gitignore`, is this even needed here? there's nothing specific to `near-api-js`.

morgsmccauley (changes requested):
Good job on this :) Added comments in-line.

packages/tsconfig.json
esaminu: Was initially using to build entire monorepo from root, can be removed now

packages/near-api-js/.eslintrc.yml
esaminu: I think this is useful to establish the convention of thin per package override. It's empty now but I expect it won't be as more packages are added

.eslintignore
esaminu: Yeah prob cleaner to scope this to package. Yeah all these are generated after some scripts e.g. browserify run

.ignore
esaminu: It's for the [`ag` search script](https://github.com/near/near-api-js/commit/c7290650ecd851202e14bc56166f08852686a333). Don't want to change it in case it's being used by a dependency or some tooling unless we figure out what's dependent on it.

.husky/pre-commit
esaminu: I think the idea is to eventually generate the lib on merge or on publish right? Ideally we don't want these files in every diff. Let's sync on this 🙂

package.json
esaminu: Yeah limits should be removed for lint and clean. I'm adding to test though for cleaner output

packages/near-api-js/.eslintrc.yml
morgsmccauley: Wouldn't we want consistent rules across the entire project? I can imagine it becoming quite confusing to have different rule sets across different packages?

esaminu: Cool we can let this pattern evolve if we end up needing it. I'm not sure if we'll need different rules per package but agreed that it would be additional overhead. Let's use one for now

.husky/pre-commit
esaminu: After our sync and researching, I think we can use [the `prepare` script](https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts) for this purpose. We can remove and gitignore the `lib` directory that way and also the `prepublish` script. The source for the `not-in-install` util currently being used in pre publish is the [`in-publish` npm package](https://www.npmjs.com/package/in-publish). The main purpose of this is to ensure the scripts aren't run on npm install which I think isn't necessary for our purposes so removing that too.

cc: @volovyk-s

willemneal: If you are moving to lerna then you can no longer use git as a dependency so I think it's best to remove it altogether.

esaminu: I was thinking about that but looks like it can be added as a sub directory? see https://gitpkg.vercel.app/

willemneal: Nice! I wish I had that when making the near-sdk-as repo!

.husky/pre-commit
```diff
@@ -2,4 +2,3 @@
 . "$(dirname "$0")/_/husky.sh"
 
 yarn build
```
morgsmccauley, line 4: Since we aren't committing `/lib` anymore we shouldn't need this. We can remove the hook all together.

packages/near-api-js/src/account_multisig.ts
```diff
@@ -390,7 +390,7 @@ export class Account2FA extends AccountMultisig {
         return requestId;
     }
 
-    async getCodeDefault(method: any): Promise<string> {
+    async getCodeDefault(): Promise<string> {
         throw new Error('There is no getCode callback provided. Please provide your own in AccountMultisig constructor options. It has a parameter method where method.kind is "email" or "phone".');
     }
```
morgsmccauley, line 396: Why have these types changed?

esaminu: removed the arg since it's unused. This surfaced as part of removing lint errors

PR #909, opened 2022-07-29
Applied engines property on package.json

hcho112:
## Motivation
Provide transparency for the users, which version of node runtime you support

## Description
Updated `engines` property inside package.json with node version greater or equal to 16.16.0 to be enforced. 

## Checklist
- [ ] Added automated tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
   "name": "@near-api-js/monorepo",
   "private": true,
+  "engines": {
+    "node": ">=16.16.0"
+  },
   "workspaces": {
     "packages": [
       "packages/*"


hcho112:
Tested to use node version 14 and attempt to run `yarn` :
image

hcho112:
To all reviewers:
As a first attempt, I have chosen version v16.16.0 because it is an official version to support M1 mac book pro and also well supported on AWS.
https://docs.aws.amazon.com/lambda/latest/dg/nodejs-image.html
https://docs.aws.amazon.com/codebuild/latest/userguide/runtime-versions.html

But happy to change if anyone else has better version to be supported. (with reasons)

morgsmccauley (approved):

PR #910, opened 2022-07-29
NAJ-113-delete-hacks-related-to-error-types

gutsyphilip:
## Motivation
Delete https://github.com/near/near-api-js/blob/master/src/utils/rpc_errors.ts#L88 and other hacks related to Error handling and types as error types are now available on `near-core`.

## Description

- Remove `getErrorTypeFromErrorMessage` and every usage of it.
- Pass error type directly with `response.error.name`
- Remove related tests

## Checklist
- [x] Added automated tests

### DIFF
--- a/packages/near-api-js/test/account.access_key.test.js
+++ b/packages/near-api-js/test/account.access_key.test.js
@@ -40,7 +40,7 @@ test('remove access key no longer works', async() => {
         await contract.setValue({ args: { value: 'test' } });
         fail('should throw an error');
     } catch (e) {
-        expect(e.message).toContain(`Querying [object Object] failed: access key ${publicKey} does not exist while viewing.`);
+        expect(e.message).toContain(`Querying failed: access key ${publicKey} does not exist while viewing.`);
         expect(e.type).toEqual('UntypedError');
     }
 });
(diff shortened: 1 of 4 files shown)


hcho112:
I have rebased latest master and fixed broken test. Can I get some review on this PR?

packages/near-api-js/test/account.access_key.test.js
```diff
@@ -40,8 +40,8 @@ test('remove access key no longer works', async() => {
         await contract.setValue({ args: { value: 'test' } });
         fail('should throw an error');
     } catch (e) {
-        expect(e.message).toEqual(`Can not sign transactions for account ${workingAccount.accountId} on network ${testUtils.networkId}, no matching key pair exists for this account`);
-        expect(e.type).toEqual('KeyNotFound');
+        expect(e.message).toContain(`Querying [object Object] failed: access key ${publicKey} does not exist while viewing.`);
```
morgsmccauley, line 43: `[object Object]` isn't that helpful, should we stringify it instead? That might bloat the error message though, maybe there's a better way we can structure the message overall?

hcho112: Maybe we may leave it as `Query failed: access key ...` since later it prints `${publicKey} does not exist while viewing.`

morgsmccauley: Sounds good 👍🏽

morgsmccauley (approved):

PR #911, opened 2022-08-01
docs revision 1

maxhr:
Docs revision.
- quick reference is now part of main docs site - https://docs.near.org/tools/near-api-js/quick-reference
- fixed README


janewang (approved):

janewang:
If we are removing examples/quick-reference.md, can we add a link to the docs? Can't find it on the docs.

maxhr:
@janewang Quick reference is here - https://docs.near.org/tools/near-api-js/quick-reference
But it's not titled 'quick reference' anymore.

PR #912, opened 2022-08-03
refactor!: Make `walletUrl` required for `Near`

morgsmccauley:
## Motivation
`walletUrl` is currently optional and can cause failures when not defined.

## Description
This PR updates the TS types to make `walletUrl` required.

## Checklist
- [ ] Added automated tests

### DIFF
--- a/packages/near-api-js/src/near.ts
+++ b/packages/near-api-js/src/near.ts
@@ -61,7 +61,7 @@ export interface NearConfig {
      * NEAR wallet url used to redirect users to their wallet in browser applications.
      * @see {@link https://docs.near.org/docs/tools/near-wallet}
      */
-    walletUrl?: string;
+    walletUrl: string;
 
     /**
      * JVSM account ID for NEAR JS SDK
@@ -82,6 +82,10 @@ export class Near {
     readonly accountCreator: AccountCreator;
 
     constructor(config: NearConfig) {
+        if (!config.walletUrl) {
+            throw new Error('walletUrl is required but was not supplied');
+        }
+
         this.config = config;
         this.connection = Connection.fromConfig({
             networkId: config.networkId,
--- a/packages/near-api-js/test/config.js
+++ b/packages/near-api-js/test/config.js
@@ -37,6 +37,7 @@ module.exports = function getConfig(env) {
             networkId: 'shared-test',
             nodeUrl: 'https://rpc.ci-testnet.near.org',
             masterAccount: 'test.near',
+            walletUrl: 'https://wallet.testnet.near.org',
         };
     default:
         throw Error(`Unconfigured environment '${env}'. Can be configured in src/config.js.`);
--- a/packages/near-api-js/test/near.test.js
+++ b/packages/near-api-js/test/near.test.js
@@ -1,0 +1,13 @@
+const { Near } = require('../src/index');
+const getConfig = require('./config');
+
+describe('Near', () => {
+    test('throws if walletUrl is not defined', async() => {
+        expect(() => new Near({})).toThrow('walletUrl is required but was not supplied');
+    });
+
+    test('instantiates successfully', () => {
+        const near = new Near(getConfig('test'));
+        expect(near).toBeDefined();
+    });
+});


esaminu (changes requested):
We need enforcement so that external users can't create a `WalletConnection` without `walletUrl` in their config and tests for this. I don't think we want to make this requirement this general and required for all connections by putting it here and I also think it needs to be runtime not ts

esaminu:
What I was thinking actually was to do the check in [the `WalletConnection` class](https://github.com/near/near-api-js/blob/master/packages/near-api-js/src/wallet-account.ts#L108). The `Near` object is too general and is passed in to many constructors and used for many things aside from connecting to a wallet (e.g. getting and account and sending transactions). Also the api to get a near instance is very un-explicit e.g. `connect(config)`. What do you think of this?

morgsmccauley:
@esaminu After searching the code base I can only see `Near` being passed in to:
* `WalletConnection`
* `connect()` (browser and node)

It doesn't seem to be as widespread unless I am missing something.

I see what you mean about throwing within `WalletConnection` instead but my concern is that it delays when the error will surface, i.e. the developer will have no idea that they have constructed `Near` incorrectly until they have passed it in to `WalletConnection`. I feel like this makes it more likely to go unnoticed.

Why does `Near` even need `walletUrl` if it is only used in `WalletConnection`. I feel the best way forward is to just remove the dependency of `walletUrl` from `Near` and make it required for `WalletConnection` alone.

esaminu:
`Near` is used to get accounts in our cookbook which opens up anything you can do with an account class - and our cookbook uses it without `walletUrl` in the config. It's also used for account creation all over the place.

> Why does Near even need walletUrl if it is only used in WalletConnection? I feel the best way forward is to just remove the dependency of walletUrl from Near and make it required for WalletConnection alone.

One thing I can see us doing is returning a `ConnectedWalletAccount` when calling `Near.account` if a `walletUrl` is present. This is the best solution for now but it makes `near-api-js` opinionated because it doesn't support e.g. ledger or injected wallets though it adds a use for `walletUrl` at the top level. At some point we should bring wallet selector into the monorepo and allow the user to pass in a wallet `id` to the `Near` instance (e.g. ledger) and we would call methods on that [abstract class](https://github.com/near/wallet-selector/blob/main/packages/core/src/lib/wallet/wallet.types.ts#L41-L51) to `signIn`, `signAndSendTransactions` etc from the `Account` instance we return from `Near.account`.

morgsmccauley:
@esaminu 

> Near object is too general and is passed in to many constructors and used for many things aside from connecting to a wallet (e.g. getting and account and sending transactions)

It is for this reason I think `walletUrl` does not make sense within `Near`. `walletUrl` will _only_ be used if it's passed to another specific class constructor. Since `walletUrl` is specific to `WalletConnection` it feels best to remove it from `Near` and add it as a specific config option for `WalletConnection`.

> One thing I can see us doing is returning a ConnectedWalletAccount when calling Near.account if a walletUrl is present.

I feel the best thing is to focus on the requirements right now instead of optimizing for potential future requirements, these can be revisited when committed to. This is why I'm leaning towards the solution proposed above.

Even with the above solution we still have the option of returning a `ConnectedWalletAccount` from `near.account`, it would just be more explicit rather than 'smart', i.e. `near.walletConnectedAccount(accountId, walletUrl)`

esaminu:
Summary of my thoughts here from discussion plus some research:

1. The original intention of having `walletUrl` in the `Near` object was to validate one level deeper in the `WalletConnection` constructor per https://github.com/near/near-api-js/pull/396#issuecomment-687368584 which is the same as the original purpose of this ticket. There was no intention to specifically use it in the `Near` class.
2. The `Near` object is similar to the [`web3.eth` object](https://web3js.readthedocs.io/en/v1.7.5/web3-eth.html) from web3.js.
3. I don’t see any value to removing the `walletUrl` key from near (the first obj arg) and introducing a new second object arg to the `WalletConnection` constructor and adding `walletUrl` to that. The first `Near` object is more well defined and reusable. Plus this adds a bunch of overhead like updating docs and guides and the possibility of having to deprecate the current way of getting a `WalletConnection`.

Suggestions:

1. Throw a clear error in `WalletConnection` constructor if no `walletUrl` exists in the config as originally intended.
2. Add new methods to the `Near` class that return `ConnectedWalletAccount` and throw if no `walletUrl` e.g. `near.walletAccount` and `near.createWalletAccount`

morgsmccauley:
Closing this as the implementation isn't exactly the direction we want to head in. I've left more context on the ticket so this can be picked up again later.

PR #913, opened 2022-08-03
chore: Update `CODEOWNERS`

morgsmccauley:
## Motivation
Thus far core members of NAJ have been assigned to all files, therefore being requested for every pull request. Approvals from code owners are not required/configured and essentially act as notifications for new pulls.

Since assigning all core members to pull requests can be a bit noisy, and we get notifications through a dedicated slack channel anyway, I’m thinking we can be a bit more selective with code owners and instead:
- Assign domain experts to specific areas of the code base
- Require reviews from these owners

Specifically, this helps with locking down our github actions reducing the risk of developers maliciously updating our (eventual) publishing pipeline, but could also be expanded to protect key packages.

## Description
This PR does two things:
- Removes the duplicate `CODEOWNERS` file in the repository root, `.github/CODEOWNERS` will be used instead
- Removes `*` code owners and assigns domain experts to certain packages

Additionally, I'll be updating the repo settings to require PR reviews from code owners.

## Checklist
- [ ] Added automated tests


.github/CODEOWNERS
```diff
@@ -1,4 +1 @@
-# CODEOWNERS: https://help.github.com/articles/about-codeowners/
-
-* @chadoh
-* @vgrichina
+.github/workflows/ @morgsmccauley @andy-haynes
```
esaminu, line 1: I think we need more than 2 people here. Maybe you can add @MaximusHaximus ? I can also volunteer if you're willing to do some knowledge transfer. Also, people in the git history of these files that are still in the org may be useful.

willemneal, line 1:
```suggestion
@morgsmccauley @andy-haynes
```

Should this not be there?

morgsmccauley: This is intentional - I only want this group to be notified when changes are made to `.github/workflows` (our Github Actions)

hcho112 (approved):
LGTM

Issue #627, opened 2021-06-29
Inconsistent API between functionCall and viewFunction.


mattlockyer:
***Describe the bug***
 Same as title.

***To Reproduce***
 Try viewFunction with syntax like this:
 ```
 const token = await contractAccount.viewFunction({
 contractId,
 methodName: 'nft_token',
 args: 

 { token_id, } 

})
 ```

***Expected behavior***
 Expect them both to accept the POJO syntax e.g. 

 { contractId, methodName, args, ...} 

***Screenshots***
 !<span class="error">[image]</span>([https://user-images.githubusercontent.com/321340/123719335-22e1f780-d836-11eb-84f6-0f7f8dffd83c.png](https://user-images.githubusercontent.com/321340/123719335-22e1f780-d836-11eb-84f6-0f7f8dffd83c.png))

zcstarr:
Just going to second this, I hit the same issue. It's killer when your'e trying to figure out what's wrong, you end up rolling the dice and going to the deprecated version.

PR #918, opened 2022-08-05
feat: changed viewFuction to support both POJO and existing arguments

hcho112:
## Motivation
[Issue-627](https://github.com/near/near-api-js/issues/627)
Users wanting to use `viewFunction` call as if they are using `functionCall`.
By passing single object that contains all arguments as props. This is actually standard practice compared to passing multiple props in specific order. 

## Description
Created adapter layer where depends on the format of given arguments, it either deconstruct from object (new approach) or passing given arguments as it is (existing approach)

## Checklist
- [x] Added automated tests

### DIFF
--- a/packages/near-api-js/src/account.ts
+++ b/packages/near-api-js/src/account.ts
@@ -482,7 +482,26 @@ export class Account {
      * @param options.blockQuery specifies which block to query state at. By default returns last "optimistic" block (i.e. not necessarily finalized).
      * @returns {Promise<any>}
      */
-    async viewFunction(
+
+    async viewFunction(obj : any, ...restArgs: any[]) {
+        if (restArgs.length === 0) {
+            const {
+                contractId,
+                methodName,
+                args,
+                options,
+            } = obj;
+            return await this.viewFunctionCall(contractId, methodName, args, options);
+        } else {
+            const contractId = obj;
+            const methodName = restArgs[0];
+            const args = restArgs[1];
+            const options = restArgs[2];
+            return await this.viewFunctionCall(contractId, methodName, args, options);
+        }
+    }
+
+    async viewFunctionCall(
         contractId: string,
         methodName: string,
         args: any = {},
--- a/packages/near-api-js/src/account.ts
+++ b/packages/near-api-js/src/account.ts
@@ -1,4 +1,5 @@
 import BN from 'bn.js';
+import depd from 'depd';
 import {
     transfer,
     createAccount,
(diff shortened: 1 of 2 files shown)


packages/near-api-js/src/account.ts
```diff
@@ -482,7 +482,26 @@ export class Account {
      * @param options.blockQuery specifies which block to query state at. By default returns last "optimistic" block (i.e. not necessarily finalized).
      * @returns {Promise<any>}
      */
-    async viewFunction(
```
esaminu, line 485: I think the best approach here may be to create a `viewFunctionV2` with the new functionality and always call that similar to [this change set](https://github.com/near/near-api-js/commit/a05667361ddab208423f5a5c0e3dd0ce182b9880#diff-ae151444b91b7e5e0a6aa753024f7900f0ff0447afaa2d18ad6619d5c3e12318L156-R224) when signAndSendTransactions was converted to having an object argument. We likely should deprecate the current way of calling `viewFunction` too.  cc @janewang

packages/near-api-js/src/account.ts
```diff
@@ -482,7 +482,26 @@ export class Account {
...
+
+    async viewFunction(obj : any, ...restArgs: any[]) {
```
esaminu, line 486: For the types I think we should reuse the [FunctionCallOptions](https://github.com/near/near-api-js/commit/a05667361ddab208423f5a5c0e3dd0ce182b9880#diff-ae151444b91b7e5e0a6aa753024f7900f0ff0447afaa2d18ad6619d5c3e12318R81) type. 

The best way to do this imo is to have a general `FunctionCallOptions` type, remove `walletMeta` and `walletCallbackUrl` and then create a `ChangeFunctionCallOptions` type and a `ViewFunctionCallOptions` type since everything in the `options` arg of `viewFunction` is in `FunctionCallOptions` except for `parse` and `blockQuery`:

```typescript
// using the same FunctionCallOptions with walletCallbackUrl and walletMeta removed
type ChangeFunctionCallOptions = FunctionCallOptions & { walletMeta?: string; walletCallbackUrl?: string; };
type ViewFunctionCallOptions = FunctionCallOptions & { parse?: (response: Uint8Array) => any; blockQuery?: BlockReference; };
```
We just need to maintain the same default arg logic for all args as currently in the `viewFunction` block, maintain the jsdoc we have for `walletMeta` and `walletCallbackUrl` and change `Account.functionCall` to use the `ChangeFunctionCallOptions` type. After this is done `viewFunctionV2` can have the `ViewFunctionCallOptions` type and v1 can have the same type signature as currently. Wdyt?

esaminu (changes requested):
Good stuff! This is a needed change. I left some comments inline

packages/near-api-js/src/account.ts
hcho112: Thank you for amazing review, I totally agree with all the suggestions.

esaminu (approved):
LGTM 💯 ! I think we need to update all tests to use the new type but this can be done as part of the removal of the deprecated `viewFunction`

Could you add an issue to our JIRA backlog for the removal of the deprecated `viewFunction` and add updating all tests to the as one of the criteria?

PR #926, opened 2022-08-11
fix: NAJ-61

AbdulBaseerKhan:
## Motivation
NAJ-61

## Description
This case can occurred (typeof result.status.Failure === 'object' can process on Failure object in case of null). Fixed!

## Checklist
- [ ] Added automated tests

### DIFF
--- a/packages/near-api-js/src/account.ts
+++ b/packages/near-api-js/src/account.ts
@@ -253,6 +253,7 @@ export class Account {
         }, []);
         this.printLogsAndFailures(signedTx.transaction.receiverId, flatLogs);
 
+       // Should be falsy if result.status.Failure is null
         if (!returnError && typeof result.status === 'object' && typeof result.status.Failure === 'object'  && result.status.Failure !== null) {
             // if error data has error_message and error_type properties, we consider that node returned an error in the old format
             if (result.status.Failure.error_message && result.status.Failure.error_type) {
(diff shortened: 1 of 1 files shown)


packages/near-api-js/src/account.ts
```diff
@@ -253,7 +253,7 @@ export class Account {
         }, []);
         this.printLogsAndFailures(signedTx.transaction.receiverId, flatLogs);
 
-        if (!returnError && typeof result.status === 'object' && typeof result.status.Failure === 'object') {
+        if (!returnError && typeof result.status === 'object' && typeof result.status.Failure === 'object'  && result.status.Failure !== null) {
```
esaminu, line 256: The number of `result.status.Failure` that pass this check go down after this change. Do you mean to use `||`?

AbdulBaseerKhan: @esaminu 
No here we have to use `&&` because `typeof null` is 'object'. `result.status.Failure = null` means there is no Failure so it should not go to error block.

esaminu: Thanks for clarifying! Can we add something like this?

```suggestion
       // Should be falsy if result.status.Failure is null
        if (!returnError && typeof result.status === 'object' && typeof result.status.Failure === 'object'  && result.status.Failure !== null) {
```
Or it would be even better if we can mock and test this but not necessary if we don't already have tests that mock this imo.

marcinbodnar (changes requested):
@AbdulBaseerKhan Please, add a comment that @esaminu proposed and rebased PR so it will be ready for approval. Thank you.

AbdulBaseerKhan:
@marcinbodnar done

marcinbodnar (approved):
LGTM

Issue #325, opened 2020-05-19
provide a better wallet_auth_key prefix


amgando:
currently wallet uses a user-defined prefix for the key in LocalStorage

note:

***undefined_*** wallet_auth_key

```
 {
 "undefined_wallet_auth_key": 

 {"accountId":"sherif-demo.testnet","allKeys":["ed25519:Azqg7VBLGYo8LdiJ4SEzgodg3JXANoJm8jbH1homoVj6"]} 

}
 ```

this could be improved with a more friendly default value for `appKeyPrefix `

[https://github.com/near/near-api-js/blob/25e7e60203965aee062ca19f352ff765859896ef/src/wallet-account.ts#L28](https://github.com/near/near-api-js/blob/25e7e60203965aee062ca19f352ff765859896ef/src/wallet-account.ts#L28)

mikedotexe:
@stefanopepe does this sound like a change that will break Wallet or have concerns with this?

stefanopepe:
> @stefanopepe does this sound like a change that will break Wallet or have concerns with this?

We are re-designing the key management, and we'll start working on it in the next sprint. The UX will probably include a way for the user to append notes/memorable information to each key, but I don't have more information to share right now.

Roping in @MaximusHaximus for an estimation of the changes

MaximusHaximus:
I did an audit, and I don't believe that we reference this `_wallet_auth_key` localstorage value anywhere in the wallet codebase directly.  As a result, as long as long as `near-api-js` handles backwards compatibility internally for anything that relies on it in `near-api-js`, it shouldn't have any impact on the wallet as far as I can tell. :)

PR #927, opened 2022-08-12
Fixed appKeyPrefix undefined value

erditkurteshiSQA:
## Motivation
It fixes issue #325 

## Description
Describe the changes made in this PR. How does it solve the above?
It fixes the appKeyPrefix by checking it's coming as undefined.

## Checklist
- [ ] Added automated tests

### DIFF
--- a/packages/near-api-js/src/wallet-account.ts
+++ b/packages/near-api-js/src/wallet-account.ts
@@ -108,8 +108,8 @@ export class WalletConnection {
 
         // Migrate dApps that previously used a faulty authDataKey.
         if (DEPRECATED_authData) {
-            window.localStorage.setItem(authDataKey, JSON.stringify(DEPRECATED_authData));
             window.localStorage.removeItem(DEPRECATED_authDataKey);
+            window.localStorage.setItem(authDataKey, JSON.stringify(DEPRECATED_authData));
         }
 
         const authData = JSON.parse(window.localStorage.getItem(authDataKey));
--- a/packages/near-api-js/src/wallet-account.ts
+++ b/packages/near-api-js/src/wallet-account.ts
@@ -83,16 +83,16 @@ export class WalletConnection {
     _completeSignInPromise: Promise<void>;
 
     constructor(near: Near, appKeyPrefix: string | null) {
-        if(typeof window === 'undefined') {
+        if (typeof window === 'undefined') {
             return new Proxy(this, {
                 get(target, property) {
-                    if(property === 'isSignedIn') {
+                    if (property === 'isSignedIn') {
                         return () => false;
                     }
-                    if(property === 'getAccountId') {
+                    if (property === 'getAccountId') {
                         return () => '';
                     }
-                    if(target[property] && typeof target[property] === 'function') {
+                    if (target[property] && typeof target[property] === 'function') {
                         return () => {
                             throw new Error('No window found in context, please ensure you are using WalletConnection on the browser');
                         };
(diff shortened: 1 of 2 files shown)


packages/near-api-js/src/wallet-account.ts
```diff
@@ -102,7 +102,16 @@ export class WalletConnection {
...
         this._near = near;
-        const authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
+        const DEPRECATED_authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
+        const DEPRECATED_authData = JSON.parse(window.localStorage.getItem(DEPRECATED_authDataKey));
+        const authDataKey = (appKeyPrefix || near.config.contractName || 'default') + LOCAL_STORAGE_KEY_SUFFIX;
+
+        // Migrate dApps that previously used a faulty authDataKey.
+        if (DEPRECATED_authData) {
+            window.localStorage.setItem(authDataKey, JSON.stringify(DEPRECATED_authData));
+            window.localStorage.removeItem(DEPRECATED_authDataKey);
```
esaminu, line 112: Should we remove before adding the new data? I'm concerned about cases where `DEPRECATED_authDataKey` and `authDataKey` have the same value

esaminu (commented):
Great stuff! 🚀 

I left a comment regarding migrating folks with faulty key (thanks for handling that)

Also, could you fix the broken assertion on https://github.com/near/near-api-js/blob/master/packages/near-api-js/test/wallet-account.test.js#L173

esaminu (approved):
LGTM 💯

packages/near-api-js/src/wallet-account.ts
```diff
@@ -102,7 +102,16 @@ export class WalletConnection {
             });
         }
         this._near = near;
-        const authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
+        const DEPRECATED_authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
```
MaximusHaximus, line 105: Should this case explicitly check for `'undefined' + LOCAL_STORAGE_KEY_SUFFIX`?  Otherwise if someone has passed an actual `appKeyPrefix`, we will be unnecessarily deleting the localstorage entry and re-creating it _every time this code runs_, right?

packages/near-api-js/test/wallet-account.test.js
```diff
@@ -154,12 +155,50 @@ it('can complete sign in', async () => {
...
 
     expect(await keyStore.getKey('networkId', 'near.account')).toEqual(keyPair);
-    expect(localStorage.getItem('contractId_wallet_auth_key'));
+    expect(localStorage.getItem('contractId_wallet_auth_key')).toBeTruthy();
     expect(history.slice(1)).toEqual([
         [{}, 'documentTitle', 'http://example.com/location']
     ]);
 });
 
+it('can get the default appKeyPrefix', async () => {
```
MaximusHaximus, line 164: 1. There needs to be a test that verifies that a migration from an existing 'undefined' key successfully completes when data was already stored in the incorrect key.
2. There needs to be a test that verifies if `undefined_LOCAL_STORAGE_KEY_SUFFIX` is not set to any value that the code gracefully succeeds at initializing the localstorage correctly.  This might require splitting the test case into a separate file (which may allow us to take advantage of Jest's per-file test isolation) or refactoring the test file to ensure that localstorage is empty when we construct our NEAR object.

MaximusHaximus (commented):
@esaminu I've got some concerns about this PR - can you take a look at my comments and see if they are legitimate problems?  If so, let's roll this back until we can be 100% sure we've got complete test coverage and have tested the code in a variety of real-world dapp scenarios.

packages/near-api-js/src/wallet-account.ts
esaminu: Agreed 👍 , I think adding a `&& DEPRECATED_authDataKey !=  authDataKey` reads nicely here

packages/near-api-js/src/wallet-account.ts
```diff
@@ -102,7 +102,16 @@ export class WalletConnection {
...
+        const DEPRECATED_authData = JSON.parse(window.localStorage.getItem(DEPRECATED_authDataKey));
```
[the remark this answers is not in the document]
esaminu: Doesn't throw on my side since `window.localStorage.getItem(DEPRECATED_authDataKey)` is `null` and `JSON.parse(null)` is fine but we should have a fallback here or a catch block still imo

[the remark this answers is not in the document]
MaximusHaximus: Actually, it seems like this will be fine, per https://html.spec.whatwg.org/multipage/webstorage.html#dom-storage-getitem since getItem will return `null` rather than `undefined` :+1:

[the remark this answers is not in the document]
esaminu: Thanks, since it's spec we'll leave as is 🙂

PR #928, opened 2022-08-12
feat!: console usage controllable by the caller

AbdulBaseerKhan:
## Motivation
NAJ-64

## Description
Simple customisable logger added and all the internal logging of api should be done by it and have exposed the configureLogging so the end user can control logging accordingly.

## Checklist
- [x] Added automated tests

### DIFF
--- a/packages/near-api-js/test/account.test.js
+++ b/packages/near-api-js/test/account.test.js
@@ -3,6 +3,7 @@ const { Account, Contract, providers } = require('../src/index');
 const testUtils  = require('./test-utils');
 const fs = require('fs');
 const BN = require('bn.js');
+const { Logger, configureLogging } = require('../src/utils/near-logger');
 
 let nearjs;
 let workingAccount;
@@ -12,12 +13,14 @@ const { HELLO_WASM_PATH, HELLO_WASM_BALANCE } = testUtils;
 jasmine.DEFAULT_TIMEOUT_INTERVAL = 50000;
 
 beforeAll(async () => {
+    configureLogging({ showLogs: () => false });
     nearjs = await testUtils.setUpTestConnection();
     workingAccount = await testUtils.createAccount(nearjs);
 });
 
 afterAll(async () => {
     await workingAccount.deleteAccount(workingAccount.accountId);
+    Logger.reset();
 });
 
 test('view pre-defined account works and returns correct name', async () => {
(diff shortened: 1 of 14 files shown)


packages/near-api-js/test/account.test.js
```diff
@@ -75,19 +78,21 @@ test('findAccessKey returns the same access key when fetched simultaneously', as
...
-        console.log = function () {
-            logs.push(Array.from(arguments).join(' '));
-        };
+        configureLogging({
+            showLogs: () => true,
+            onLog: (args) => {
+                logs.push(Array.from(args).join(' '));
+            },
+            onWarn: () => {}, // Do nothing onWarn
+        });
```
esaminu, line 91: I think you need to use spread in `args` here to recreate the same behavior as before

packages/near-api-js/test/account.test.js
```diff
@@ -113,15 +117,18 @@ describe('with deploy contract', () => {
...
```
esaminu, line 127: Same here, need to spread `args` in the `onLog` property to recreate old behavior

packages/near-api-js/test/promise.test.js
```diff
@@ -28,15 +33,17 @@ describe('with promises', () => {
...
         logs = [];
-        console.log = function() {
-            logs.push(Array.from(arguments).join(' '));
-        };
+        configureLogging({
+            showLogs: () => true,
+            onLog: (args) => {
+                logs.push(Array.from(args).join(' '));
+            }, onWarn: () => {}
+        });
```
esaminu, line 42: Same as above with `args` spread in the `onLog` property

esaminu (commented):
Looking good! I left some comments and a rebase is required, then we can land this 💯

esaminu:
@AbdulBaseerKhan is this a breaking change btw? I see the PR title but I'm not sure how it's a breaking change 🙂

marcinbodnar:
@AbdulBaseerKhan Could you please respond to Osman's comments? Thank you.

packages/near-api-js/test/account.test.js
AbdulBaseerKhan: @esaminu I have tested this behaviour and was working as the previous one.

esaminu: It looks like it wouldn't work if multiple arguments are used. Does it still work in this case?

denbite:
the idea of having an extendable logger was implemented in https://github.com/near/near-api-js/pull/1215

I'd suggest closing this one so we don't leave it out of date

Issue #220, opened 2020-02-07
Nonce type inconsistency.


bowenwang1996:
When constructing a transaction we pass `nonce` as a `number` but when decoding the type of `nonce` becomes `BN` since `number` doesn't actually cover `u64`. It would be nice if the type of `nonce` is consistent on both end.

Decision: update type to a string. Add deprecation warning. [Daryl Collins](https://nearinc.atlassian.net/secure/ViewProfile.jspa?accountId=6139108a24ba8b00701167be)

vgrichina:
I think it's bigger issue generally – basically almost every place that takes `BN` usually also allows `number` and `string`. We should formalize it by introducing `BNLike` type or smth like that and making sure conversion always happens without precision loss.

PR #933, opened 2022-08-16
Migrating nonce type from number to string

hcho112:
## Motivation
#220 

## Description
This PR contains implementations to migrate number typed nonce to string.
Added a string type check to throw error if `nonce` with `number` is given.
Updated all existing tests

## Checklist
- [ ] Added automated tests

### DIFF
--- a/packages/near-api-js/src/account.ts
+++ b/packages/near-api-js/src/account.ts
@@ -21,6 +21,7 @@ import {
     ViewStateResult,
     AccountView,
     AccessKeyView,
+    AccessKeyViewRaw,
     CodeResult,
     AccessKeyList,
     AccessKeyInfoView,
@@ -202,7 +203,7 @@ export class Account {
         const block = await this.connection.provider.block({ finality: 'final' });
         const blockHash = block.header.hash;
 
-        const nonce = ++accessKey.nonce;
+        const nonce = accessKey.nonce.add(new BN(1));
         return await signTransaction(
             receiverId, nonce, actions, baseDecode(blockHash), this.connection.signer, this.accountId, this.connection.networkId
         );
@@ -294,13 +295,18 @@ export class Account {
         }
 
         try {
-            const accessKey = await this.connection.provider.query<AccessKeyView>({
+            const rawAccessKey = await this.connection.provider.query<AccessKeyViewRaw>({
                 request_type: 'view_access_key',
                 account_id: this.accountId,
                 public_key: publicKey.toString(),
                 finality: 'optimistic'
             });
 
+            // store nonce as BN to preserve precision on big number
+            const accessKey = {
+                ...rawAccessKey,
+                nonce: new BN(rawAccessKey.nonce),
+            }
             // this function can be called multiple times and retrieve the same access key
             // this checks to see if the access key was already retrieved and cached while
             // the above network call was in flight. To keep nonce values in line, we return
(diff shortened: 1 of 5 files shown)


esaminu (approved):
LGTM 🚀  

Can we use conventional commit syntax on the PR description and mark this as a breaking change?

hcho112:
wait sorry lets not merge this yet. I had deep dive discussion with @MaximusHaximus and it is now essential to make this on this PR.

hcho112:
After my discovery on the fact that RPC is returning `nonce` on `accessKey` as number, I have re-write this PR to cater current implementation as well as future proof. (by wrapping `nonce` with `BN` so that it is guaranteed to operate consistently when it's a number or string is given)

packages/near-api-js/src/account.ts
```diff
@@ -579,13 +585,15 @@ export class Account {
...
             finality: 'optimistic'
         });
+        // Replace raw nonce into a new BN
+        const newResponse = { keys: response?.keys?.map((key) => ({ ...key, access_key: { ...key.access_key, nonce: new BN(key.access_key.nonce) } }))};
         // A breaking API change introduced extra information into the
         // response, so it now returns an object with a `keys` field instead
         // of an array: https://github.com/nearprotocol/nearcore/pull/1789
-        if (Array.isArray(response)) {
-            return response;
+        if (Array.isArray(newResponse)) {
```
morgsmccauley, line 593: This check seems irrelevant now that `newResponse` is always an object. Do we need additional logic here to cater for `view_access_key_list` returning either an array or an object, or can we trust that it will always be an object of the shape `{ keys: [] }`

packages/near-api-js/src/account.ts
```diff
@@ -293,13 +294,18 @@ export class Account {
         }
 
         try {
-            const accessKey = await this.connection.provider.query<AccessKeyView>({
+            const rawAccessKey = await this.connection.provider.query<AccessKeyViewRaw>({
```
morgsmccauley, line 297: Would it be better to place this coercion logic within `JsonRpcProvider`? `nonce` is only really used in this file, but if we were to use these same `query` `methods` else where, we'd most likely have to duplicate this logic. Similarly, someone adding a new `query` type in `JsonRpcProvider` which contains `nonce` may not know that coercion is necessary because it happens within this file.

We'd have to inspect the incoming `request_type` which doesn't seem ideal, but I think the centralization of it would be beneficial.

Thoughts?

morgsmccauley (changes requested):
Mostly looks good, just a couple questions :)

packages/near-api-js/src/account.ts
hcho112: First of all, changes on this PR is based on the mindset of `nonce` should be returned as a `string` NOT a `number`. (It is currently returning as a number) The problem with returning with number is that, javascript `number` type is less than what max `u64` can be as an integer. So at some point, it will be essential for RPC to make this change for the sake of precision on `nonce`. 

So, changes on this PR allows to support both `number` and `string` type of nonce where it will become a future proof regardless when RPC is making the `nonce` type change. So assuming that `nonce` is returned as `string` that is larger than js `number` integer, in order to increment, we can't just use `++`. So it should still need to be wrapped with `new BN`, wherefore this logic is actually unavoidable.

morgsmccauley: What I am suggesting here is we move this logic in to `JsonRpcProvider` so that it is encapsulated and will be reused if these queries are ever used elsewhere.

But on second thought it may be better to leave this as is, and deal with that case if/when we need.

morgsmccauley (approved):
LGTM

PR #935, opened 2022-08-22
Update viewFunction with only accepting single object as argument

hcho112:
## Motivation
This PR is removing old syntax that is expected to be deprecated. Since this PR contains BREAKING CHANGE, prior to merge, please follow up with team to confirm correct schedule. 

## Description
This PR will update `viewFunction` in `account.ts` to only support single object as argument. (Deprecating on supporting old convention of accepting array of arguments)
Also updates existing tests

## Checklist
- [ ] Added automated tests

### DIFF
--- a/.changeset/great-jobs-destroy.md
+++ b/.changeset/great-jobs-destroy.md
@@ -1,0 +1,6 @@
+---
+"@near-js/cookbook": minor
+"near-api-js": major
+---
+
+Update `viewFunction` on account model to only accept single object as argument
--- a/.changeset/great-jobs-destroy.md
+++ b/.changeset/great-jobs-destroy.md
@@ -1,6 +1,17 @@
 ---
-"@near-js/cookbook": minor
 "near-api-js": major
 ---
 
-Update `viewFunction` on account model to only accept single object as argument
+`account.viewFunction` now takes a single object argument rather than individual args. Callsites will now need to be updated like so:
+```diff
+-await account.viewFunction(
+-  'wrap.near',
+-  'storage_balance_of',
+-  { account_id: 'example.near' }
+-);
++await account.viewFunction({
++  contractId: 'wrap.near',
++  methodName: 'storage_balance_of',
++  args: { account_id: 'example.near' },
++});
+```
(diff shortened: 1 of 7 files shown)


packages/near-api-js/src/account.ts
```diff
@@ -483,40 +483,17 @@ export class Account {
      * @param contractId NEAR account where the contract is deployed
```
morgsmccauley, line 483: This JSDoc is configured for separate arguments, now that `viewFunction` expects a single object argument I believe we need something similar to the previous `options` - https://jsdoc.app/tags-param.html

morgsmccauley (commented):
Looks good, just one comment

morgsmccauley (approved):

morgsmccauley (dismissed):

.changeset/great-jobs-destroy.md
```diff
@@ -0,0 +1,6 @@
+---
+"@near-js/cookbook": minor
```
morgsmccauley, line 2: 🔪

.changeset/great-jobs-destroy.md
```diff
@@ -0,0 +1,6 @@
...
+"near-api-js": major
+---
+
+Update `viewFunction` on account model to only accept single object as argument
```
morgsmccauley, line 6: Because this is a breaking change it would pay to add more detail here, including migration info. What do you think of the following:

`account.viewFunction` now takes a single object argument rather than individual args. Callsites will now need to be updated like so:
```diff
-await account.viewFunction(
-  'wrap.near',
-  'storage_balance_of',
-  { account_id: 'example.near' }
-);
+await account.viewFunction({
+  contractId: 'wrap.near',
+  methodName: 'storage_balance_of',
+  args: { account_id: 'example.near' },
+});
```

morgsmccauley (changes requested):
Just a couple comments around the included Changeset.

Also, because this is a major bump, let's hold off on merging this. It might be better to publish the minor/patch changes first so that they aren't lumped in to the major change. That would give devs more granularity in terms of upgrading.

morgsmccauley (dismissed):

morgsmccauley (approved):

Issue #416, opened 2020-09-29
Some uncommented key in the source code


miohtama:
There is some test key hardcoded in the source code:

[https://github.com/near/near-api-js/blob/master/test/test-utils.js#L19](https://github.com/near/near-api-js/blob/master/test/test-utils.js#L19)

I suggest if you hardcore artefacts in the source then leave a comment on what and why they are there, so the source code readers have better context working with near-api-js.

PR #936, opened 2022-08-22
Fix/naj 62

erditkurteshiSQA:
## Motivation
It fixes issue #416 

## Description
Commented and explained the hardcoded public key.

## Checklist
- [ ] Added automated tests


esaminu:
@erditkurteshiSQA this is a FAK on `ci-testnet` actually which is a dedicated rpc for tests. We should start experimenting with local sandbox testing via [near/workspaces-js](https://github.com/near/workspaces-js) and then eventually move to that and remove the key. 

This may not be in scope here so feel free to just update the comment but let's create a ticket for the workspaces / local sandbox migration.

cc @volovyks @MaximusHaximus

esaminu (approved):

PR #947, opened 2022-08-25
appKeyPrefix undefined value

erditkurteshiSQA:
## Motivation
Follow up PR for #927

## Description
It fixes issue https://github.com/near/near-api-js/issues/325

## Checklist
- [ ] Added automated tests

### DIFF
--- a/packages/near-api-js/src/wallet-account.ts
+++ b/packages/near-api-js/src/wallet-account.ts
@@ -106,10 +106,12 @@ export class WalletConnection {
         this._near = near;
         const DEPRECATED_authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
         const DEPRECATED_authData = JSON.parse(window.localStorage.getItem(DEPRECATED_authDataKey));
-        const authDataKey = (appKeyPrefix || near.config.contractName || 'default') + LOCAL_STORAGE_KEY_SUFFIX;
+
+        appKeyPrefix = appKeyPrefix || near.config.contractName || 'default';
+        const authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
 
         // Migrate dApps that previously used a faulty authDataKey.
-        if (DEPRECATED_authData) {
+        if (DEPRECATED_authData && DEPRECATED_authDataKey != authDataKey) {
             window.localStorage.removeItem(DEPRECATED_authDataKey);
             window.localStorage.setItem(authDataKey, JSON.stringify(DEPRECATED_authData));
         }
@@ -117,7 +119,6 @@ export class WalletConnection {
         const authData = JSON.parse(window.localStorage.getItem(authDataKey));
         this._networkId = near.config.networkId;
         this._walletBaseUrl = near.config.walletUrl;
-        appKeyPrefix = appKeyPrefix || near.config.contractName || 'default';
         this._keyStore = (near.connection.signer as InMemorySigner).keyStore;
         this._authData = authData || { allKeys: [] };
         this._authDataKey = authDataKey;
--- a/packages/near-api-js/src/wallet-account.ts
+++ b/packages/near-api-js/src/wallet-account.ts
@@ -104,16 +104,17 @@ export class WalletConnection {
             });
         }
         this._near = near;
-        const DEPRECATED_authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
-        const DEPRECATED_authData = JSON.parse(window.localStorage.getItem(DEPRECATED_authDataKey));
-
-        appKeyPrefix = appKeyPrefix || near.config.contractName || 'default';
-        const authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
-
-        // Migrate dApps that previously used a faulty authDataKey.
-        if (DEPRECATED_authData && DEPRECATED_authDataKey != authDataKey) {
-            window.localStorage.removeItem(DEPRECATED_authDataKey);
-            window.localStorage.setItem(authDataKey, JSON.stringify(DEPRECATED_authData));
+        const DEPRECATED_appKeyPrefix = appKeyPrefix;
+        const finalAppKeyPrefix = appKeyPrefix || near.config.contractName || 'default';
+        const DEPRECATED_authDataKey = DEPRECATED_appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
+        const authDataKey = finalAppKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
+        
+        if (DEPRECATED_authDataKey != authDataKey) {
+            const DEPRECATED_authData = JSON.parse(window.localStorage.getItem(DEPRECATED_authDataKey));
+            if (DEPRECATED_authData) {
+                window.localStorage.removeItem(DEPRECATED_authDataKey);
+                window.localStorage.setItem(authDataKey, JSON.stringify(DEPRECATED_authData));
+            }
         }
 
         const authData = JSON.parse(window.localStorage.getItem(authDataKey));
--- a/packages/near-api-js/src/wallet-account.ts
+++ b/packages/near-api-js/src/wallet-account.ts
@@ -110,8 +110,8 @@ export class WalletConnection {
 
         // Migrate dApps that previously used a faulty authDataKey.
         if (DEPRECATED_authData) {
-            window.localStorage.setItem(authDataKey, JSON.stringify(DEPRECATED_authData));
             window.localStorage.removeItem(DEPRECATED_authDataKey);
+            window.localStorage.setItem(authDataKey, JSON.stringify(DEPRECATED_authData));
         }
 
         const authData = JSON.parse(window.localStorage.getItem(authDataKey));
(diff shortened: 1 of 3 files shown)


packages/near-api-js/src/wallet-account.ts
```diff
@@ -102,11 +102,21 @@ export class WalletConnection {
...
+
+        appKeyPrefix = appKeyPrefix || near.config.contractName || 'default';
         const authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
+
+        // Migrate dApps that previously used a faulty authDataKey.
+        if (DEPRECATED_authData && DEPRECATED_authDataKey != authDataKey) {
+            window.localStorage.removeItem(DEPRECATED_authDataKey);
+            window.localStorage.setItem(authDataKey, JSON.stringify(DEPRECATED_authData));
+        }
+
```
esaminu, line 116: Just to make this logic easier to follow:

```suggestion
        const DEPRECATED_appKeyPrefix = appKeyPrefix;
        const finalAppKeyPrefix = appKeyPrefix || near.config.contractName || 'default';

        const DEPRECATED_authDataKey = DEPRECATED_appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
        const authDataKey = finalAppKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;

        if (DEPRECATED_authDataKey != authDataKey) {
            const DEPRECATED_authData = JSON.parse(window.localStorage.getItem(DEPRECATED_authDataKey));
            if (DEPRECATED_authData) {
                window.localStorage.removeItem(DEPRECATED_authDataKey);
                window.localStorage.setItem(authDataKey, JSON.stringify(DEPRECATED_authData));
            }
        }
```

packages/near-api-js/test/wallet-account.test.js
```diff
@@ -17,6 +17,7 @@ const nearApi = require('../src/index');
```
esaminu, line 17: Can we add tests for the assertions in https://github.com/near/near-api-js/pull/927/files/40e049b3ee19c00bd80d21c0890f84da2ab04d44#r953151759 by setting things in localStorage before constructing a `WalletConnection` instance?

Also do we have to reset localStorage before each test?

esaminu (changes requested):
Almost there. Added some comments about readability and some additional assertions 👍

packages/near-api-js/test/wallet-account.test.js
```diff
@@ -154,12 +155,63 @@ it('can complete sign in', async () => {
...
+        connection: {
+            networkId: 'networkId',
+            signer: new nearApi.InMemorySigner(keyStore)
+        },
+        account() {
+            return {
+                state() { }
+            };
+        }
+    };
```
esaminu, line 179: Can we do this in `beforeEach` just like `nearFake`? Also this is more concise:
```suggestion
    nearFakeWithoutContract = {...nearFake, config: {...nearFake.config, contractName: undefined}};
```

packages/near-api-js/test/wallet-account.test.js
```diff
@@ -154,12 +155,63 @@ it('can complete sign in', async () => {
...
+
+    const keyPair = nearApi.KeyPair.fromRandom('ed25519');
+    global.window.location.href = `http://example.com/location?account_id=near.account&public_key=${keyPair.publicKey}`;
```
esaminu, line 182: Can we add `all_keys` to the url here and make equality assertion on L188?

```suggestion
    global.window.location.href = `http://example.com/location?account_id=near.account&public_key=${keyPair.publicKey}&all_keys=${keyPair.toString()}`;
```

Then we can do: 

```js
expect(localStorage.getItem('default_wallet_auth_key')).toEqual(JSON.stringify({accountId: 'near.account', allKeys: [keyPair.toString()]}));
```

packages/near-api-js/test/wallet-account.test.js
```diff
@@ -154,12 +155,63 @@ it('can complete sign in', async () => {
...
+
+it('can get the near-app appKeyPrefix', async () => {
+    const keyPair = nearApi.KeyPair.fromRandom('ed25519');
+    global.window.location.href = `http://example.com/location?account_id=near.account&public_key=${keyPair.publicKey}`;
+    await keyStore.setKey('networkId', 'pending_key' + keyPair.publicKey, keyPair);
+
+    const newWalletConnection = new nearApi.WalletConnection(nearFake, 'near-app');
+    await newWalletConnection._completeSignInWithAccessKey();
+
+    expect(localStorage.getItem('near-app_wallet_auth_key')).toBeTruthy();
```
esaminu, line 199: Same comment as above re equality assertion

packages/near-api-js/test/wallet-account.test.js
```diff
@@ -154,12 +155,63 @@ it('can complete sign in', async () => {
...
+    global.window.location.href = `http://example.com/location?account_id=near.account&public_key=${keyPair.publicKey}`;
+    await keyStore.setKey('networkId', 'pending_key' + keyPair.publicKey, keyPair);
+
+    localStorage.setItem('undefined_wallet_auth_key');
+
+    const newWalletConnection = new nearApi.WalletConnection(nearFake, 'near-app');
+    await newWalletConnection._completeSignInWithAccessKey();
+
+    expect(localStorage.getItem('near-app_wallet_auth_key')).toBeTruthy();
+});
```
esaminu, line 213: This test does not execute migration since `authDataKey` is `near-app_wallet_auth_key` when the `WalletConnection` is constructed which is the same value as `DEPRECATED_authDataKey`. The following block would test it. Please check if there are other cases we are missing:
```suggestion
describe('can migrate from an existing appKeyPrefix successfully when not defined', () => {
    it('migrates to default_wallet_auth_key when no contract in near', async () => {
        const keyPair = nearApi.KeyPair.fromRandom('ed25519');
        await keyStore.setKey('networkId', 'pending_key' + keyPair.publicKey, keyPair);

        localStorage.setItem('undefined_wallet_auth_key', JSON.stringify({accountId: 'near.account', allKeys: [keyPair.toString()]}));

        new nearApi.WalletConnection(nearFakeWithoutContract);

        expect(localStorage.getItem('undefined_wallet_auth_key')).toBeNull();
        expect(localStorage.getItem('default_wallet_auth_key')).toEqual(JSON.stringify({accountId: 'near.account', allKeys: [keyPair.toString()]}));
    });

    it('migrates to contractId_wallet_auth_key when contract exists in near', async () => {
        const keyPair = nearApi.KeyPair.fromRandom('ed25519');
        await keyStore.setKey('networkId', 'pending_key' + keyPair.publicKey, keyPair);

        localStorage.setItem('undefined_wallet_auth_key', JSON.stringify({accountId: 'near.account', allKeys: [keyPair.toString()]}));

        new nearApi.WalletConnection(nearFake);

        expect(localStorage.getItem('undefined_wallet_auth_key')).toBeNull();
        expect(localStorage.getItem('contractId_wallet_auth_key')).toEqual(JSON.stringify({accountId: 'near.account', allKeys: [keyPair.toString()]}));
    });
});
```

esaminu (changes requested):
Looks good, added some comments re testing 👍

packages/near-api-js/src/wallet-account.ts
```diff
@@ -104,11 +104,22 @@ export class WalletConnection {
             });
         }
         this._near = near;
-        const authDataKey = appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
+        const DEPRECATED_appKeyPrefix = appKeyPrefix;
+        const finalAppKeyPrefix = appKeyPrefix || near.config.contractName || 'default';
+        const DEPRECATED_authDataKey = DEPRECATED_appKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
+        const authDataKey = finalAppKeyPrefix + LOCAL_STORAGE_KEY_SUFFIX;
+        
+        if (DEPRECATED_authDataKey != authDataKey) {
```
esaminu, line 112: The only case where these aren't equal would be if `appKeyPrefix` is falsy

andy-haynes:
Is the original issue from 2020 that seeing the `undefined` prefix on their LocalStorage keys is potentially discomforting to users? If the entire purpose of the prefix is to differentiate between keys on another wallet on the same domain (😱) then replacing it with `default`  doesn't do anything to alleviate the collision problem.

Mutating live LocalStorage keys carries some risk that we should be very critical about taking on, especially given that this mutation would need to remain indefinitely for future compatibility. @MaximusHaximus and I discussed instead adding a breaking change which _requires_ a prefix, and communicate that `"undefined"` will need to be explicitly passed in order to avoid invalidating user keys. This gives control over the change to the `WalletConnection` consumers and spares NAJ the burden of needing to mutate keys forever.

Overall though I don't think this issue is anything other than cosmetic, given that we cannot retroactively force consumers to use a unique key. I don't believe it's worth taking on the risk of mutating LocalStorage keys indefinitely to make LocalStorage key names more palatable.

esaminu:
Good points all around. I'll open another PR that throws if no `appKeyPrefix` is passed in and we can close this in favor of that solution. The minority of users that were doing 

```typescript
new WalletConnection(near)
```

will now have to do

```typescript
new WalletConnection(near, "undefined")
```

If they want to access the keys they had potentially accumulated

esaminu:
Alternative solution implemented in #1014

PR #953, opened 2022-08-29
fix: docs - docosaurus on gh-pages

maxhr:
- Fixed gh actions to deploy to gh pages. Maybe need to adjust some settings in this repo, but it works on my fork:
  - https://github.com/maxhr/near-api-js/actions/runs/2945068582
  - https://maxhr.github.io/near-api-js/
- Made it work with docosaurus to match docs.near.org look & feel and for advanced UI
- Upgraded TypeDoc version to support the TypeScript upgrade
- Fixed all internal typedoc links.


morgsmccauley (commented):
Overall this looks good but It might be better to have `typdoc/` sit under `packages/` rather than living in the root directory. This would allow Lerna to manage it meaning:
* both `near-api-js` and `typedoc` dependencies can be installed in a single `yarn install`
* NPM scripts can be called from the root directory

which would avoid having to `cd` around to build the docs. From an organization standpoint I think this a bit nicer too.

maxhr:
@morgsmccauley thanks for the review! I disagree because:
- It's not a releasable npm package
- The developer never has run any scripts from it, only the CI does
- Soon it'll also contain documentation about the overlall monorepo, not only the individual lerna packages

So i'll rename the dir to `docs`, it makes more sense.

maxhr:
@morgsmccauley now when I think about it, we should extract the docs into a separate repo. Let's release it as is now, and I'll attend to it next.

maxhr:
@morgsmccauley can you re-review this please?

morgsmccauley:
@maxhr just curious - why should this live in a separate repo? Wouldn't we want builds to trigger off this repo, so it would make sense to keep it here?

morgsmccauley (approved):
LGTM - thanks for the explanation

PR #964, opened 2022-08-30
feat: Implementing `getActiveDelegatedStakeBalance` on `Account` 

hcho112:
## Motivation
NAJ-47, allowing NAJ to request total staked balance on given account id

## Description
Added new function inside `Account` model called `getTotalStakeBalance`.
 

## Checklist
- [x] Added automated tests


frol:
This implementation might be misleading as it does not take into account the delegated stake to the pools that are not in the current validators set. I would be fine with `getActiveDelegatedStakeBalance` or something along those lines. Also, Wallet already mislead people enough with the "stake", so then Explorer has to explain that there is a "native stake" and "delegated stake". Could you consider making sure you indicate that this method will fetch the delegated stake?

hcho112:
Thanks for the feedback @frol , I have replaced the name of the function as recommended. (Also added short description)

packages/near-api-js/src/account.ts
```diff
@@ -629,4 +630,35 @@ export class Account {
             available: availableBalance.toString()
         };
     }
+
+    /**
+     * Returns total delgated stake of the account
```
frol, line 635:
```suggestion
     * Returns the NEAR tokens balance of a given account that is delegated to the staking pools that are part of the validators set in the current epoch.
     
     NOTE: If the tokens are delegated to a staking pool that is currently on pause or does not have enough tokens to participate in validation, they won't be accounted for.
```

packages/near-api-js/src/account.ts
```diff
@@ -629,4 +630,35 @@ export class Account {
...
+
+        const prefix = getValidatorRegExp(this.connection.networkId);
+        const promises = [...pools]
+            .filter((v) => v.indexOf('nfvalidator') === -1 && v.match(prefix))
+            .map(async (validator) => (
+                await this.viewFunction({
+                    contractId: validator,
+                    methodName: 'get_account_total_balance',
+                    args: { account_id: this.accountId },
+                })
```
frol, line 652: To avoid races, it is better to make the view-function calls on the same block. I would fetch the final block and use its hash in the view-function call (NEAR JSON RPC allows to specify the `block_id`: https://docs.near.org/api/rpc/contracts#call-a-contract-function)

hcho112:
@frol Thank you for detailed feedback, your suggested description is so much better!

packages/near-api-js/src/account.ts
```diff
@@ -629,4 +630,40 @@ export class Account {
...
+        const { current_validators, next_validators, current_proposals } = await this.connection.provider.validators(null);
+        const pools:Set<string> = new Set();
+        [...current_validators, ...next_validators, ...current_proposals]
+            .forEach((validator) => pools.add(validator.account_id));
+
+        const prefix = getValidatorRegExp(this.connection.networkId);
+        const block = await this.connection.provider.block({ finality: 'final' });
+        const blockHash = block.header.hash;
+        const promises = [...pools]
+            .filter((v) => v.indexOf('nfvalidator') === -1 && v.match(prefix))
```
esaminu, line 650: I don't think NAJ should be limited to only a few pre defined staking pool factories. IMO we need to remove this filter and change L659 to `Promise.allSettled`. If we remove this filter we can remove hardcoded regexes in [validator_reg_exp.ts](https://github.com/near/near-api-js/pull/964/files#diff-fe4ab4432b326f7d1f048a0c4feeb3b14e09483769a4de68841fb748115c4fc9)

@marcinbodnar would you know why we are excluding `accountId`s that have the string `'nfvalidator'`?

packages/near-api-js/test/account.test.js
```diff
@@ -307,4 +307,10 @@ describe('with deploy contract', () => {
         });
         expect(result).toEqual('hello trex');
     });
+
+    test('get total stake balance', async() => {
```
esaminu, line 311: Is it possible to have coverage for delegating some stake and getting the updated balance?

packages/near-api-js/src/account.ts
```diff
@@ -629,4 +630,40 @@ export class Account {
...
+            .map(async (validator) => (
```
esaminu, line 651: We can just return `this.viewFunction` here no need for async / await

esaminu (changes requested):
Looking good, left some inline comments

packages/near-api-js/src/connection.ts
```diff
@@ -53,4 +73,11 @@ export class Connection {
         const signer = getSigner(config.signer);
         return new Connection(config.networkId, provider, signer, config.jsvmAccountId);
     }
+
+    getListStakingDeposit(accountId: string): Promise<StakingDeposit[]> {
```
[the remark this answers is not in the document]
hcho112: @esaminu actually I should have removed `getListStakingDeposit` as it is no longer being used. After @frol 's feedback on not relying on indexer, decided to not use them. Will remove this on next commit

packages/near-api-js/src/connection.ts
```diff
@@ -29,6 +30,25 @@ function getSigner(config: any): Signer {
     }
 }
 
+/**
+ * @param networkId network id in string
+ * @returns {string} url
+ */
+function getIndexerUrl(networkId: string) :string {
```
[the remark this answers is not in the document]
hcho112: this function won't be used, will remove as well

packages/near-api-js/src/account.ts
hcho112: this part of code was influenced by [staking.js](https://github.com/near/near-wallet/blob/master/packages/frontend/src/redux/actions/staking.js#L409) from wallet repo.
I tried to ran them without the filter previously, and some validators actually error out during `new Contract` step. (So I assumed that some validators need to be excluded therefore I follow the implementation similar to what has been done on wallet repo)

packages/near-api-js/src/account.ts
```diff
@@ -629,4 +630,40 @@ export class Account {
...
+     * @returns {Promise<string>}
+     */
+    async getActiveDelegatedStakeBalance(): Promise<string>  {
+        const { current_validators, next_validators, current_proposals } = await this.connection.provider.validators(null);
+        const pools:Set<string> = new Set();
+        [...current_validators, ...next_validators, ...current_proposals]
+            .forEach((validator) => pools.add(validator.account_id));
+
+        const prefix = getValidatorRegExp(this.connection.networkId);
+        const block = await this.connection.provider.block({ finality: 'final' });
```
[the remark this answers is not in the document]
hcho112: hmm after I have passed blockHash to `provider.validators` it actually errors out and code breaks.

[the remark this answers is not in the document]
esaminu: What we need to do here is use `header.epoch_id` from the `block` request to fetch the validators.

Thanks @nikurt

packages/near-api-js/src/account.ts
esaminu: The `allSettled` should allow us to get the correct total delegated stake even if some validators error out. I am not sure that we should limit NAJ to only consider staked amounts delegated to pools only from some predefined staking pool factories. Could we get further opinions here? @MaximusHaximus @janewang

hcho112: @marcinbodnar @frol can we get your opinion on this?

frol: Instead of hard-coding the account ids, I would suggest you handle the error during fetching the delegated balance. If there is any error except when `error.name = 'HANDLER_ERROR' && error.cause.name = 'CONTRACT_EXECUTION_ERROR'` (and ideally also check the error message has `FunctionCallError(MethodResolveError(MethodNotFound))` error), we can skip this validator, but otherwise we should handle the error (potentially retry), as there could be some network errors or some other glitches, and it is not correct behavior to silently skip staking pools that we just could not get the answer from due to some glitch.

esaminu: We already have retry behavior for `viewFunction` in the [`exponentialBackoff` in `sendJsonRpc` method](https://github.com/near/near-api-js/blob/master/packages/near-api-js/src/providers/json-rpc-provider.ts#L321) which will retry if `error.type === 'TimeoutError'`. So I think we can just consider errors thrown by the `this.viewFunction` call final and conclusive and use `allSettled` here. My main concern with skipping validators that don't match a regex is adding a dependency to NAJ and ignoring arbitrary staking pools and factories that may be deployed or exist in the future.

esaminu (approved):
LGTM 👍

packages/near-api-js/src/account.ts
frol: I don't think it is a good idea to ignore errors especially when it comes to balances. It can backfire if you return that a user has zero NEAR delegated balance when the problem was that there was no internet connection or RPC was down.

hcho112:
Thank you for different insight @frol on this PR. So I have worked to accomodate the both suggestions from @frol and @esaminu by: 
1. Not applying any filtering on the request
2. Updated return response to include more information about validators that were successfully returned staked value and also failed/rejected validators. (So if RPC is return or issue with portion of validator, it will be fully reflected on response)

frol (dismissed):
Thanks for addressing all the concerns. This is the best thing we can do when we only rely on RPC as our data source. In the future, we might want to add a new endpoint to Pagoda Enhanced API which will provide the data based on the indexed data and thus have a complete picture, but so far it is the best thing we can offer.

hcho112:
@esaminu @frol thank you all for the time taken to review/feedback on PR, much appreciated 🙏

morgsmccauley (approved):

Issue #972, opened 2022-09-04
Migrate from Lerna to Changesets & Turborepo


morgsmccauley:
There are a few concerns I have with Lerna so I would like to suggest some alternatives while the cost of change is still low. The main concerns are:

*   Changelogs are generated from git commits - commits are tied closely to the implementation and don’t necessarily convey the overall change. Generating changelogs from PR titles with [lerna-changelog](https://github.com/lerna/lerna-changelog) is a little better but it discourages writing detailed descriptions.
*   Version bumps aren’t transparent - version bumps are determined at publish time, i.e. after merge, rather than when the change/PR has been raised. It isn't immediately obvious what the next version will be which may potentially lead to unwanted bumps.

Taking the above in to account I’m proposing we move to the following tools:

### [Changesets](https://github.com/changesets/changesets) - versioning, changelogs, & publishing

Changesets workflow is centered around “changeset” files, a human authored file containing both the version bump and the changelog entry. These files are committed and included in PR’s along with the code change itself ([example](https://github.com/morgsmccauley/turbo-near-api-js/pull/9)). When planning a release, changeset files are combined to create the full changelog and determine required version bumps. After reviewing the result, a publish can be triggered. This process is automated with a Github action which tracks changesets merged to master and writes the result to a PR, and once this PR is merged, packages will be published ([example](https://github.com/morgsmccauley/turbo-near-api-js/pull/11)).

Compared to Lerna, this has following benefits:

*   Rather than relying on the Git history, version bumps and change logs entries are human authored, creating reader-friendly changelogs.
*   Packages versions are known ahead of time and will only be published after merging the version PR, providing control over our release cadence.

### [Turborepo](https://github.com/vercel/turborepo) - build tooling

By removing Lerna we also lose the ability to run commands across all packages within the monorepo. Similar tools are baked in to Yarn workspaces but their capabilities are limited. Turborepo is a build system aimed to fill this gap by providing:

*   Caching - Executed tasks are cached on disk and replayed if inputs relating to the command have not changed.
*   Pipelines - Task relationships are defined in config to optimize when they are run, rather than being executed [serially like Lerna](https://turborepo.org/docs/guides/migrate-from-lerna#task-scheduling).

Turbo offers many other features but the above alone guarantees faster build times, especially as the repo scales.

### [Pnpm](https://github.com/pnpm/pnpm) - package management

[...]

PR #975, opened 2022-09-05
build: Replace `Yarn` with `PNPM`

morgsmccauley:
## Motivation
#972 

## Description
This PR migrates the repos package manager from `yarn` to `pnpm`. The migration was mostly seamless, but there were a few issues relating to missing/misplaced dependencies, I have commented in-line highlighting these.

Rather than migrating `cookbook` separately, I have moved it under `packages/` so it can be managed by `pnpm` workspaces. This means `cookbook` will be bootstrapped, as well as all other `packages`, when running `pnpm install`. It will also manage the symlinks between shared `packages`.

## Checklist
- [ ] Added automated tests


packages/cookbook/package.json
```diff
@@ -1,16 +1,14 @@
...
-  "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
-  },
   "author": "",
   "license": "ISC",
   "dependencies": {
     "chalk": "^4.1.1",
     "homedir": "^0.6.0",
-    "near-api-js": "file:../../"
+    "near-api-js": "workspace:*"
```
morgsmccauley, line 12: This indicates to `pnpm` that it should create a symlink to the `near-api-js` package within the workspace.

packages/cookbook/package.json
```diff
@@ -1,16 +1,14 @@
 {
-  "name": "cookbook",
+  "name": "@near-js/cookbook",
+  "private": true,
```
morgsmccauley, line 3: Prevents this package from being published.

packages/near-api-js/package.json
```diff
@@ -26,12 +26,12 @@
     "devDependencies": {
         "@types/bn.js": "^5.1.0",
         "@types/http-errors": "^1.6.1",
+        "@types/node": "^18.7.14",
```
morgsmccauley, line 29: This was a [phantom dependency](https://rushjs.io/pages/advanced/phantom_deps/#phantom-dependencies) 🙀 lucky it was only types.

packages/near-api-js/package.json
```diff
@@ -26,12 +26,12 @@
...
         "browserify": "^16.2.3",
         "bundlewatch": "^0.3.1",
         "concurrently": "^7.3.0",
         "danger": "^11.1.1",
         "danger-plugin-yarn": "^1.3.2",
-        "husky": "^7.0.4",
```
morgsmccauley, line 34: Moved to workspace root.

packages/near-api-js/package.json
```diff
@@ -40,28 +40,27 @@
         "semver": "^7.1.1",
         "ts-jest": "^26.5.6",
         "ts-morph": "^15.1.0",
-        "typescript": "^4.7.4",
```
morgsmccauley, line 43: Moved to workspace root as it's used across the repo.

package.json
```diff
@@ -2,16 +2,8 @@
   "name": "@near-api-js/monorepo",
   "private": true,
   "engines": {
-    "node": ">=16.14.0"
-  },
-  "workspaces": {
-    "packages": [
-      "packages/*"
-    ],
-    "nohoist": [
```
morgsmccauley, line 11: This hoisting is no longer a concern - pnpm doesn't hoist all package dependencies to the workspace root like yarn does, all packages have their own node_modules folder with all their defined dependencies (but these are all symlinked for perf reasons).

hcho112 (approved):
LGTM, had brief conversation with @morgsmccauley on benefit of using `pnpm` 

1. It stores packages in single root location. (Packages do have node_modules directory but its sim linked to root pnpm package directory
2. Stop from accidentally using phantom dependencies. (only use packages that are installed) 

Good work!

PR #985, opened 2022-09-13
Secp256k1 support PR rebased

esaminu:
## Motivation
Rebased version of #810 

## Description
Adds support for secp256k1 curve

## Checklist
- [ ] Added automated tests

### DIFF
--- a/.changeset/dry-buttons-exercise.md
+++ b/.changeset/dry-buttons-exercise.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
+Add support for the secp256k1 curve
--- a/packages/near-api-js/package.json
+++ b/packages/near-api-js/package.json
@@ -20,6 +20,7 @@
         "js-sha256": "^0.9.0",
         "mustache": "^4.0.0",
         "node-fetch": "^2.6.1",
+        "secp256k1": "^4.0.3",
         "text-encoding-utf-8": "^1.0.2",
         "tweetnacl": "^1.0.1"
     },
@@ -67,7 +68,7 @@
         "files": [
             {
                 "path": "dist/near-api-js.min.js",
-                "maxSize": "105kB"
+                "maxSize": "250kB"
             }
         ]
     },
--- a/packages/near-api-js/src/transaction.ts
+++ b/packages/near-api-js/src/transaction.ts
@@ -2,7 +2,7 @@ import sha256 from 'js-sha256';
 import BN from 'bn.js';
 
 import { Enum, Assignable } from './utils/enums';
-import { serialize, deserialize } from 'borsh';
+import { serialize, deserialize, BinaryReader, BinaryWriter } from 'borsh';
 import { KeyType, PublicKey } from './utils/key_pair';
 import { Signer } from './signer';
 
@@ -97,6 +97,26 @@ export function deleteAccount(beneficiaryId: string): Action {
 export class Signature extends Assignable {
     keyType: KeyType;
     data: Uint8Array;
+    
+    static borshDeserialize(reader: BinaryReader) {
+        const keyType = reader.readU8();
+        let data: Uint8Array
+        switch (keyType) {
+            case KeyType.ED25519: 
+            data = reader.readFixedArray(64)
+            break
+            case KeyType.SECP256K1: 
+            data = reader.readFixedArray(65)
+            break
+            default: throw new Error(`Unknown key type ${keyType}`);
+        }
+        return new Signature({keyType, data})
+     }
+    
+    borshSerialize(writer: BinaryWriter) {
+        writer.writeU8(this.keyType);
+        writer.writeFixedArray(this.data)
+    }
 }
 
 export class Transaction extends Assignable {
@@ -147,10 +167,7 @@ export class Action extends Enum {
 type Class<T = any> = new (...args: any[]) => T;
 
 export const SCHEMA = new Map<Class, any>([
-    [Signature, {kind: 'struct', fields: [
-        ['keyType', 'u8'],
-        ['data', [64]]
-    ]}],
+    [Signature, {kind: 'function'}],
     [SignedTransaction, {kind: 'struct', fields: [
         ['transaction', Transaction],
         ['signature', Signature]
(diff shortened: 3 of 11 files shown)


packages/near-api-js/src/transaction.ts
```diff
@@ -97,6 +97,26 @@ export function deleteAccount(beneficiaryId: string): Action {
...
+                break;
+            case KeyType.SECP256K1: 
+                data = reader.readFixedArray(65);
+                break;
+            default: throw new Error(`Unknown key type ${keyType}`);
+        }
+        return new Signature({keyType, data});
+    }
+    
+    borshSerialize(writer: BinaryWriter) {
```
esaminu, line 116: Should we not create a new empty `BinaryWriter` here? If there's a use case for appending this to an existing buffer then we should create an empty `BinaryWriter` when none is provided.

packages/near-api-js/src/utils/key_pair.ts
```diff
@@ -52,17 +58,41 @@ export class PublicKey extends Assignable {
...
+                break;
+            case KeyType.SECP256K1: 
+                data = reader.readFixedArray(64);
+                break;
+            default: throw new Error(`Unknown key type ${keyType}`);
+        }
+        return new PublicKey({keyType, data});
+    }
+    
+    borshSerialize(writer: BinaryWriter) {
```
esaminu, line 92: Same comment as above:

> Should we create a new empty `BinaryWriter` here? If there's a use case for appending this to an existing buffer then we should create an empty `BinaryWriter` when none is provided.

packages/near-api-js/src/utils/key_pair.ts
```diff
@@ -78,17 +108,20 @@ export abstract class KeyPair {
...
         switch (curve.toUpperCase()) {
             case 'ED25519': return KeyPairEd25519.fromRandom();
+            case 'SECP256K1': return KeyPairSecp256k1.fromRandom();
             default: throw new Error(`Unknown curve ${curve}`);
         }
     }
 
     static fromString(encodedKey: string): KeyPair {
         const parts = encodedKey.split(':');
+        //TODO: clarify what we must do here
```
morgsmccauley, line 118: We shouldn't assume what the caller wants us to do so I think we should just remove this check and require a prefixed key be passed in, i.e. `secp256k1:` or `ed25519:`.

We already throw an error below for misformed keys, but we can go a step further and introduce a [Template Literal Types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) to enforce it at the type level:
```ts
static fromRandom(curve: `${'secp256k1' | 'ed25519'}:${stirng}`) {
  // ...
}

**Review by morgsmccauley — CHANGES_REQUESTED** [review]

**Comment by morgsmccauley** [comment]
We'll also need a `.changeset` file 😄

**Comment by morgsmccauley** [comment]
Sorry misclick

File: packages/near-api-js/package.json [hunk 4]
```diff
@@ -68,7 +69,7 @@
         "files": [
             {
                 "path": "dist/near-api-js.min.js",
-                "maxSize": "105kB"
+                "maxSize": "250kB"
```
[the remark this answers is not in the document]
esaminu: We went from 68.23KB in current master to 174.92KB which is a big jump due to the secp256k1 dependency https://bundlephobia.com/package/secp256k1@4.0.3 

@MaximusHaximus thoughts on how if at all this affects how we release this?

packages/near-api-js/src/transaction.ts
```diff
@@ -147,10 +167,7 @@ export class Action extends Enum {
 type Class<T = any> = new (...args: any[]) => T;
 
 export const SCHEMA = new Map<Class, any>([
-    [Signature, {kind: 'struct', fields: [
-        ['keyType', 'u8'],
-        ['data', [64]]
-    ]}],
+    [Signature, {kind: 'function'}],
```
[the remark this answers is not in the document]
esaminu: This answers my comments about use case for appending to an existing `BinaryWriter`. When kind is `'function'` then [borsh calls `borshSerialize` on the instance](https://github.com/near/borsh-js/blob/6db74764cedb88097bd6d5636407f0b3481fba4c/borsh-ts/index.ts#L315) instead of serializing each field.

Resolving all 3 👍

packages/near-api-js/src/utils/key_pair.ts
esaminu: Answered here: https://github.com/near/near-api-js/pull/985/files#r981167051

packages/near-api-js/src/transaction.ts
esaminu: Answered here: https://github.com/near/near-api-js/pull/985/files#r981167051

yarn.lock
```diff
@@ -2845,7 +2845,7 @@ electron-to-chromium@^1.4.202:
   resolved "https://registry.yarnpkg.com/electron-to-chromium/-/electron-to-chromium-1.4.202.tgz#0c2ed733f42b02ec49a955c5badfcc65888c390b"
```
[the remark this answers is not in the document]
esaminu: Rebased 👍

packages/near-api-js/src/utils/key_pair.ts
esaminu: Agreed 👍

morgsmccauley: We can delete this comment now :)

.changeset/dry-buttons-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
+Add support for the secp256k1 curve
```
morgsmccauley, line 5: It would be good to add more detail here, can we please expand on this? Maybe something along the lines of: "Add support for generating secp256k1 key pairs for signing and verification of messages". Not super familiar with the changes so maybe there is more.

frol:
@gtsonevv Oh, I missed your message, sorry for the delay. I believe you are in a better position to judge if it is ready to merge. To be honest, I have never used secp256k1 keys on NEAR.

PR #995, opened 2022-09-21
Action to trigger external docs-generator

maxhr:
Upon `Release` workflow completed, an external webhook is called to trigger a workflow on `@near/docs-generator` to make the docs for this release.


.github/workflows/docs-generator-trigger.yml
```diff
@@ -0,0 +1,30 @@
...
+    branches: [master]
+
+concurrency:
+  group: "docs"
+  cancel-in-progress: true
+
+jobs:
+  trigger-generator:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
morgsmccauley, line 16:
```suggestion
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
    environment: DOCS
```
Can we please add an [environment](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idenvironment) here. This will allow us to add secret protection so the secret is only accessible under certain conditions, i.e. when run via `master` and not from Pull Requests.

morgsmccauley (approved):

PR #998, opened 2022-09-27
Transaction builder class rebased

esaminu:
## Motivation
Updated and rebased #891 

## Description
Adds builder pattern transaction creation:

```typescript
bob.createTransaction("alice.near").transfer("1000000000000000000").signAndSend();
```

## Checklist
- [x] Added automated tests


packages/near-api-js/src/account.ts
```diff
@@ -631,4 +632,101 @@ export class Account {
...
+    get accountCreated(): boolean {
+        return this.accountToBeCreated;
+    }
+
+    get transferAmount(): BN {
+        return this._transferAmount ?? ZERO_NEAR;
+    }
+
+    signAndSend(): Promise<FinalExecutionOutcome> {
+        //@ts-expect-error is protected
```
willemneal, line 724: Is there a good reason why this is still protected?

packages/near-api-js/test/account.test.js
```diff
@@ -158,6 +158,27 @@ describe('with deploy contract', () => {
         expect(await workingAccount.viewFunction(contractId, 'getValue', {})).toEqual(setCallValue);
     });
 
+    test('make function calls via txn', async () => {
```
morgsmccauley, line 161:
```suggestion
    test('make function calls via TransactionBuilder', async () => {

File: packages/near-api-js/test/account.test.js [hunk 3]
```diff
@@ -158,6 +158,27 @@ describe('with deploy contract', () => {
...
+            contractId,
+            'hello', // this is the function defined in hello.wasm file that we are calling
+            { name: 'trex' }
+        );
+        expect(result).toEqual('hello trex');
+
+        const setCallValue = testUtils.generateUniqueString('setCallPrefix');
+        const result2 = await workingAccount
+            .createTransaction(contractId)
+            .functionCall('setValue', { value: setCallValue })
```
morgsmccauley, line 172: Should we add more tests here? We're only really testing `functionCall` and `signAndSend` but I feel there is much more logic to cover?

morgsmccauley (changes requested):
Looks good, just a few comments :)

packages/near-api-js/src/account.ts
```diff
@@ -710,4 +711,100 @@ export class Account {
...
+/**
+ * Transaction Builder class. Initialized to an account that will sign the final transaction
+ */
+export class Transaction {
+    readonly receiverId: string;
+    readonly actions: Action[] = [];
+    private accountToBeCreated = false;
+    private _transferAmount?: BN;
+
+    constructor(private signer: Account, receiver: Account | string) {
```
andy-haynes, line 734: Given that we have an open issue around using named parameters, I'm thinking we should start writing new code using all named parameters.

packages/near-api-js/src/account.ts
```diff
@@ -710,4 +711,100 @@ export class Account {
...
+    async hasDeployedContract(): Promise<boolean> {
+        return (await this.state()).code_hash !== EMPTY_CONTRACT_HASH;
+    }
 }
+
+
+/**
+ * Transaction Builder class. Initialized to an account that will sign the final transaction
+ */
+export class Transaction {
```
andy-haynes, line 728: Is there a compelling reason to not break this out into its own module instead of exporting it from `account.ts`? This file is already absurdly long and if other code is going to import this then it makes more sense for it to have its own module.

packages/near-api-js/src/account.ts
```diff
@@ -710,4 +711,100 @@ export class Account {
...
+        this.actions.push(transfer(this._transferAmount));
+        return this;
+    }
+
+    push(action: Action): this {
+        this.actions.push(action);
+        return this;
+    }
+
+    get accountCreated(): boolean {
```
andy-haynes, line 794: What is the use case for this (and `transferAmount`)? Without more context it seems like it would be better for the calling code to manage these values since they don't seem related to building a set of transactions.

packages/near-api-js/src/account.ts
```diff
@@ -710,4 +711,100 @@ export class Account {
...
+    }
 }
+
+
+/**
+ * Transaction Builder class. Initialized to an account that will sign the final transaction
+ */
+export class Transaction {
+    readonly receiverId: string;
+    readonly actions: Action[] = [];
```
andy-haynes, line 730: Is there a use case for clearing the set of actions? Could it be done automatically somewhere?

andy-haynes (changes requested):

packages/near-api-js/src/account.ts
esaminu: Changed the base so it's no longer protected 🙂

packages/near-api-js/src/account.ts
willemneal: My original intent was that there was a interdependency between the two classes. An Account can create a txn builder, which in turn requires a signer (an Account).  But perhaps it really just requires a signer and the account can pass the signer.

```ts
this.signer.signAndSendTransaction({ receiverId: this.receiverId, actions: this.actions })
this.signer.signTransaction(this.receiverId, this.actions);
```

packages/near-api-js/src/account.ts
willemneal: This class originally lived with workspaces-js and it needed to track whether an account was created, because the testnet workspace would delete these accounts refunding the root account after a test. It should instead extend this class.

esaminu: Agreed I don't think this class needs to maintain this metadata about the transaction. Also seems to me that these can be derived from the action list. Removing

packages/near-api-js/src/account.ts
esaminu: Yeah I think constructing these with signer makes a lot of sense and breaks the require cycle. Moving this class into it's own module 👍

packages/near-api-js/src/account.ts
esaminu: Moving this to inside the constructor to be more explicit 👍

packages/near-api-js/src/account.ts
esaminu: I've pulled the access key caching and transaction sending functionality into a new `TransactionSender` class that both `Account` and `TransactionBuilder` extend

packages/near-api-js/src/account.ts
esaminu: I think this is more for methods rather than constructors e.g. `Account` constructor args are `Connection` and `accountId`. I went for `Connection`, `senderId` and `receiverId` for `TransactionBuilder` since I felt `Connection` doesn't necessarily group well with the other args stylistically but maybe `Connection` and a `TransactionOptions` object with `senderId` and `receiverId`?

packages/near-api-js/src/transaction_builder.ts
```diff
@@ -0,0 +1,84 @@
...
+        this.actions.push(deleteKey(PublicKey.from(publicKey)));
+        return this;
+    }
+
+    deployContract(code: Uint8Array | Buffer): this {
+        this.actions.push(deployContract(code));
+        return this;
+    }
+
+    functionCall({ methodName, args = {}, gas = DEFAULT_FUNCTION_CALL_GAS, attachedDeposit, stringify }: Omit<FunctionCallOptions, 'contractId' | 'jsContract'>): this {
```
esaminu, line 55: I updated this to have more overlap with `Account.functionCall`

packages/near-api-js/src/constants.ts
```diff
@@ -1,9 +1,17 @@
...
+* For discussion see https://github.com/nearprotocol/NEPs/issues/67
+*/
+export const DEFAULT_FUNCTION_CALL_GAS = new BN('30000000000000');
+
+/**
+ * Hash if no contract is present for an account.
+ */
+export const EMPTY_CONTRACT_HASH = '11111111111111111111111111111111';
+
+export const ZERO_NEAR = new BN('0');
```
esaminu, line 17: @willemneal do we need this? `Account.functionCall` does not have a fallback for `attachedDeposit` can you describe the failure mode you're trying to avoid here?

willemneal: Do you mean the ZERO_NEAR? I'm a little confused with the context.  Why did the default for function call have zero attached deposit?  If that's the question it's because that is the common case. Not really for any failure, just convenience.

esaminu: Yes I'm wondering why we aren't doing the same for function call actions sent from `Account.functionCall`. Maybe the fallback is handled at the protocol level. Is there a difference between a function call action with no deposit specified and one with `0`? `Account.functionCall` actions don't specify a deposit when none is passed in currently.

packages/near-api-js/test/account.test.js
esaminu: Added coverage for the same cases as `Account` class 👍

packages/near-api-js/src/account.ts
andy-haynes: >I think this is more for methods rather than constructors e.g. Account constructor args are Connection and 
accountId.

Ideally we'd change this for `Account` and `Connection` too 😄 

The goal behind named parameters is extendability and maintainability. They make it trivial to extend and refactor arguments and eliminate the [connascence of position](https://en.wikipedia.org/wiki/Connascence#Connascence_of_position_(CoP)) that makes methods unnecessarily brittle. JavaScript's handling of arguments can make positional arguments a liability, so having a single object argument with named parameters is not just a cosmetic concern.

The constructor benefits from this as much as any other method, and in fact it is a very common pattern in JavaScript development to have destructured object parameters in the constructor. Moreover having a single type representing dependencies for instantiating a class is convenient, consistent, and self-documenting.

> I went for Connection, senderId and receiverId for TransactionBuilder since I felt Connection doesn't necessarily group well with the other args stylistically but maybe Connection and a TransactionOptions object with senderId and receiverId?

There is no implication of grouping or association by including two arguments as keys on a single parameter object - again the benefit to naming your parameters is to make meaningful distinctions between them and enforce clarity in calling code. If you feel there is value to grouping parameters then it absolutely makes sense to do so within the same object, e.g.

```ts
type TransactionOptions = {
  receiverId: string;
  senderId: string;
};

type TransactionBuilderConfig = {
  connection: Connection;
  options: TransactionOptions;
};
...
constructor({ connection, options: { receiverId, senderId } }: TransactionBuilderConfig)
```

Personally I think it's a bit too verbose if the set of arguments is small and the names are clear, which in this case they are. I find grouping parameters is more useful when the set of arguments are more complex and where the arguments meanings benefit from additional context via partitioning.

packages/near-api-js/src/account.ts
```diff
@@ -687,4 +485,12 @@ export class Account {
             total: summary.total.toString(),
         };
     }
+
+    createTransaction(receiver: Account | string): TransactionBuilder {
+        return new TransactionBuilder(this.connection, this.accountId, typeof receiver === 'string' ? receiver : receiver.accountId);
```
andy-haynes, line 490: Rather than refactor the `Account` class to share some functionality, why not pass the `signTransaction` and `signAndSendTransaction` methods to the new class? `TransactionBuilder` doesn't need to know how to send transactions, as evidenced by the name 🙂 

Passing the methods would eliminate the need to break up the `Account` class, which seems to have been done solely for the purpose of factoring out two methods (and their dependencies). What would you think about something like this instead?

```ts
createTransaction(receiverId: string): TransactionBuilder {
  return new TransactionBuilder({
    signTransaction: (actions: Action[]) => this.signTransaction(this.accountId, actions),
    signAndSendTransaction: (actions: Action[]) => this.signAndSendTransactions({ receiverId: this.accountId, actions }),
  });
}
...
class TransactionBuilder {
  signTransaction: (actions: Action[]) => Promise<[Uint8Array, SignedTransaction]>;
  signAndSendTransaction: (actions: Action[]) => Promise<FinalExecutionOutcome>;

  constructor({ signTransaction, signAndSendTransaction }) {
    this.signTransaction = signTransaction;
    this.signAndSendTransaction = signAndSendTransaction;
  }
  ...
}
```

Now `TransactionBuilder` doesn't need a `connection` or `accountId` - it just builds transactions and exposes `sign` and `signAndSend` methods that look like
```ts
sign(): Promise<[Uint8Array, SignedTransaction]> {
  return this.signTransaction(this.actions);
}
```

With this the refactor is not required and we can keep `Account` logic consolidated.

packages/near-api-js/src/account.ts
```diff
@@ -687,4 +485,12 @@ export class Account {
...
+    }
+  
+    async hasDeployedContract(): Promise<boolean> {
```
andy-haynes, line 493: Is this relevant to the PR?

packages/near-api-js/src/constants.ts
andy-haynes: Are these changes relevant to `TransactionBuilder` at all? If not could they go in a separate PR please?

.changeset/rich-rice-laugh.md
```diff
@@ -0,0 +1,8 @@
+---
+"near-api-js": minor
+---
+
+Add builder pattern transaction creation feature as an alternative way to create and send transactions:
+```typescript
+bob.createTransaction("alice.near").transfer("1000000000000000000").signAndSend();
```
andy-haynes, line 7: Could we make it more clear that `bob` is an instance of `Account` here?

Maybe something like
```ts
new Account(connection, 'bob.near').createTransaction("alice.near").transfer("1000000000000000000").signAndSend();
```

packages/near-api-js/src/transaction_builder.ts
```diff
@@ -0,0 +1,84 @@
...
+        this.actions.push(stake(amount, PublicKey.from(publicKey)));
+        return this;
+    }
+
+    transfer(amount: BN): this {
+        this.actions.push(transfer(amount));
+        return this;
+    }
+
+    push(action: Action): this {
```
andy-haynes, line 72: Can we rename this to `addTransaction` instead please?

packages/near-api-js/src/transaction_builder.ts
```diff
@@ -0,0 +1,84 @@
...
+        super(connection, senderId);
+        this.receiverId = receiverId;
+        this.connection = connection;
+        this.senderId = senderId;
+        this.actions = [];
+        this.accessKeyByPublicKeyCache = {};
+    }
+
+    addKey(publicKey: string | PublicKey, accessKey: AccessKey = fullAccessKey()): this {
+        this.actions.push(addKey(PublicKey.from(publicKey), accessKey));
```
andy-haynes, line 31: Any reason not to use the exposed method rather than mutating the state directly? Using `this.push` here instead ensures there's only one way that state gets mutated, and makes the code less brittle.

andy-haynes (changes requested):
The refactor here seems a little heavy for what it accomplishes, I'd prefer if we explored implementing a "dumb" `TransactionBuilder` instead of one that overlaps functionality in the `Account` class.

packages/near-api-js/src/constants.ts
esaminu: Yeah `TransactionBuilder` function call actions are defaulting to `ZERO_NEAR` when no deposit is passed in

packages/near-api-js/src/account.ts
esaminu: Makes sense, grouping into a `TransactionBuilderConfig` object with `connection`, `receiverId` and `senderId`

packages/near-api-js/src/transaction_builder.ts
esaminu: renaming to `addAction`

.changeset/rich-rice-laugh.md
esaminu: Updated

packages/near-api-js/src/constants.ts
esaminu: I'm removing these features incl the fallback and `hasDeployedContract` for now. We can discuss if they're necessary in a separate PR / issue

packages/near-api-js/src/account.ts
andy-haynes: I see what you mean about its use outside of `Account`, thanks for the clarification.

> Separating out how `Account` sends transactions from `Account` specific functionality has been overdue imo.

I think you're 100% right on this, however I would argue that factoring this behavior out into a parent class doesn't go far enough in separating these concerns. Inheritance may be more convenient in the short term for this but we have an opportunity to use composition to accomplish the same thing in a more robust way. By passing instances of `TransactionSender` to `Account` and `TransactionBuilder` rather than putting this in a parent class, we can avoid the coupling that inheritance guarantees.

Another benefit is that mocking becomes trivial - different `TransactionSender` mocks (e.g. one for throwing an error, another returning unexpected data, etc.) could be shared across unit tests, eliminating the need for transaction signing integration tests on consumers of `TransactionBuilder`.

esaminu: Great points. I rebased this PR on #1018 and #1019 👍

PR #1002, opened 2022-09-30
Naj 44 make sign and send transaction functionality public

esaminu:
## Motivation
NAJ-44

## Description
Make `Account.signAndSendTransaction` public so transactions can be sent directly from Account instances

## Checklist
- [x] Added automated tests

### DIFF
--- a/.changeset/beige-suits-march.md
+++ b/.changeset/beige-suits-march.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": major
+---
+
+Make `Account.signAndSendTransaction` `public` so transactions can be sent directly from `Account` instances
--- a/.changeset/lazy-candles-hide.md
+++ b/.changeset/lazy-candles-hide.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
+Make Account.signAndSendTransaction public so transactions can be sent directly from Account instances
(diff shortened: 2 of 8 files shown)


andy-haynes:
NAJ-44 mentions
> We cannot just remove the `protected` accessor because that would be a breaking change (every implementation of `Account` would have to update).

Is there more context to this on the original issue (I can't find the link)? I'm not seeing how this is a breaking change, assuming that "implementation of `Account`" refers to creating a subclass of it.

esaminu:
@andy-haynes Yes it appears that's what the ticket is referring to. i.e. if you extend account you'd have to make this method `public` too as opposed to `protected`. There's a discussion this here as well: https://github.com/near/near-api-js/pull/557#discussion_r624559026

This is not an issue imo but we have to bump a major here (updating changeset)

andy-haynes:
> @andy-haynes Yes it appears that's what the ticket is referring to. i.e. if you extend account you'd have to make this method `public` too as opposed to `protected`. There's a discussion this here as well: [#557 (comment)](https://github.com/near/near-api-js/pull/557#discussion_r624559026)
> 
> This is not an issue imo but we have to bump a major here (updating changeset)

Sounds good, thanks!

andy-haynes (approved):

hcho112 (approved):
LGTM

Issue #980, opened 2022-09-09
Error types have degraded


lostpebble:
**Describe the bug**
When we used to call the `account.state()` method and the account does not actually exist on the network, we would get an error thrown which included:

```ts
 { "type": "AccountDoesNotExist" }
```

But no such type gets returned anymore- all we get now is:

```ts
{ "type": "HANDLER_ERROR" }
```

This was a big surprise as it broke our interface for account funding (we had to check if the account exists to show an interface which requests funding to activate it).

Please can you reinstate error types which help to decipher what exactly went wrong. "HANDLER_ERROR" doesn't really tell us anything as to what has gone wrong along the way, it could be one of any number of things.

We've had to resort to doing string matching on the error message now, which is obviously not ideal as that could change at any point. A solid error type ID is always the best.

PR #1003, opened 2022-09-30
fix: error types

marcinbodnar:
## Motivation
This is a partial restoration of changes from #910 . At this moment near-api-js is not returning proper error types for errors: `AccountDoesNotExist`, `AccessKeyDoesNotExist`, `CodeDoesNotExist`, `InvalidNonce`. Issue: #980 .

## Description
1. Restoring `getErrorTypeFromErrorMessage` function, with a default value change. The previous implementation was always returning `UntypedError` if the error message doesn't match pointed strings, which means it could possibly change the type even if it already exists, which was not ideal - this function should return the proper type if it matches, and don't change the type if not matches. If the request doesn't have a type at all, we have a proper condition to change the type if it doesn't exist in the `TypedError` class.
2. Restoring usage of `getErrorTypeFromErrorMessage` in `JsonRpcProvider.query` method, but with the current message.
3. Restoring usage of `getErrorTypeFromErrorMessage ` in `JsonRpcProvider.sendJsonRpc` method.
4. Restoring the test and also removing `UntypedError` case from it.

## Checklist
- [x] Added automated tests

### DIFF
--- a/.changeset/wild-teachers-marry.md
+++ b/.changeset/wild-teachers-marry.md
@@ -1,0 +1,7 @@
+---
+"near-api-js": patch
+---
+
+Fix error types. WIth this changes both `JsonRpcProvider.query` and `JsonRpcProvider.sendJsonRpc` methods will return proper error type for these errors: `AccountDoesNotExist`, `AccessKeyDoesNotExist`, `CodeDoesNotExist`, `InvalidNonce`.
+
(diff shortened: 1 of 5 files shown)


hcho112 (dismissed):
LGTM, wonder if we should add Changesets here?

marcinbodnar:
> LGTM, wonder if we should add Changesets here?

Not sure, what do you think @esaminu ?

esaminu:
@marcinbodnar imo this would be a patch version bump since it's a backward compatible fix

hcho112 (approved):
LGTM!

PR #1006, opened 2022-10-03
Revert "Revert "Naj 44 make sign and send transaction functionality public""

morgsmccauley:
Reverts near/near-api-js#1005

https://github.com/near/near-api-js/pull/1002 was reverted so we could temporarily avoid a major bump of `near-api-js`


andy-haynes (approved):

leapsamvel:
Hi,

Is there a timeline for publishing this into NPM?

PR #1015, opened 2022-10-11
test: add cookbook tests and CI

### LINKED ISSUE #622: add unit tests for cookbook
As we add examples to the "cookbook" we should have unit tests in place to ensure they stay up to date.

Add unit testing to all cookbook examples.

### LINKED ISSUE #657: Add CI checks that cookbook and reference examples work
Currently we are not tracking whether changes made to this repo will break the demonstration snippets/examples in the `examples` directory.

We'll likely have to consider using contract helper for CI in order to create full-access keys for some of the tests that require sending native tokens, like the `wrap-near.js` example.

andy-haynes:
## Motivation
This PR updates the `cookbook` package to standardize examples and add integration tests.

Fixes #622 
Fixes #657 

## Description
- fixes and/or removes redundant code in several broken recipes
- uses existing `eslintrc` for standardized conventions across packages
- recipes have standardized sets of parameters but are still invokable via `node` with default parameters
- integration tests for recipes
- adds Kurtosis for `localnet` integration testing in CI

As a consequence of these new changes, running the `cookbook` tests requires a localnet instance (expected at `localhost:8332`). Instructions have been added on (relatively simple) setup but it still means more dependencies to be able to run `cookbook` integration tests. These tests will run on PRs (including forks) however, so it shouldn't be too annoying for most.

## Checklist
- [x] Added automated tests


packages/cookbook/src/accounts/access-keys/create-full-access-key.js
```diff
@@ -0,0 +1,41 @@
+const { KeyPair, keyStores, connect } = require('near-api-js');
+const path = require('path');
+const os = require('os');
+
+async function createFullAccessKey({ accountId, keyStore, networkId, nodeUrl }) {
```
andy-haynes, line 5: I went with this pattern to make it easier for new developers to see the dependencies for each recipe in their simplest form. In particular extracting `keyStore` allows the local script invocation to use a file system keystore while the tests can use the in-memory implementation. I've tried to remove unnecessary dependencies where possible (e.g. removing `keyStore` for view function calls).

packages/cookbook/src/accounts/access-keys/create-full-access-key.js
```diff
@@ -0,0 +1,41 @@
...
+    await account.addKey(publicKey);
+
+    return publicKey;
+}
+
+module.exports = {
+    createFullAccessKey,
+};
+
+if (require.main === module) {
```
andy-haynes, line 22: This ensures that the original logic w/ default parameters still works as expected when invoked via `node`. Most of the parameters are hardcoded but where the original implementation accepted command line arguments I've left the behavior the same.

Ideally I think each one would have a common CLI interface which would fall back on default arguments, allowing developers new to the ecosystem an opportunity to more seamlessly explore the API by tweaking parameters. However with all the changes in this PR it seems better left for a follow-up effort.

packages/cookbook/src/accounts/create-account.js
```diff
@@ -0,0 +1,62 @@
+const { connect, KeyPair, keyStores, utils } = require('near-api-js');
```
andy-haynes, line 1: This was originally two separate scripts with only the differences being the network specified. I consolidated them into this new file, which takes the network ID (`testnet` or `mainnet`) as a parameter but otherwise works the same.

However I was not able to write a successful integration for this against `localnet` as it appears to rely on creating sub-accounts (i.e. `Account.createAccount(...)` does work in `localnet`) rather than calling the `new_account` method on the root account.

packages/cookbook/README.md
```diff
@@ -45,3 +45,10 @@ Example:
 ```bash
 node utils/get-state
 ```
+
+
+## Testing
```
andy-haynes, line 50: It seems a little inconvenient that `pnpm test` doesn't do anything to set up `localnet` but this requires a Docker image checkout and local installation of `kurtosis-cli` so it seems a little heavy for a `package.json` script. I don't think it will be a huge impediment for developers given that:
- most devs won't be running tests for this package (unless they're running them at the root)
- `localnet` is configured on the PR actions so the test results will be available
- the setup instructions here are simple with minimal dependencies

andy-haynes:
@idea404 @maxhr I was having trouble finding reviewers for this one - if you have any thoughts on these please let me know. I've tried to make the examples consistent and more accessible as well as adding a new approach to CI that we could use across other packages as well.

andy-haynes:
@idea404 I rebased the repo, you can see if that makes any difference. It seems to work OK on `master` so I'm not sure what other differences there might be.

I won't be able to look at the failing test script for a while but I'm hoping it's just a matter of updating a reference to a Kurtosis package.

vikinatora:
Hey @andy-haynes, I'm currently conducting a clean-up on the PRs of NAJ. Is this PR still of interest to NAJ? If so our team can take over and try to finalize it so it gets merged. Otherwise, we should close it.

andy-haynes:
> Hey @andy-haynes, I'm currently conducting a clean-up on the PRs of NAJ. Is this PR still of interest to NAJ? If so our team can take over and try to finalize it so it gets merged. Otherwise, we should close it.

The tests would still be useful to prevent regressions in public examples, so it depends how you define "of interest to NAJ". At this point I don't know what it would take to get it merged, so I can't speak to the value proposition as a function of effort.

vikinatora:
@andy-haynes Sure, we will keep in mind this PR and prioritize it in the future.

Issue #795, opened 2022-02-13
Strict mode in tsconfig


frolovdev:
**Is your feature request related to a problem? Please describe.**
Let's enable a strict mode in typescript

**Describe the solution you'd like**
Enable strict mode by fixing current errors

https://www.typescriptlang.org/tsconfig#strict

**Describe alternatives you've considered**
N/A

**Additional context**

An error log is not so big, so it can be done in one pr without any gradual migration. I can help with a pull request.

```
src/account.ts(302,13): error TS2322: Type 'null' is not assignable to type '{ publicKey: PublicKey; accessKey: AccessKeyView; }'.
src/account.ts(330,17): error TS2322: Type 'null' is not assignable to type '{ publicKey: PublicKey; accessKey: AccessKeyView; }'.
src/account.ts(427,88): error TS2345: Argument of type 'BN | undefined' is not assignable to parameter of type 'BN'.
  Type 'undefined' is not assignable to type 'BN'.
src/account.ts(436,59): error TS2345: Argument of type 'BN | undefined' is not assignable to parameter of type 'BN'.
  Type 'undefined' is not assignable to type 'BN'.
src/account_multisig.ts(15,42): error TS2769: No overload matches this call.
  Overload 1 of 2, '(number: string | number | number[] | Uint8Array | Buffer | BN, base?: number | "hex" | undefined, endian?: "le" | "be" | undefined): BN', gave the following error.
    Argument of type 'string | null' is not assignable to parameter of type 'string | number | number[] | Uint8Array | Buffer | BN'.
      Type 'null' is not assignable to type 'string | number | number[] | Uint8Array | Buffer | BN'.
  Overload 2 of 2, '(number: string | number | number[] | Uint8Array | Buffer | BN, endian?: "le" | "be" | undefined): BN', gave the following error.
    Argument of type 'string | null' is not assignable to parameter of type 'string | number | number[] | Uint8Array | Buffer | BN'.
      Type 'null' is not assignable to type 'string | number | number[] | Uint8Array | Buffer | BN'.
src/account_multisig.ts(225,105): error TS2345: Argument of type 'null' is not assignable to parameter of type 'BN | undefined'.
src/account_multisig.ts(226,95): error TS2345: Argument of type 'null' is not assignable to parameter of type 'BN | undefined'.
src/account_multisig.ts(251,83): error TS2345: Argument of type 'null' is not assignable to parameter of type 'AccessKey'.
[...]

PR #1016, opened 2022-10-13
chore(ts): enforce noImplicitAny

andy-haynes:
## Motivation
This PR enables `noImplicitAny` to the tsconfig. It addresses part of #795 

## Description
- added missing type annotations
- added `@types` packages for dependencies without type declarations
- suppresses error around type narrowing when `global.fetch` is assigned to the local `fetch`

## Checklist
- [ ] Added automated tests

### DIFF
--- a/packages/near-api-js/src/account_multisig.ts
+++ b/packages/near-api-js/src/account_multisig.ts
@@ -1,12 +1,22 @@
 'use strict';
 
 import BN from 'bn.js';
+
 import { Account, SignAndSendTransactionOptions } from './account';
 import { Connection } from './connection';
 import { parseNearAmount } from './utils/format';
 import { PublicKey } from './utils/key_pair';
-import { Action, addKey, deleteKey, deployContract, fullAccessKey, functionCall, functionCallAccessKey } from './transaction';
-import { FinalExecutionOutcome, TypedError } from './providers';
+import {
+    AccessKey,
+    Action,
+    addKey,
+    deleteKey,
+    deployContract,
+    fullAccessKey,
+    functionCall,
+    functionCallAccessKey,
+} from './transaction';
+import {FinalExecutionOutcome, TypedError} from './providers';
 import { fetchJson } from './utils/web';
 import { FunctionCallPermissionView } from './providers/provider';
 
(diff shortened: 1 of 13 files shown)


packages/near-api-js/src/account_multisig.ts
```diff
@@ -438,13 +443,20 @@ export class Account2FA extends AccountMultisig {
     }
 
     async get2faMethod() {
-        let { data } = await this.getRecoveryMethods();
-        if (data && data.length) {
-            data = data.find((m) => m.kind.indexOf('2fa-') === 0);
+        const { data } = await this.getRecoveryMethods();
```
andy-haynes, line 446: Types here were pretty loose so I cleaned this up a bit.

packages/near-api-js/src/account_multisig.ts
```diff
@@ -88,19 +98,14 @@ export class AccountMultisig extends Account {
             throw e;
         }
 
-        // TODO: Are following even needed? Seems like it throws on error already
-        if (!result.status) {
-            throw new Error('Request failed');
-        }
-        const status: any = { ...result.status };
-        if (!status.SuccessValue || typeof status.SuccessValue !== 'string') {
+        if (!result.status || typeof result.status !== 'object' || !result.status.SuccessValue || result.status.SuccessValue !== 'string') {
```
andy-haynes, line 101: TS didn't like the spread here so I went ahead and condensed this a bit (bonus `TODO` removal 😎)

packages/near-api-js/src/providers/provider.ts
```diff
@@ -315,6 +315,7 @@ export interface AccessKeyWithPublicKey {
 export interface QueryResponseKind {
     block_height: BlockHeight;
     block_hash: BlockHash;
+    error: Error;
```
andy-haynes, line 318: Not totally sure about this one but it looks like the fields we reference directly at least overlap with the shape on the RPC error response.

packages/near-api-js/src/providers/json-rpc-provider.ts
```diff
@@ -149,7 +149,7 @@ export class JsonRpcProvider extends Provider {
         if (result && result.error) {
             throw new TypedError(
                 `Querying failed: ${result.error}.\n${JSON.stringify(result, null, 2)}`,
-                getErrorTypeFromErrorMessage(result.error, result.error.name)
+                getErrorTypeFromErrorMessage(result.error.toString(), result.error.name)
```
andy-haynes, line 152: I think this was happening implicitly when the `/.../.test()` method is invoked with an `Error` object. Since the field being tested is ostensibly the error message, this should be a safe substitute.

vikinatora:
Hey @andy-haynes, I'm currently conducting a clean-up on the PRs of NAJ. Is this PR still of interest to NAJ? If so our team can take over and try to finalize it so it gets merged. Otherwise, we should close it.

PR #1019, opened 2022-10-18
Refactor: remove duplicate types and move out multisig

esaminu:
## Motivation
Augments the refactor in #1018

## Description
Further decouples multisig logic for more explicit multisig vs regular transactions and removes duplicate types and `TransactionSender` instances

## Checklist
- [x] Added automated tests

### DIFF
--- a/packages/near-api-js/src/account.ts
+++ b/packages/near-api-js/src/account.ts
@@ -31,7 +31,7 @@ import { PublicKey } from './utils/key_pair';
 import { PositionalArgsError } from './utils/errors';
 import { printLogs } from './utils/logging';
 import { DEFAULT_FUNCTION_CALL_GAS } from './constants';
-import { TransactionSender } from './transaction_sender';
+import { SignAndSendTransactionOptions, TransactionSender } from './transaction_sender';
 
 export interface AccountBalance {
     total: string;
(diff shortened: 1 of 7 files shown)


packages/near-api-js/src/account.ts
```diff
@@ -46,25 +46,6 @@ export interface AccountAuthorizedApp {
...
-    /**
-     * Metadata to send the NEAR Wallet if using it to sign transactions.
-     * @see {@link RequestSignTransactionsOptions}
-     */
-    walletMeta?: string;
-    /**
-     * Callback url to send the NEAR Wallet if using it to sign transactions.
-     * @see {@link RequestSignTransactionsOptions}
-     */
-    walletCallbackUrl?: string;
```
andy-haynes, line 64: It turns out these two are used in the `Account.signAndSendTransaction` inheritance chain despite not being referenced in the base implementation directly on `Account`. We're better off keeping this interface here and narrowing the interface used by `TransactionSender` to omit these two arguments.

I'll make those changes in the base branch and rename the interfaces in `TransactionSender` to have less terrible names for what they are now.

esaminu: hmm AFAICT it's being used in wallet-account.ts. This likely warrants a new interface in that scope:

```typescript
interface WalletTransactionOptions extends Omit<SignAndSendTransactionOptions, 'signerId'> {
    /**
     * Metadata to send the NEAR Wallet if using it to sign transactions.
     * @see {@link RequestSignTransactionsOptions}
     */
    walletMeta?: string;
    /**
     * Callback url to send the NEAR Wallet if using it to sign transactions.
     * @see {@link RequestSignTransactionsOptions}
     */
    walletCallbackUrl?: string;
}
```

andy-haynes: I'd prefer to leave the existing interface as is and decouple these interfaces here. The fact that the interface in `account.ts` has grown to accommodate child classes indicates some drift that we don't want associated with the basic concept of transaction signing. Given that we want to ultimately break up this class hierarchy I think some small overlap is OK in the interim.

packages/near-api-js/src/account_multisig.ts
```diff
@@ -495,38 +487,3 @@ export class Account2FA extends AccountMultisig {
 
 // helpers
 const toPK = (pk) => PublicKey.from(pk);
-const convertPKForContract = (pk) => pk.toString().replace('ed25519:', '');
```
andy-haynes, line 498: This no longer needs to be refactored as part of this PR

PR #1022, opened 2022-10-26
feat: Account2FA methods for batch converting multisig LAKs

andy-haynes:
## Motivation
This is the `near-api-js` component for https://pagodaplatform.atlassian.net/browse/WEP-192

## Description
This PR adds two methods to facilitate batch conversion of multisig LAKs back to FAKs outside of the workflow implemented in the `disable` methods. For users with > 48 keys, the current implementation will not work because it attempts to perform all key conversions at the time of disabling, exceeding the limit of 100 actions per transaction.

To mitigate this, a new workflow will be added to the wallet to determine whether multisig can be disabled with a single transaction (by calling `canDisableMultisig`). If not, the user will need to call `batchConvertKeys` to sign multisig transactions converting the target LAKs back to FAKs until only the current wallet signing key remains. Each batch of key conversions will require a 2FA prompt, but if progress is interrupted at any point it can be resumed with no side effects. Once all keys but the wallet signing key have been converted, the user may use the existing `Disable 2FA` workflow to complete the disable workflow.

The only caveat here is that by converting LAKs to FAKs outside of the disable workflow, multisig will remain deployed but converted access keys will now sign transactions directly as opposed to calling multisig methods. While this undermines the guarantees of 2FA, the workflow is only intended as a cleanup step immediately prior to rolling back deployment of the multisig contract. In addition the wallet signing key remains a LAK throughout this conversion process, ensuring that transactions signed in the wallet will still require a successful 2FA prompt. So long as users complete the batch conversion + disable in a single browser instance this will be the case.

## Checklist
- [x] Added automated tests

### DIFF
--- a/packages/near-api-js/src/account_multisig.ts
+++ b/packages/near-api-js/src/account_multisig.ts
@@ -17,7 +17,8 @@ export const MULTISIG_GAS = new BN('100000000000000');
 export const MULTISIG_DEPOSIT = new BN('0');
 export const MULTISIG_CHANGE_METHODS = ['add_request', 'add_request_and_confirm', 'delete_request', 'confirm'];
 export const MULTISIG_CONFIRM_METHODS = ['confirm'];
-const LAK_CONVERSION_LIMIT = 50;
+const LAK_CONVERSION_BATCH_SIZE = 50; // number of LAKs that can be converted (deleted + re-created) in one transaction
+const LAK_DISABLE_THRESHOLD = 48; // maximum number of LAKs that can be converted by the `delete` method
 
 type sendCodeFunction = () => Promise<any>;
 type getCodeFunction = (method: any) => Promise<string>;
@@ -400,12 +401,10 @@ export class Account2FA extends AccountMultisig {
      * Returns `false` if multisig cannot be disabled by calling disable() directly
      * i.e. account has too many LAKs to delete and re-create as FAKs for a single transaction
      */
-    async canDisableMultisig() {
+    async isConversionRequiredForDisable() {
         const keys = await this.get2faLimitedAccessKeys();
 
-        // there are max 4 actions required to disable 2FA on top of the key conversion (2 actions per key)
-        // (100 - 4) / 2 = 48 = LAK_CONVERSION_LIMIT - 2
-        return keys.length <= (LAK_CONVERSION_LIMIT - 2);
+        return keys.length > LAK_DISABLE_THRESHOLD;
     }
 
     /**
@@ -420,21 +419,28 @@ export class Account2FA extends AccountMultisig {
      *
      * To be deprecated after 2FA migration; this addresses a specific issue with multisig deployed via wallet.
      * @param signingPublicKey the public key used to sign transactions in the wallet
+     * @param [contractBytes]{@link https://github.com/near/near-wallet/blob/master/packages/frontend/src/wasm/main.wasm?raw=true}
+     * @param [cleanupContractBytes]{@link https://github.com/near/core-contracts/blob/master/state-cleanup/res/state_cleanup.wasm?raw=true}
      */
-    async batchConvertKeys(signingPublicKey: string) {
-        if (!signingPublicKey || !PublicKey.fromString(signingPublicKey)) {
-            throw new Error('invalid public key');
+    async batchConvertKeysAndDisable({ signingPublicKey, contractBytes, cleanupContractBytes }: {
+        signingPublicKey: string,
+        contractBytes: Uint8Array,
+        cleanupContractBytes: Uint8Array,
+    }) {
+        if (!signingPublicKey) {
+            throw new Error('the public key used to sign multisig transactions must be provided');
         }
 
         await this.validateMultisigState();
 
-        let batch = 0;
-        const keysToConvert = (await this.get2faLimitedAccessKeys())
+        let converted = 0;
+        const accessKeys = await this.get2faLimitedAccessKeys();
+        const keysToConvert = accessKeys
             .filter(({ public_key }) => public_key !== signingPublicKey);
 
-        while (batch < keysToConvert.length) {
+        while (converted < (accessKeys.length - LAK_DISABLE_THRESHOLD)) {
             const conversionActions = keysToConvert
-                .slice(batch, batch + LAK_CONVERSION_LIMIT)
+                .slice(converted, converted + LAK_CONVERSION_BATCH_SIZE)
                 .reduce((conversionActions: Action[], { public_key }) => {
                     conversionActions.push(deleteKey(PublicKey.from(public_key)));
                     conversionActions.push(addKey(PublicKey.from(public_key), fullAccessKey()));
@@ -447,8 +453,10 @@ export class Account2FA extends AccountMultisig {
                 actions: conversionActions
             });
 
-            batch += LAK_CONVERSION_LIMIT;
+            converted += LAK_CONVERSION_BATCH_SIZE;
         }
+
+        return this.disable(contractBytes, cleanupContractBytes);
     }
 
     async sendCodeDefault() {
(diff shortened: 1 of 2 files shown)


packages/near-api-js/src/account_multisig.ts
```diff
@@ -329,17 +330,23 @@ export class Account2FA extends AccountMultisig {
...
 
-    async get2faDisableKeyConversionActions() {
-        const { accountId } = this;
-        const accessKeys = await this.getAccessKeys();
-        const lak2fak = accessKeys
-            .filter(({ access_key }) => access_key.permission !== 'FullAccess')
+    async get2faLimitedAccessKeys() {
+        return (await this.getAccessKeys())
             .filter(({ access_key }) => {
+                if (access_key.permission === 'FullAccess') {
```
andy-haynes, line 336: Small change when I factored this out, seems excessive to do sequential filters so I consolidated them

packages/near-api-js/test/account_multisig.test.js
```diff
@@ -131,5 +145,99 @@ describe('account2fa transactions', () => {
...
+
+describe('2fa batch key conversion', () => {
+    let sender;
+    beforeAll(async () => {
+        sender = await testUtils.createAccount(nearjs);
+        sender = await getAccount2FA(sender);
+    });
+
+    beforeEach(async () => {
+        sender = await getAccount2FA(sender, undefined, false);
```
andy-haynes, line 159: This ended up being a much faster way to run these tests than creating an account per test as above. The `beforeAll` block creates the account and deploys the contract while the `beforeEach` only resets the instance between tests to retain isolation.

packages/near-api-js/src/account_multisig.ts
```diff
@@ -329,17 +330,23 @@ export class Account2FA extends AccountMultisig {
...
+                    return false;
+                }
+
                 const perm = (access_key.permission as FunctionCallPermissionView).FunctionCall;
-                return perm.receiver_id === accountId &&
+                return perm.receiver_id === this.accountId &&
                     perm.method_names.length === 4 &&
```
roshaans, line 342: What does `perm.method_names.length === 4` accomplish here?

packages/near-api-js/src/account_multisig.ts
```diff
@@ -348,16 +355,24 @@ export class Account2FA extends AccountMultisig {
         ];
     }
 
+    /**
```
roshaans, line 358: nice refactoring here :)

roshaans (commented):
Implementation looks great! I was unable to run the tests because of the following error `ENOENT: no such file or directory, open './test/data/multisig.wasm'`

packages/near-api-js/src/account_multisig.ts
```diff
@@ -381,6 +396,57 @@ export class Account2FA extends AccountMultisig {
...
+            const conversionActions = keysToConvert
+                .slice(batch, batch + LAK_CONVERSION_LIMIT)
+                .reduce((conversionActions: Action[], { public_key }) => {
+                    conversionActions.push(deleteKey(PublicKey.from(public_key)));
+                    conversionActions.push(addKey(PublicKey.from(public_key), fullAccessKey()));
+
+                    return conversionActions;
+                }, [] as Action[]);
+
+            await this.signAndSendTransaction({
```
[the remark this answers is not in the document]
andy-haynes: Correct, it will call the overridden method defined in `Account2FA`: https://github.com/near/near-api-js/blob/master/packages/near-api-js/src/account_multisig.ts#L240

andy-haynes:
> Implementation looks great! I was unable to run the tests because of the following error `ENOENT: no such file or directory, open './test/data/multisig.wasm'`

@roshaans looks like this is just a brittle test, I think you'll need to run it inside the `/test` directory for that path to resolve.

packages/near-api-js/src/account_multisig.ts
```diff
@@ -381,6 +396,61 @@ export class Account2FA extends AccountMultisig {
...
+            throw new Error('invalid public key');
+        }
+
+        await this.validateMultisigState();
+
+        let batch = 0;
+        const keysToConvert = (await this.get2faLimitedAccessKeys())
+            .filter(({ public_key }) => public_key !== signingPublicKey);
+
+        while (batch < keysToConvert.length) {
```
esaminu, line 435: Why not do this only while `canDisableMultisig` is `false`? There would be less prompts that way and UX would be better

esaminu (changes requested):
LGTM aside from the comment below if we're taking this approach. Another approach would be to just display some cleanup list like UI prompting the user to select keys to delete and then deleting them. But this approach is fine if we're ok with the middle state of having some FAKs and deleting keys based on public key is too technical for the user. It may be safe to assume that users with this many keys are somewhat technical though but this is a product question

packages/near-api-js/src/account_multisig.ts
```diff
@@ -381,6 +396,61 @@ export class Account2FA extends AccountMultisig {
...
+     *
+     * Note that this undermines the security provided by 2FA by converting existing multisig LAKs into FAKs
+     * capable of bypassing the 2FA prompt when signing transactions outside the wallet (transactions signed
+     * in the wallet will continue to require 2FA approval). This method should only be used to convert LAKs
+     * as a way to make 2FA disabling possible.
+     *
+     * To be deprecated after 2FA migration; this addresses a specific issue with multisig deployed via wallet.
+     * @param signingPublicKey the public key used to sign transactions in the wallet
+     */
+    async batchConvertKeys(signingPublicKey: string) {
```
esaminu, line 424: suggestion: maybe we should name the arg to something more general like `whitelist` or `keysToNotConvert` just so it's clearer that this is the key we're leaving behind and that its not being used to sign anything in the context of this method. It would also not tie us to the wallet scenario.

packages/near-api-js/src/account_multisig.ts
```diff
@@ -381,6 +396,61 @@ export class Account2FA extends AccountMultisig {
         });
     }
 
+    /**
+     * Returns `false` if multisig cannot be disabled by calling disable() directly
+     * i.e. account has too many LAKs to delete and re-create as FAKs for a single transaction
+     */
+    async canDisableMultisig() {
```
esaminu, line 403: IMO this should be made more specific like `isBelowMultisigActionsLimit` or similar. It provides more context to future readers since there's many reasons why multisig can't be disabled here.

packages/near-api-js/src/account_multisig.ts
andy-haynes: Good point. After reviewing the workflow in the wallet I think it would be better to just call the `disable` method in this method directly once the number of keys is under the threshold.

packages/near-api-js/src/account_multisig.ts
andy-haynes: This is a really specific scenario that only near.org wallet implements, but even then I don't see any reason to leave more than the one key you'll use to sign the transaction to disable. Specifying that one key is used to sign the transaction to disable multisig is more intentional and leaves less opportunity for keys to end up in an inconsistent state.

andy-haynes:
> Another approach would be to just display some cleanup list like UI prompting the user to select keys to delete and then deleting them

I like this idea, it would mean half as many transactions to confirm via 2FA. It would require a bit more work wallet side to notify the users since it inverts the current flow slightly. This approach has the side effect of dovetailing into the existing flow (i.e. first disable 2FA, then delete FAKs) while what you're suggesting is to delete LAKs, disable 2FA, then delete FAKs. I'd still prefer the batches be deleted without user interaction between batches, I don't think selecting keys to delete is a good UX when we can't provide any meaningful way to distinguish between keys.

I'll bring this up as an alternative since it potentially saves a lot of time in the migration workflow. I won't plan on merging this until we make a decision one way or the other.

andy-haynes:
After digging into this more on the `near-wallet` side I'm convinced this us a better fit for the `TwoFactor` class. It ends up being very straightforward to bring this implementation over.

Thank you @esaminu for the review, will carry over your feedback into the `near-wallet` implementation ❤️

Issue #1010, opened 2022-10-06
WalletConnection should return redirect links instead of redirect the whole page.


siendoricardo:
**Is your feature request related to a problem? Please describe.**
I want to be able to use the link in an iframe or popup instead of redirecting the complete webapp.

**Describe the solution you'd like**
Allowing WalletConnection to have methods that return the redirect URI instead of ultimately returning the web page.

**Describe alternatives you've considered**
Embedding the whole near-api-js connect into its own code space and running that into a popup. 

**Additional context**
https://github.com/near/near-api-js/blob/37ca2222e7afda3b77929d695b9d84bba21b418e/packages/near-api-js/src/wallet-account.ts#L199

denbite:
the idea of having methods for generating a Wallet URL without redirecting looks also reasonable to me and was implemented in https://github.com/near/near-api-js/pull/1260

PR #1025, opened 2022-11-01
feat(near-api-js): wallet sign-in page redirection is now optional

DevSlashRichie:
## Motivation
Sometimes, we want the sign-in process not to redirect the whole page, but to be open in a popup window.
As mentioned in https://github.com/near/near-api-js/issues/1010

## Description
She added to different methods (`requestSignInLink`, `requestSignInLink`, and `completeSignInWithAccessKeyLink`) to the `WalletConnection`, to get the needed URLs to the NEAR wallet in order to manually process them like in a popup or _blank redirection.

## Checklist
- [x] Added automated tests (The methods added are already tested as they are executed by the ones with a written test)

### DIFF
--- a/.changeset/beige-suits-march.md
+++ b/.changeset/beige-suits-march.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": major
+---
+
+Make `Account.signAndSendTransaction` `public` so transactions can be sent directly from `Account` instances
--- a/.changeset/fresh-fans-protect.md
+++ b/.changeset/fresh-fans-protect.md
@@ -1,5 +1,0 @@
----
-"near-api-js": patch
----
-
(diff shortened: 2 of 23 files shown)


frol:
@DevSlashRichie We are bringing near-api-js back to be actively maintained. This change makes total sense to me, but it needs to be rebased now (probably faster to just manually re-implement it as a lot of files have actually moved to other places). Do you have the capacity to rebase?

denbite:
the idea of having methods for generating a Wallet URL without redirecting looks also reasonable to me and was implemented in https://github.com/near/near-api-js/pull/1260

I'd suggest closing this one so we don't leave it out of date

vikinatora:
Resolved by #1260.

PR #1026, opened 2022-11-02
Updating getAccountId function

mohitxskull:
## Motivation
its easy to check for null value than empty string

## Description
return null value when there is no accountid to return so it gets easier to check it!

## Checklist
- [ ] Added automated tests


denbite:
I don't believe that this change will lead to a valuable outcome
@frol shall we proceed with closing it?

vikinatora:
I also don't support this change - not sure what OP means by easier to check. Closing the PR.

PR #1029, opened 2022-11-10
fix: `getNetworks` and `getAccounts` in `browser_local_storage_key_st…

therealharpaljadeja:
`getNetworks` and `getAccounts` is returning incorrect values, precisely there is a split string manipulation and the results are off by 1.

## Motivation
```js
let accounts = await keyStore.getAccounts("testnet");
```

The above line of code returns incorrect values.

## Description

The keyStore code basically does string manipulation using split and the developer made a mistake thinking that the structure of the keys in the keystore is [prefix]:[accountId]:[network].

However, there is one more string in between and the actual structure looks like this
[prefix]:ed25519:[account]:[network]

This is why when querying for accounts using `networkId` it returns nothing and getting networks returns accounts. Basically everything is offset by 1.

The below picture should explain it pretty well.

image
 

## Checklist
- [ ] Added automated tests

### DIFF
--- a/packages/near-api-js/src/key_stores/browser_local_storage_key_store.ts
+++ b/packages/near-api-js/src/key_stores/browser_local_storage_key_store.ts
@@ -1,18 +1,18 @@
-import { KeyStore } from './keystore';
-import { KeyPair } from '../utils/key_pair';
+import { KeyStore } from "./keystore";
+import { KeyPair } from "../utils/key_pair";
 
-const LOCAL_STORAGE_KEY_PREFIX = 'near-api-js:keystore:';
+const LOCAL_STORAGE_KEY_PREFIX = "near-api-js:keystore:";
 
 /**
  * This class is used to store keys in the browsers local storage.
- * 
+ *
  * @see [https://docs.near.org/docs/develop/front-end/naj-quick-reference#key-store](https://docs.near.org/docs/develop/front-end/naj-quick-reference#key-store)
  * @example
  * ```js
  * import { connect, keyStores } from 'near-api-js';
- * 
+ *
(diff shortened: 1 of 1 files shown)


therealharpaljadeja:
I think my code formatter also introduced new changes, how can I resolve that 😅

idea404:
Hi harpa, feel free to open an Issue describing the wider problem. 
As for the formatting, see steps followed [in CI](https://github.com/near/near-api-js/blob/master/.github/workflows/pull-request.yml) and run them on your branch on your machine

therealharpaljadeja:
Tried `pnpm lint`

image

andy-haynes:
@therealharpaljadeja there is an ongoing issue here around `lint` usage: https://github.com/near/near-api-js/issues/1032

andy-haynes:
@therealharpaljadeja 

> The keyStore code basically does string manipulation using split and the developer made a mistake thinking that the structure of the keys in the keystore is [prefix]:[accountId]:[network].

This is the expected format, yes. See the localstorage keys in `wallet.testnet.near.org` for an example.

> However, there is one more string in between and the actual structure looks like this
[prefix]:ed25519:[account]:[network]

That appears to be a public key, which would not be a valid NEAR account ID. It looks like you're dealing with malformed data.

 > This is why when querying for accounts using networkId it returns nothing and getting networks returns accounts. Basically everything is offset by 1.

This is not true for the implementations that currently rely on this, including `wallet.near.org`. These changes would break implementations using the correct storage.

Where are you seeing this error? My guess is that whatever is setting those values is using a public key instead of an account ID.

andy-haynes:
Please file an issue with the maintainer of the site using public keys in place of account IDs.

PR #1030, opened 2022-11-15
refactor: json-rpc-provider package

andy-haynes:
## Motivation
Moves the `JsonRpcProvider` class to its own package.

## Description
This PR adds two packages, `@near-js/core` and `@near-js/json-rpc-provider` to the monorepo. `json-rpc-provider` primarily contains the `JsonRpcProvider` while `core` contains its dependencies.

Generally code has been lifted from `near-api-js` into one of these packages and left intact, then the reference in `near-api-js` is replaced with one to the new package.

## Checklist
- [ ] Added automated tests


packages/json-rpc-provider/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@near-js/json-rpc-provider",
+  "version": "0.0.1",
+  "description": "Client for communicating with the NEAR RPC using JSON-RPC",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "pnpm compile",
+    "compile": "tsc -p tsconfig.json"
```
idea404, line 8: any ideas on testing here and core and naj? splitting them like here with the code would make sense to me

idea404 (commented):
Like the approach, deps for core look clean, missing tests and CI specs

packages/json-rpc-provider/package.json
andy-haynes: Since the break was relatively clean I think it's just a matter of moving tests directly out of `near-api-js` but I haven't looked closely at the tests to verify. Some integration tests may end up staying in `near-api-js` if their dependencies are too thorny but I expect most can just live in their new packages.

However I'm first trying to work out a reasonable approach to CI dependencies with `pnpm` and `turborepo`. I may need to create NPM packages first if it ends up being a bootstrapping issue but in theory it should all work locally. Specifically `near-api-js` has an install-time dependency on `@near-js/core`, which must first have its own dependencies installed and be built (transpiled into JS). It seems like since the install and build steps are independent, they lack the awareness of the more complex relationship between dependencies.

andy-haynes: Fixed by removing the `prepare` command 🎉 

`prepare` was running the `build` command, which requires that `install` has already run... but `prepare` runs automatically when `install` is run, so every clean `build` fails because `install` that won't run until `build` succeeds. By removing `prepare`, `pnpm` can do its thing and manage workspace dependencies.

andy-haynes: I've moved the low hanging fruit, some of the others will require a more thoughtful refactor.

andy-haynes:
Closed in favor of #1056

PR #1033, opened 2022-11-15
feat(async_keys): keys should be async to allow HSMs

waynenilsen:
## Motivation

We use AWS KMS to secure our private keys. Thankfully secp256k1 is supported by the near blockchain. However, this client is not robust enough to allow for a custom AWS KMS signer. There is a built in assumption that `sign` and `verify` are _always_ sync which is not true in our case.

There is another problematic behavior that may need to be addressed in the presence of a system that does not allow (or want) private key export which is the `toString` function is exporting the private key (which we will not have). That can be handled by throwing at runtime if it comes down to that but it is not particularly desirable. Private key export via `toString` feels like quite a bad idea generally anyway.

## Description

Update signatures to return promises. For local implementations use `Promise.resolve`.

This is, unfortunately, a breaking change I think and technically would require a major version bump.

@volovyks for review or to forward to someone 🙏 🙇 

Looking for feedback / thoughts on this

## Checklist
- [ ] Added automated tests (not done but sortof n/a due to this being a refactor)
- [ ] Manually tested the change (same)

### DIFF
--- a/packages/near-api-js/src/key_stores/in_memory_key_store.ts
+++ b/packages/near-api-js/src/key_stores/in_memory_key_store.ts
@@ -44,7 +44,7 @@ export class InMemoryKeyStore extends KeyStore {
      * @param keyPair The key pair to store in local storage
      */    
     async setKey(networkId: string, accountId: string, keyPair: KeyPair): Promise<void> {
-        this.keys[`${accountId}:${networkId}`] = keyPair.toString();
+        this.keys[`${accountId}:${networkId}`] = await keyPair.toString();
     }
 
     /**
--- a/packages/near-api-js/src/key_stores/unencrypted_file_system_keystore.ts
+++ b/packages/near-api-js/src/key_stores/unencrypted_file_system_keystore.ts
@@ -98,7 +98,7 @@ export class UnencryptedFileSystemKeyStore extends KeyStore {
      */
     async setKey(networkId: string, accountId: string, keyPair: KeyPair): Promise<void> {
         await ensureDir(`${this.keyDir}/${networkId}`);
-        const content: AccountInfo = { account_id: accountId, public_key: keyPair.getPublicKey().toString(), private_key: keyPair.toString() };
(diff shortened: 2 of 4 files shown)


volovyks:
@DavidM-D do we have someone working on NAJ at the moment?

andy-haynes:
@waynenilsen thank you for your contribution!

Rather than changing the `KeyPair` class I'm thinking what we really want here is a new class derived from `Signer`. `KeyPair` instances rely on knowing and making available their private key, which puts them fundamentally at odds with hardware wallets. They are better left as abstractions around elliptic curves.

`Signer` already has asynchronous methods for `getPublicKey` and `signMessage`, which is what we expose in our [Ledger implementation](https://github.com/near/near-ledger-js/blob/master/index.js). This also precludes making a breaking change here as we'd only be adding a new implementation for a `LedgerSigner`. What do you think about this approach instead?

waynenilsen:
Yes the signer abstraction is sufficient this is not strictly necessary. I will say however that I think it makes sense to support asynchronous keystores of various kinds. Close if you wish, I have actually already started forward on the custom Signer path which can definitely be implemented outside this library.

andy-haynes:
Hey @waynenilsen , I got my wires crossed trying to catch up on tickets prior to leave. I conflated your changes with Ledger, which was an issue that arose independently and of course has nothing to do with what you want to accomplish with KMS. Anyways apologies for the misunderstanding.

Looking at your issue in more depth, I still think a `KmsSigner` class is the best fit here. Is there something specific you're trying to achieve with KeyPair? I ask because it appears that if you don't require a `KeyStore` instance then you shouldn't really need to interact with `KeyPair`s. With a new `KmsSigner`  in place of `InMemorySigner` it seems like you'd avoid a lot of that mess.

Implementing secp256k1 would be a separate issue, which was started (https://github.com/near/near-api-js/pull/985) but hasn't been active for some time; I'll follow up on this. In your case you'd only need the implementation for the `PublicKey.verify` method.

PR #1045, opened 2022-12-19
feat: add support for ABI

itegulov:
## Motivation
[NEAR ABI](https://github.com/near/abi) is a language-agnostic format describing the interface of a contract. It enables developers not to have to manually specify the list of `viewMethods` and `changeMethods`. Additionally, it makes it possible to add caller-side type safety by validating arguments against their JSON Schema counterparts before making the call.

## Description
This PR adds a new optional field `abi` that the user can pass when instantiating a `Contract` that is used to populate `viewMethods`/`changeMethods` with new entries. [`ajv`](https://ajv.js.org/) is used as the JSON Schema validation engine due to its flexible support of various JSON Schema features.

Validation is still very much PoC and needs to be tested on a bigger set of samples. 

## Checklist
- [x] Added automated tests


packages/near-api-js/tsconfig.json
```diff
@@ -12,7 +12,7 @@
         "alwaysStrict": true,
         "outDir": "./lib",
         "declaration": true,
-        "preserveSymlinks": true,
+        "preserveSymlinks": false,
```
itegulov, line 15: I have encountered a weird bug where adding `ajv` dependency messes up with `pnpm`'s dependency model. It can't find some of the `ajv` transitive dependencies even though they are clearly there. I and @morgsmccauley have debugged the issue and found out that toggling this line fixes this.

morgsmccauley: Adding more context while it's fresh:

This seems to be an issue within `pnpm` that has already been fixed: https://github.com/pnpm/pnpm/issues/496, so not sure why this is still a problem.

`pnpm` will place all 'visible' dependencies in their own `node_modules` folder and then symlink to that directory from the top level:
```
node_modules/
|  ajv ➛ node_modules/.pnpm/ajv@8.11.2/node_modules/ajv
|  .pnpm/
|  |  ajv@8.11.2/
│  |  |  node_modules/
|  │  |  |  ajv/
|  |  |  |  fast-deep-equal/ ➛ node_modules/.pnpm/fast-deep-equal@3.1.3/node_modules/fast-deep-equal
|  |  |  |  json-schema-traverse/ ➛ node_modules/.pnpm/json-schema-traverse@1.0.0/node_modules/json-schema-traverse
|  |  |  |  require-from-string/ ➛ node_modules/.pnpm/require-from-string@2.0.2/node_modules/require-from-string
|  |  |  |  uri-js/ ➛ node_modules/.pnpm/uri-js@4.4.1/node_modules/uri-js
```
This strategy makes all transitive dependencies inaccessible from your package code while still making them accessible to the dependency itself. In the above example, all of `ajv`s dependencies are next to it (in the bottom most directory), meaning `ajv` should be able to `import`/`require` it.

But when we use TS to build, we get the following error, which complains that `ajv` can't access `uri-js`. This doesn't make any sense because we can see it within `node_modules/`:
```sh
> near-api-js@1.1.0 compile /Users/morganmccauley/Developer/Repositories/near-api-js/packages/near-api-js
> tsc -p ./tsconfig.json

node_modules/ajv/dist/compile/resolve.d.ts(3,36): error TS2307: Cannot find module 'uri-js' or its corresponding type declarations.
node_modules/ajv/dist/types/index.d.ts(1,22): error TS2307: Cannot find module 'uri-js' or its corresponding type declarations.
node_modules/near-abi/lib/index.d.ts(1,29): error TS2307: Cannot find module 'json-schema' or its corresponding type declarations.
```

Setting [preserveSymlinks](https://www.typescriptlang.org/tsconfig#preserveSymlinks) to `false` is counter-intuitive here, but works.. When disabled, module resolution happens from where the symbolic link is defined, rather from where it actually resolves to. I would think that we would want the latter, since that is where the transitive dependencies are defined, but this doesn't seem to be the case.

maybe we should raise this issue within `pnpm` itself 🤔

packages/near-api-js/src/contract.ts
```diff
@@ -90,45 +119,92 @@ export class Contract {
     constructor(account: Account, contractId: string, options: ContractMethods) {
         this.account = account;
         this.contractId = contractId;
-        const { viewMethods = [], changeMethods = [] } = options;
-        viewMethods.forEach((methodName) => {
-            Object.defineProperty(this, methodName, {
+        const { viewMethods = [], changeMethods = [], abi: abiRoot } = options;
```
andy-haynes, line 122: Is there a use case for calling the constructor with `options` that have values in `viewMethods`/`changeMethods` and also methods specified in `abiRoot.body.functions`? I see that if there is any overlap in method names the constructor will break trying to redefine properties.

It seems odd to now have two ways of initializing contract methods. If they are mutually exclusive then maybe it makes more sense to only source methods from `abiRoot.body.functions` when provided. This is basically equivalent to changing the constructor signature to `Contract(Account, string, ContractMethods | AbiRoot)` which feels a bit more deliberate.

What do you think?

packages/near-api-js/src/contract.ts
```diff
@@ -90,45 +119,92 @@ export class Contract {
...
+        // are added explicitly just to reduce the amount of warnings as these are very popular
+        // types.
+        const ajv = new Ajv({
+            strictSchema: false,
+            formats: {
+                uint32: true,
+                uint64: true
+            }
+        });
+        addFormats(ajv);
```
andy-haynes, line 148: What would you think about putting this in a private method that gets invoked here? If this is going to be extended over time it would be nice for it to have its own dedicated space, this constructor is already doing all the heavy lifting 😅

packages/near-api-js/src/contract.ts
```diff
@@ -90,45 +119,92 @@ export class Contract {
...
                 writable: false,
                 enumerable: true,
-                value: nameFunction(methodName, async (args: object = {}, options = {}, ...ignored) => {
+                value: nameFunction(name, async (args: object = {}, options = {}, ...ignored) => {
                     if (ignored.length || !(isObject(args) || isUint8Array(args)) || !isObject(options)) {
                         throw new PositionalArgsError();
                     }
+
+                    if (abi) {
+                        if (abi.params && abi.params.serialization_type !== AbiSerializationType.Json) {
```
andy-haynes, line 160: 💄 You can use optional chaining here
```js
abi.params?.serialization_type !== AbiSerializationType.Json)
```

andy-haynes (changes requested):
Awesome work @itegulov! I have some feedback around tests and how this extends the `Contract` interface but it's looking good.

Would you mind please also running `npx changeset` to register this as a `minor` change?

packages/near-api-js/test/contract_abi.test.js
```diff
@@ -0,0 +1,211 @@
...
+});
+
+describe('add', () => {
+    test('can be called successfully', async () => {
+        await contract.add({ a: [1, 2], b: [3, 4] });
+    });
+
+    test('throws ArgumentSchemaError if required argument was not supplied', async () => {
+        try {
+            await contract.add({});
```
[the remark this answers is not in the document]
itegulov: Oh wow, that is so much better 😅 

I was mostly inspired by the existing tests like [this one](https://github.com/near/near-api-js/blob/master/packages/near-api-js/test/contract.test.js#L66-L71), so I guess they should be refactored too some day

packages/near-api-js/src/contract.ts
itegulov: Yeah, so my intention was to leave an option to provide ABI **and** some custom methods on the off-chance that they are not in the ABI. But I guess this might not make sense until we actually encounter a situation where a user needs this, so I would be totally down to change this to `ContractMethods | AbiRoot`.

andy-haynes: How likely is it that an ABI wouldn't include all methods? I'm not really familiar with how you would get the ABI in the first place, i.e. if you create them dynamically from a deployed contract or need to generate them statically.

itegulov: You generate them statically from the source code. Assuming there are no bugs in the ABI generation code it should cover all methods exposed by the contract. That being said, it does not stop the user from injecting new methods into the resulting wasm through some postprocessing means.

itegulov: So, overall, I would say it is fairly unlikely to happen (I have never seen this pattern among our users), and I guess if you are really in a pickle, you could just instantiate two separate `Contract`s for now.

itegulov: Ok, I went ahead and prohibited the conflicting usage of `viewMethods`/`changeMethods` with ABI. LMK if you disagree.

packages/near-api-js/src/contract.ts
```diff
@@ -13,6 +16,52 @@ function nameFunction(name: string, body: (args?: any[]) => any) {
...
 
+function validateArguments(args: object, abiFunction: AbiFunction, ajv: Ajv, abiRoot: AbiRoot) {
+    if (!isObject(args)) return;
+
+    if (abiFunction.params?.serialization_type !== AbiSerializationType.Json) {
+        throw new UnsupportedSerializationError(abiFunction.name, abiFunction.params.serialization_type);
+    }
+
+    if (abiFunction.result?.serialization_type !== AbiSerializationType.Json) {
+        throw new UnsupportedSerializationError(abiFunction.name, abiFunction.params.serialization_type);
```
andy-haynes, line 27: Should this match the predicate?

```suggestion
        throw new UnsupportedSerializationError(abiFunction.name, abiFunction.result?.serialization_type);
```

packages/near-api-js/src/contract.ts
```diff
@@ -13,6 +16,52 @@ function nameFunction(name: string, body: (args?: any[]) => any) {
     }[name];
 }
 
+function validateArguments(args: object, abiFunction: AbiFunction, ajv: Ajv, abiRoot: AbiRoot) {
+    if (!isObject(args)) return;
+
+    if (abiFunction.params?.serialization_type !== AbiSerializationType.Json) {
+        throw new UnsupportedSerializationError(abiFunction.name, abiFunction.params.serialization_type);
```
andy-haynes, line 23: I missed that these get referenced in the error when I suggested optional chaining - these will need to be guarded as well to avoid throwing an exception while throwing an exception.

```suggestion
        throw new UnsupportedSerializationError(abiFunction.name, abiFunction.params?.serialization_type);
```

andy-haynes (changes requested):
Looks good, I just missed a couple things in `validateArguments`

packages/near-api-js/src/contract.ts
itegulov: Wait, I just realized that this is not equivalent to the original line. `abi.params?.serialization_type !== AbiSerializationType.Json` also returns true if `abi.params`/`abi.result` is undefined which is not what we want.

packages/near-api-js/src/contract.ts
itegulov: Good catch!

packages/near-api-js/src/contract.ts
itegulov: Yeah, see my comment in the other thread. The flow was not actually supposed to reach this line if `params`/`result` is missing

packages/near-api-js/src/contract.ts
andy-haynes: 🤦 my bad!

andy-haynes (approved):

PR #1066, opened 2023-01-31
[WIP] feat: contract.ts for view without required account

strangeweb3guy:
view-contract.ts

## Pre-flight checklist

- [✅] I have read the [Contributing Guidelines on pull requests](https://github.com/near/near-api-js/blob/master/CONTRIBUTING.md).
- [✅] Commit messages follow the [conventional commits](https://www.conventionalcommits.org/) spec
- [] **If this is a code change**: I have written unit tests.
- [✅] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved my working plan.

## Motivation

Discussion - https://github.com/near/near-api-js/discussions/1064

## Test Plan

WIP will add tests later

### DIFF
--- a/packages/near-api-js/src/contract.ts
+++ b/packages/near-api-js/src/contract.ts
@@ -6,6 +6,24 @@ import { AbiFunction, AbiFunctionKind, AbiRoot, AbiSerializationType } from 'nea
 import { Account } from './account';
 import { getTransactionLastResult } from './providers';
 import { PositionalArgsError, ArgumentTypeError, UnsupportedSerializationError, UnknownArgumentError, ArgumentSchemaError, ConflictingOptions } from './utils/errors';
+import { Connection } from './connection';
+import { viewFunction } from './utils/view-function';
+
+function isAccount(connection: Account | Connection): connection is Account {
+    return 'connection' in connection;
+}
+
+function isContractMethods(options: ContractMethods | ViewContractMethods): options is ContractMethods {
+    return 'changeMethods' in options;
+}
+
+function isFullContract (options: ViewContractOptions | FullContractOptions): options is FullContractOptions {
+    return 'account' in options;
+}
+
+function isViewContract (options: ViewContractOptions | FullContractOptions): options is ViewContractOptions {
+    return 'connection' in options;
+}
 
 // Makes `function.name` return given name
 function nameFunction(name: string, body: (args?: any[]) => any) {
@@ -80,15 +98,14 @@ interface ChangeMethodOptions {
 export interface ContractMethods {
     /**
      * Methods that change state. These methods cost gas and require a signed transaction.
-     * 
+     *
      * @see {@link account!Account.functionCall}
      */
     changeMethods: string[];
 
     /**
      * View methods do not require a signed transaction.
-     * 
-     * @see {@link account!Account#viewFunction}
+     *
      */
     viewMethods: string[];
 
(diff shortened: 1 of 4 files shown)


packages/near-api-js/src/contract.ts
```diff
@@ -98,14 +128,27 @@ export interface ContractMethods {
...
+    connection: Connection;
+    contractId: string;
+    viewContractMethods: ViewContractMethods;
+}
+
+interface FullContractOptions {
+    account: Account;
+    contractId: string;
+    contractMethods: ContractMethods;
+}
```
andy-haynes, line 141: Given what I've said about removing `Account` from any method/constructor signatures, I don't think there's a strong case for making a distinction between a view-only `Contract` and one that also has change methods. For now it's OK for there to be an implicit understanding that the `Connection` instance must have a `Signer` attached to it in order to sign transactions.

Later on we can introduce breaking changes that make this more explicit, e.g. by providing only the `JsonRpcProvider` rather than an entire `Connection` and by expanding the arguments to change methods to include the signing account ID and a way to sign the transaction.

packages/near-api-js/src/utils/view-function.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @param viewFunctionCallOptions.contractId NEAR account where the contract is deployed
+ * @param viewFunctionCallOptions.methodName The view-only method (no state mutations) name on the contract as it is written in the contract code
+ * @param viewFunctionCallOptions.args Any arguments to the view contract method, wrapped in JSON
+ * @param viewFunctionCallOptions.parse Parse the result of the call. Receives a Buffer (bytes array) and converts it to any object. By default result will be treated as json.
+ * @param viewFunctionCallOptions.stringify Convert input arguments into a bytes array. By default the input is treated as a JSON.
+ * @param viewFunctionCallOptions.jsContract Is contract from JS SDK, automatically encodes args from JS SDK to binary.
+ * @param viewFunctionCallOptions.blockQuery specifies which block to query state at. By default returns last "optimistic" block (i.e. not necessarily finalized).
+ * @returns {Promise<any>}
+ */
+export async function viewFunction({
```
andy-haynes, line 34: I agree that this needs to be refactored out of `Account` but that's not something that needs to be done as part of no longer requiring `Contract` to be instantiated with an `Account`. My main concern is that this carries too much baggage with it from `Account` and needs to be deconstructed rather than just moved in place.

As with `_changeMethod`, we can remove the need for a provided `Account` object and instantiate one when making the view call:
```js
return new Account(this.connection, 'dontcare').viewFunction({
```

andy-haynes (changes requested):
Thank you for the submission @kiskesis!

If I'm understanding the motivation right, based on this implementation and #1064, the main frustration point is that instantiating `Contract` requires an `Account` instance which is entirely asinine for view contract methods. However rather than refactoring view functions out of `Account`, I think the ideal short-term (i.e. non-breaking change) solution is to remove the `Account` requirement entirely from the constructor. Long-term we'll want something with a much simpler interface (i.e. no awareness of `jsContract`, no `Connection` object, etc.).

I would argue that just by addressing the original point of getting the hard `Account` requirement out of the constructor does more to make view functions accessible than refactoring the method does. I'd still like to see that refactor at some point, but in the interest of keeping PRs light I think it fits better as a separate piece of work.

packages/near-api-js/src/utils/view-function.ts
kiskesis: Totally agree, it should be part of different PR

vikinatora:
Resolved by #1325

PR #1071, opened 2023-02-26
fix: fix validators function to accept null to show latest block info

kiskesis:
## Pre-flight checklist

- ✅ I have read the [Contributing Guidelines on pull requests](https://github.com/near/near-api-js/blob/master/CONTRIBUTING.md).
- ✅ Commit messages follow the [conventional commits](https://www.conventionalcommits.org/) spec
- ✅ **If this is a code change**: I have written unit tests.
- ✅ **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

PR #1096, opened 2023-03-31
fix: fix validators function to accept null to show latest block info

andy-haynes:
This is the rebased PR for #1071 since splitting up `near-api-js` into packages.


frol (approved):

andy-haynes:
Leaving this open at the moment as a reference but will not merge in favor of rebasing and merging #1071

PR #1097, opened 2023-04-01
feat: meta transactions

andy-haynes:
This PR implements functionality for delegate actions and meta transactions for [NEP 366: Meta Transactions](https://github.com/near/NEPs/blob/master/neps/nep-0366.md).

For `@near-js/transactions`, this adds:
- new action classes and creator for `DelegateAction` actions
- methods for constructing, signing, and encoding `DelegateAction`s
- new schema entries for `DelegateAction`
- new classes for [NEP 461](https://github.com/near/NEPs/pull/461) prefixes

In `@near-js/accounts` I've added `Account` methods:
- `signedDelegate` builds and signs a delegate action
- `signAndSendMetaTransaction` builds and signs a meta transaction derived from the delegate action

A new cookbook example in included to demonstrate the usage of the new `Account` methods.


packages/transactions/src/schema.ts
```diff
@@ -15,13 +15,45 @@ import {
...
+    return serialize(SCHEMA, transaction);
+}
+
+/**
+ * Borsh-decode a Transaction instance from a buffer
+ * @param bytes Buffer data to be decoded
+ */
+export function decodeTransaction(bytes: Buffer) {
+    return deserialize(SCHEMA, Transaction, bytes);
+}
```
andy-haynes, line 49: Initially I wanted to opportunistically refactor the Transaction and SignedTransaction classes to decouple the class definitions from SCHEMA, but ultimately decided it was too much unrelated effort for a half measure.

I opted to leave these new methods in place and use them in favor of the encode/decode class methods but they're not strictly relevant to this PR.

packages/accounts/src/account.ts
```diff
@@ -145,6 +151,14 @@ interface ActiveDelegatedStakeBalance {
     total: BN | string;
 }
 
+// by default expire meta transactions 600 blocks after the current block height
+const DEFAULT_META_TRANSACTION_BLOCK_HEIGHT_TTL = 600;
```
andy-haynes, line 155: Is it helpful to have a default in the client? (ideally co-located with other relayer configuration)

packages/accounts/src/account.ts
```diff
@@ -455,6 +469,85 @@ export class Account {
...
+     * @param signedDelegate SignedDelegate instance containing the information and verification required for relayer execution of meta transactions
+     */
+    async signAndSendSignedDelegate({
+        actions,
+        blockHeightTtl,
+        receiverId,
+        relayerUrl,
+    }: { relayerUrl: string } & SignedDelegateOptions): Promise<FinalExecutionStatus> {
+        const signedDelegate = await this.signedDelegate({ actions, blockHeightTtl, receiverId });
+        return fetchJson(relayerUrl, JSON.stringify(Array.from(encodeSignedDelegate(signedDelegate))));
```
andy-haynes, line 531: Current implementation of relayers requires the JSON body be a serialized byte array (i.e. not wrapped in an object).

NEP-366 describes an interface for this, so it's possible the implementation will change:
```rs
struct DelegateActionMessage {
    signed_delegate_action: SignedDelegateAction
}
```

> - User forms `DelegateActionMessage` with the `SignedDelegateAction`.
> - User sends `DelegateActionMessage` data to the relayer.

packages/cookbook/transactions/meta-transaction-relayer.js
```diff
@@ -0,0 +1,50 @@
...
+        receiverId,
+        relayerUrl: 'https://relayer.org/relay',
+    });
+
+    console.log(response);
+    return response;
+}
+
+module.exports = {
+    sendNearThroughRelayer,
```
andy-haynes, line 24: I created the `cookbook` examples from a new template from an in-review PR that supports testing snippets here. When that PR is landed I can add integration tests for these.

packages/transactions/src/sign.ts
```diff
@@ -1,11 +1,26 @@
...
-import { Action } from './actions';
+import { Action, SignedDelegate } from './actions';
 import { createTransaction } from './create_transaction';
-import { SCHEMA, Signature, SignedTransaction, Transaction } from './schema';
+import type { DelegateAction } from './delegated.js';
+import { encodeDelegateAction, encodeTransaction, SignedTransaction, Transaction } from './schema';
+import { Signature } from './signature';
+
+interface MessageSigner {
+    sign(message: Uint8Array): Promise<Uint8Array>;
```
andy-haynes, line 12: I'm not super attached to this, it was just my way of inching away from coupling transaction signing methods and network while avoiding stepping into the current state of keystores and their 1:1 account-network mapping. Definitely open to alternatives or tweaks to this approach.

morgsmccauley (commented):
Good job on this! Most looks good, just a couple questions around the necessity of some parts :)

morgsmccauley:
sorry misclick

.changeset/lovely-balloons-talk.md
```diff
@@ -0,0 +1,8 @@
+---
+"@near-js/transactions": minor
+"@near-js/accounts": minor
+"@near-js/providers": patch
+"@near-js/cookbook": patch
+---
+
+Add support for delegate actions and meta transactions
```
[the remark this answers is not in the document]
andy-haynes: Interesting, I made the distinction mostly in the context of relayers:
- **delegate action** is an action signed by the relay caller
- **meta transaction** is a transaction signed by the relayer

I omitted the hyphen to reflect its title and usage in [NEP 366](https://github.com/near/NEPs/blob/master/neps/nep-0366.md)

packages/accounts/src/account.ts
```diff
@@ -455,6 +469,85 @@ export class Account {
...
+    }
+
+    /**
+     * Sign and send a meta transaction, consisting of a single signed delegate action, on behalf of a sender (i.e. as a meta transaction relayer)
+     *
+     * @param signedDelegate SignedDelegate action to wrap in the meta transaction
+     */
+    async signAndSendMetaTransaction(signedDelegate: SignedDelegate): Promise<FinalExecutionOutcome> {
```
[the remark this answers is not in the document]
andy-haynes: That makes sense, I will move it to the cookbook.

packages/transactions/src/delegated.ts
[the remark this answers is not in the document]
andy-haynes: I like it 😎

packages/accounts/src/account.ts
```diff
@@ -455,6 +469,85 @@ export class Account {
...
+    /**
+     * Compose, sign, and submit (to relayer) a set of actions to be executed in a meta transaction
+     *
+     * @param actions Actions to be included in the meta transaction
+     * @param blockHeightTtl Number of blocks past the current block height for which the SignedDelegate action may be included in a meta transaction
+     * @param receiverId Receiver account of the meta transaction
+     * @param relayerUrl Relayer URL endpoint responsible for building and executing meta transactions
+     * @param signedDelegate SignedDelegate instance containing the information and verification required for relayer execution of meta transactions
+     */
+    async signAndSendSignedDelegate({
```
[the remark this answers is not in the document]
andy-haynes: That's a good point, I think to do this right you'd want a relayer interface that could be extended to talk to non-REST endpoints. Given the nascent stage of things it's best not to have strong opinions up front 😄

morgsmccauley (dismissed):
Good job 👏

packages/transactions/src/delegate.ts
```diff
@@ -0,0 +1,60 @@
...
+    | FunctionCall
+    | Stake
+    | Transfer;
+
+export class DelegateAction extends Assignable {
+    senderId: string;
+    receiverId: string;
+    actions: Array<NonDelegateAction>;
+    nonce: BN;
+    maxBlockHeight: number;
```
MaximusHaximus, line 31: I think this must be a BN type also; block heights can be huge (U64), and looking at the code below we do use BN to set this property

packages/transactions/src/actions.ts
```diff
@@ -2,6 +2,9 @@ import { PublicKey } from '@near-js/crypto';
 import { Assignable } from '@near-js/types';
 import BN from 'bn.js';
 
+import { DelegateAction } from './delegate.js';
+import { Signature } from './signature.js';
```
MaximusHaximus, line 6: Should these file extensions be .ts?

packages/transactions/src/delegate.ts
```diff
@@ -0,0 +1,60 @@
...
+import {
+    CreateAccount,
+    DeployContract,
+    FunctionCall,
+    Transfer,
+    Stake,
+    AddKey,
+    DeleteKey,
+    DeleteAccount,
+} from './actions.js';
```
MaximusHaximus, line 14: Should this file extension be .ts?

packages/transactions/src/prefix.ts
```diff
@@ -0,0 +1,36 @@
...
+abstract class ActionableMessagePrefix extends NEPPrefix {
+    /** Given the NEP number, set the prefix using 2^30 as the offset **/
+    protected constructor(prefix: number) {
+        super({ prefix: ACTIONABLE_MESSAGE_BASE + prefix });
+    }
+}
+
+/**
+ * Class for constructing prefixes on non-actionable (off-chain) messages
+ * @todo uncomment when off-chain messages are supported
```
MaximusHaximus, line 26: Is there an NEP, release version or date where we expect this to be true that we could reference here, or is this an indefinite 'some day' kind of thing? :)

MaximusHaximus (changes requested):
Nice work -- looking good.  Just spotted a few things which I added comments about in-line and was thinking about some naming tweaks.

Just one thing that I noticed while looking through the PR was the DelegateAction type being local to the transactions package -- I would've expected it to be in the `@near-js/types` package which has a Transaction type that would need to reference all of the action types for accuracy.

I think to avoid circular references between the types package and the rest of the packages we're going to need to pull almost all of the types into the types package and only keep internal/local types in the individual packages that use the @near-js/types package -- I think this is why a Transaction type exists in the types package, but because Action types don't, the Transaction type in the types package defines `actions` as an <any> type.  We should have a chat about this and it's not any kind of a blocker for this PR to land for us to iterate on.

packages/transactions/src/schema.ts
```diff
@@ -128,4 +169,19 @@ export const SCHEMA = new Map<Class, any>([
     [DeleteAccount, { kind: 'struct', fields: [
         ['beneficiaryId', 'string']
     ]}],
+    [DelegateAction, { kind: 'struct', fields: [
+        ['senderId', 'string'],
+        ['receiverId', 'string'],
+        ['actions', [Action]],
```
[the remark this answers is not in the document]
andy-haynes: `DelegableAction` is a union type but a class is required here.

packages/transactions/src/delegate.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export type NonDelegateAction =
```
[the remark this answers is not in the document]
andy-haynes: Looking at it more closely I see it's actually ignoring the union types altogether:
- `{ actions: Array<AddKey> }`  => `actions = [new SignedDelegate({})]` ❌ 
- `{ actions: Array<AddKey | DeleteKey> }`  => `actions = [new SignedDelegate({})]` ✅ 

It looks like by having a union with > 1 members, the common shape is demoted to their common `Assignable` base class and so anything with `constructor(properties: any)` will pass the union. This wasn't something I knew about union types 💡 

So with the current implementation being set dressing, I think options are:
- double down on this approach and incorporate delegability into the class hierarchy
- implement the chain-side logic in the client with runtime checks that throw exceptions when a non-delegate action is added to the signed delegate
- ignore for now and let the chain fail them

andy-haynes:
> we're going to need to pull almost all of the types into the types package and only keep internal/local types in the individual packages that use the @near-js/types package

The crux of the issue is that many of these are classes implementing methods with heavy dependencies not appropriate for `@near-js/types`. `Transaction` is a good example but `PublicKey` is even better. To do it cleanly requires another substantial refactor with breaking changes.

>  I think this is why a Transaction type exists in the types package, but because Action types don't, the Transaction type in the types package defines actions as an type. 

The `Transaction` type in `@near-js/types` is a duplicate definition that originally came from the `providers.ts` file before the refactor. They should be merged when appropriate.

packages/transactions/src/delegate.ts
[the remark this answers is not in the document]
MaximusHaximus: Let's let the chain fail them and re-visit the classes later for ideal typesafety :+1:

MaximusHaximus (approved):
:+1: I think this is shippable at this stage 💖

PR #1103, opened 2023-04-05
feat: implement client API and types for next light client block

austinabell:
## Pre-flight checklist

  - No, but these are types based on the stable NEAR RPC -- I can open issue if necessary

## Motivation

I needed the types for an API I'm building, and this functionality would be needed for client-side light client block validation

## Test Plan

Unit tests match the format and specificity of other related ones, I believe

## Related issues/PRs

Not an issue or PR, but figured I'd link that this is related to (and necessary for) https://near.social/#/devgovgigs.near/widget/gigs-board.pages.Post?id=148


packages/near-api-js/test/providers.test.js
```diff
@@ -295,6 +295,41 @@ test('json rpc light client proof', async() => {
...
 });
 
+test('json rpc get next light client block', withProvider(async (provider) => {
+    const stat = await provider.status();
+
+    // Get block in at least the last epoch (epoch duration 43,200 blocks on mainnet and testnet)
+    const height = stat.sync_info.latest_block_height;
+    const protocolConfig = await provider.experimental_protocolConfig({ finality: 'final' });
+
+    if (height < protocolConfig.epoch_length) {
```
andy-haynes, line 305: I'm a little wary of adding a test like this, branching execution paths based on state extrinsic to the test environment means that for any given test run we are only running an unknown subset of these paths. We should really only execute known code paths to ensure identical, deterministic coverage on every run.

Even as an integration test I'm not convinced there's lots of value to testing the methods that only call `sendJsonRpc` and pass along the parsed response. Typically tests of this nature would be mocked to return known sets of values from RPC to make it a known quantity for any given test. That said this applies to all methods returning `sendJsonRpc` so introducing mocked methods and data doesn't need to be addressed by this PR.

What would you think about repurposing this as a `cookbook` example instead? It's a great demonstration of how to use this method and interpret the logic behind what it's returning.

packages/near-api-js/test/providers.test.js
```diff
@@ -295,6 +295,41 @@ test('json rpc light client proof', async() => {
...
+        // Not enough blocks to query one from last epoch, query current and assert that it returns
+        // empty. This may have a race condition where the query overlaps a new block being produced
+        let res = await provider.nextLightClientBlock({ last_block_hash: stat.sync_info.latest_block_hash });
+        if (res) {
+            // It's possible that the first epoch ended during this test (edge case),
+            // ensure that requesting with this new epoch will return the empty result
+            res = await provider.nextLightClientBlock({ last_block_hash: res.next_block_inner_hash });
+        }
+        expect(res).toStrictEqual({});
```
andy-haynes, line 314: I'm a little confused here, is the intention always to throw away the first result else fail the test?

packages/near-api-js/test/providers.test.js
```diff
@@ -295,6 +295,41 @@ test('json rpc light client proof', async() => {
     await expect(provider.lightClientProof(lightClientRequest)).rejects.toThrow(/.+ block .+ is ahead of head block .+/);
 });
 
+test('json rpc get next light client block', withProvider(async (provider) => {
```
andy-haynes, line 298: New tests for this should go in `@near-js/providers` as `near-api-js/tests` will be removed soon. This isn't documented anywhere but these tests only remain to ensure that the migration to packages did not break any test coverage.

andy-haynes (changes requested):
Thanks for the submission! The method looks good but I have some concerns around the integration test.

Also please run `pnpm changeset` in the monorepo and follow the prompts to generate a `changeset` JSON file with minor updates for `@near-js/providers` and `@near-js/types` (I didn't realize this wasn't part of the PR template but it will be added soon). This will also be necessary for #1105 and #1106

packages/near-api-js/test/providers.test.js
austinabell: > I'm a little wary of adding a test like this, branching execution paths based on state extrinsic to the test environment means that for any given test run we are only running an unknown subset of these paths. We should really only execute known code paths to ensure identical, deterministic coverage on every run.

Yeah, I wasn't happy with that setup, but I don't think there is a reasonable way to test this code path for _all_ networks. I mean, how tests could be (and maybe should be to be more deterministic) is starting a new network, running the tests, and discarding the data. I could just remove this section, and I believe it passes for all configs this repo uses (testnet, mainnet), but adding different networks for tests like the one mentioned would just cause issues for someone else.

I could just remove this section and return early, essentially skipping the test if we cannot retrieve a block from last epoch, but I figured this would just test to make sure that's the case.
 
> Even as an integration test I'm not convinced there's lots of value to testing the methods that only call `sendJsonRpc` and pass along the parsed response. Typically tests of this nature would be mocked to return known sets of values from RPC to make it a known quantity for any given test. That said this applies to all methods returning `sendJsonRpc` so introducing mocked methods and data doesn't need to be addressed by this PR.

Yeah, unless the provider is mocked, there isn't a way to have a test vector for this, since it's for testnet and mainnet? And if it is mocked, you'd lose the test verifying that the RPC interface matches this client.

> What would you think about repurposing this as a `cookbook` example instead? It's a great demonstration of how to use this method and interpret the logic behind what it's returning.

I don't think this specifically is a good use case since you'd never be doing this specific thing. I'm very open to doing a cookbook example when the validation is done for that flow, but this is not useful on its own.

Happy to just remove this test if that's preferred

packages/near-api-js/test/providers.test.js
austinabell: no, if the result returns `{}` (meaning there is not a newer block produced while the test is running), it doesn't send another request. This is just to ensure there isn't the race condition of a block being produced and indexed since the test started.

Definitely thinking about it more, there still is a race condition with this test, so maybe it's just better to remove checking that it's at the current head. I just liked the idea of verifying that it's at the head, which is why that code path is used.

austinabell:
Honestly, I don't have an opinion on how or if this is tested, just let me know which direction is preferred, and I'll make the changes. I only don't have the bandwidth to improve the testing framework outside of this test, any other change I'm happy to make.

I'll also move the tests to the `providers` package if tests are kept.

andy-haynes:
> Honestly, I don't have an opinion on how or if this is tested, just let me know which direction is preferred, and I'll make the changes.

For something a new method that's only calling `sendJsonRpc` with a new method I'd prefer to omit them in the interest of keeping this light and getting the functionality out. This is just an interesting case of test dependencies that I'm sure will come up again.

> Also, what is the semver rules for this repo? Does this constitute a minor version bump or patch? My assumption would be minor, but seems patch is more commonly done.

I don't know that there's been any discussion on it but I think minor makes the most sense here for a new method.

> Would there be conflicts with the other PRs, since I assume there would only be one update for the batch of changes, right?

There shouldn't be any conflicts, the changset files accumulate individually until publish time at which point they are collated with only the necessary version bumps.

packages/near-api-js/test/providers.test.js
andy-haynes: > I don't think there is a reasonable way to test this code path for all networks

I think you're probably right 😄 

> And if it is mocked, you'd lose the test verifying that the RPC interface matches this client.

Off the top of my head something that mocks `sendJsonRpc` and returns actual RPC responses (i.e. queried from RPC and saved in source as test assets) provides a reasonable canary in the short term, it's just difficult to keep those assets from stagnating as new changes modify them.

> I'm very open to doing a cookbook example when the validation is done for that flow, but this is not useful on its own.

👍 that would be awesome

austinabell:
> but I think minor makes the most sense here for a new method.

The providers and types packages were on version 0.0.2, so I assumed to just increase a patch version, and increased NAJ by a minor version. Let me know if you want all a minor bump.

I moved the test and removed that conditional testing, not 100% sure that's what you were suggesting, but let me know if [`9b7de98` (#1103)](https://github.com/near/near-api-js/pull/1103/commits/9b7de9843899629d45466ed4ff7cbf01eb359126) is not what you had in mind :)

austinabell:
> > I moved the test and removed that conditional testing
> 
> @austinabell Will the test always pass without the conditional (assuming it can reach RPC)? If so then I'm happy with it, if not I think it should be removed.

It will always pass with the current config and networks. It only will fail if there haven't been a epochs worth of blocks produced on the network. I added the note in the test so it's easier to debug if there are any large changes to the testing framework in the future

andy-haynes (approved):

austinabell:
Sorry for invalidating your approval. I realized I could improve the test a bit and pushed the changes before I saw the approval

andy-haynes (approved):

Issue #1095, opened 2023-03-31
Precision loss with light client timestamps


austinabell:
### Prerequisites

- [X] I'm using the latest version of `near-api-js`.
- [X] I have tried to start with a fresh project and reproduce the defect with minimal code changes.
- [x] I have read the console error messages carefully (if applicable).

### Description

The type used for timestamps in light client view has precision loss. The type used is a number:
https://github.com/near/near-api-js/blob/75d1e19b0135c990a91f58524d3bb24502f4bf8e/packages/types/src/provider/protocol.ts#L179

When the timestamps are in nanos, the type should be u64, as the values go above the max JS safe integer of `2^53 – 1`. For example, pulled a random block at height `80884288` and the timestamp is `1671316629471912329`, which, if you parse that into a JS number, it gives `1671316629471912400`.

Edit: Noticed that there is also another field `timestamp_nanos` on the RPC JSON response, which is the same number, but serialized as a string. This seems unnecessary to have this redundancy, because you should be able to deserialize a JSON integer into a big number in JS, but maybe a temporary solution is just to include this field in `BlockHeaderInnerLiteView`. I don't personally like that compromise, though.

While this is being changed, might want to change the `height` field also to a u64 type. This won't lose precision anytime soon, but if you do something like fuzz this API, it will break when the height is `> 2^53-1`.

Unclear what would be the ideal type for this lib, it doesn't seem like there is a type used for parsing from a JSON int without losing precision. Would love input on what maintainers would be willing for this to be changed to.

### Steps to reproduce

pull any block and the parsed `BlockHeaderInnerLiteView` timestamp should not match the original JSON. There is a probabilistic chance that these are equal, but only if the ts is a number that can be represented by JS floating point number.

### Expected behavior

This type should be u64 to not lose precision

### Actual behavior

Value loses precision

### Your environment

- NEAR JavaScript API version used:
- Frontend framework (if applicable):
- Relevant dependencies (if applicable):

### Self-service

- [X] I'd be willing to fix this bug myself.

austinabell:
Closing because #1106 is sufficient

PR #1106, opened 2023-04-07
fix: add missing field to light client header

austinabell:
## Pre-flight checklist

  - Issue #1095, no response from maintainers so no approval on working plan

## Motivation

I don't like this pattern of having a duplicate field for the same thing, but this type matches nearcore now, and these view types seem purely cosmetic in this lib where custom parsing wouldn't really fit (especially for this very simple change). Proposing this PR as it's the quickest workaround for #1095, but I don't feel strongly about any API change to resolve it if maintainers indicate they'd prefer something else.

## Test Plan

Added test for this field's existence, verified schema against nearcore source, checked json response from RPC for this method

## Related issues/PRs

Mentioned above


austinabell:
This test only exists in the NAJ package, and not in the providers test file. Do you want me to move it in this PR or will that be done in a separate PR when the duplicate files are deleted? @andy-haynes

andy-haynes:
> This test I modified only exists in the NAJ package, and not in the providers test file. Do you want me to move it in this PR or will that be done in a separate PR when the duplicate files are deleted? @andy-haynes

@austinabell the `json rpc light client proof` test is in `packages/accounts/test/providers.test.js` at the moment (as it relies on the `Account` class), I don't have a strong opinion whether it stays in `packages/near-api-js` but it should be updated in `accounts` please

andy-haynes (approved):

PR #1113, opened 2023-04-23
fix: add missing fields to execution outcome and test

austinabell:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/near/near-api-js/blob/master/CONTRIBUTING.md).
- [x] Commit messages follow the [conventional commits](https://www.conventionalcommits.org/) spec
- [x] **If this is a code change**: I have written unit tests.
- [x] **If this changes code in a published package**: I have run `pnpm changeset` to create a `changeset` JSON document appropriate for this change.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

This RPC type is missing fields that are required for proof verification. There is another `metadata` field and `compute_usage` fields that are missing also, but I opted to keeping these changes light.

I defaulted to having this be a patch changeset, but can change to minor if intended. Unclear how semver is with TS adding fields to interfaces (esp in this case since those types shouldn't be constructed manually)

## Test Plan

updated an existing test

PR #1116, opened 2023-04-25
feat: implement light client verification methods

austinabell:
## Pre-flight checklist

  - Going to wait for input on where this should live in the monorepo

## Motivation

Part of the implementation for https://near.social/#/devgovgigs.near/widget/gigs-board.pages.Post?id=148

This enables NAJ to verify light client blocks and execution proofs. The API is designed to be low-level and unopinionated. There can definitely be higher-level APIs built on this to automatically sync and verify transactions or just be more ergonomic, but anything I could think of came with tradeoffs. This might give some context as to why the API surfaces (specifically `validateExecutionProof`) are the way they are -- for flexibility and extensibility.

Open questions/comments:
- Very unclear where this should live in the monorepo. I just put it in the root level of the NAJ package for now, and hopefully, a maintainer can indicate where they want something like this. I tried to follow patterns within the repo. Waiting to do changeset until this is decided
  - The block verification test had to be duplicated to NAJ to test this verification. If API is kept as is, should the test just exist in NAJ or is the duplication fine?
- Test framework will be implemented here https://github.com/austinabell/near-light-client-tests. There are some scripts in that repo that are testing the logic of these functions more in-depth now, but there will be test vectors added in the future.
- The block verification fails for certain blocks below the 55mil range, due to a change in the protocol at that time. Unclear if there can be or is intended to be backwards-compatibility support for this, and unclear what that would look like as it's not documented. Context: https://near.zulipchat.com/#narrow/stream/295558-pagoda.2Fcore/topic/Light.20client.20block.20queries.20failing/near/350605623
- borsh-js sucks to work with, kept all types private and did not expose anything to the public API. Was not 100% sure what to do for the borsh schema, as it couldn't be combined with the one from the `transactions` package, so the `PublicKey` schema type is duplicated in the changes. There are also a bunch of types added that might seem unnecessary, but it's just because the borsh-js API is very limiting
- Some types are cast as `any` until https://github.com/near/near-api-js/pull/1113 comes in
- `validateLightClientBlock` taking last known block lite view and block producers separately as params is intentional. These might be kept separately, and the RPC marks returning bp set as optional (although the current nearcore RPC never returns null)
- `validateExecutionProof` taking the `merkleRoot` as bytes is intentional. One might not be using the JS light client to pull the `header->merkleRoot` or just have this root encoded in a different way. Bytes is most unopinionated here, instead of having to know that it is a base58 encoded string from the RPC response. This also keeps the API more stable if the data is deserialized properly in the future.
- `computeBlockHash` could be a utility that is moved outside of this module, but might be confusing if it exists in another namespace. Possibly re-export it from utilities/provider?

## Test Plan

Added tests here, building out a test framework for this now.

## Related issues/PRs

https://github.com/near/near-api-js/pull/1113 is the only open one relating to this. All others have been pulled in.


packages/near-api-js/src/light-client.ts
```diff
@@ -0,0 +1,532 @@
...
+ * producer set.
+ *
+ * @param lastKnownBlock The last light client block retrieved. This must be the block at the epoch before newBlock.
+ * @param currentBlockProducers The block producer set for the epoch of the last known block.
+ * @param newBlock The new block to validate.
+ */
+export function validateLightClientBlock(
+    lastKnownBlock: LightClientBlockLiteView,
+    currentBlockProducers: ValidatorStakeView[],
+    newBlock: NextLightClientBlockResponse
```
andy-haynes, line 265: For methods with multiple arguments please consider using named parameters to [avoid coupling to positional arguments](https://connascence.io/position.html), e.g.
```ts
validateLightClientBlock({ lastKnownBlock, currentBlockProducers, newBlock }: LightClientValidationOptions)
```

packages/near-api-js/test/providers.test.js
```diff
@@ -222,7 +223,7 @@ test('final tx result', async() => {
     expect(nearApi.providers.getTransactionLastResult(result)).toEqual({});
 });
 
-test('final tx result with null', async() => {
+test('final tx result with null', async () => {
```
andy-haynes, line 226: Please update this test in `packages/accounts/test/providers.test.js` and leave this one unchanged as duplicated tests are slated for removal.

andy-haynes:
> Very unclear where this should live in the monorepo.

> The API is designed to be low-level and unopinionated. There can definitely be higher-level APIs built on this

Sounds like a new package makes sense here, what do you think?

> couldn't be combined with the one from the transactions package, so the PublicKey schema type is duplicated in the changes

Why not add the new types to the existing serialization map? I'll concede it's not a great fit in `transactions` but the goal is to eventually get serialization into a lower-level package.

> Test framework will be implemented here https://github.com/austinabell/near-light-client-tests

Can the tests not live in this repo?

> computeBlockHash could be a utility that is moved outside of this module, but might be confusing if it exists in another namespace. Possibly re-export it from utilities/provider?

Seems like it could just live in `utils` unless you see no compelling reason for it to be used outside the context of light clients. I would strongly prefer not to duplicate new functionality across multiple packages.

andy-haynes (changes requested):
Just a few comments inline around structure

packages/near-api-js/src/light-client.ts
austinabell: I was trying to avoid creating a new type for this, but sounds good, will do!

austinabell:
> Sounds like a new package makes sense here, what do you think?

I don't have an opinion here, but I can move it into it's own package! Just didn't want to add any package bloat if it was against the goals of the refactoring that's going on :)

> Why not add the new types to the existing serialization map? I'll concede it's not a great fit in `transactions` but the goal is to eventually get serialization into a lower-level package.

Because there is only one overlapping type, it wasn't clear that having the larger map would be a clear win. This type of refactor sounds like something best split from this PR. Do you mind if it's handled separately and I open an issue for this?

> Can the tests not live in this repo?

They could, but the scripts are meant to be more non-deterministic to sample the chain to give confidence in the impl, and the tests vectors and framework I'm just starting might add bloat to the repo. After I'm done with it, we can revisit and see if we want? It also shouldn't take too long, so could wait for that if it's a blocker.

> Seems like it could just live in `utils` unless you see no compelling reason for it to be used outside the context of light clients. I would strongly prefer not to duplicate new functionality across multiple packages.

I don't think there is because the param is just the response type for the light client header and data. I don't think it's even feasible to manually construct that. I could be missing a use case here though.

austinabell:
I made some of the changes. I'll make the following outstanding changes tomorrow (listing mainly for my own memory):
- Docs on verification steps
- Creating new types to replace positional arguments
- Move test changes to `packages/accounts/test/providers.test.js` (this one, the test case would have to be moved to this new package now, though. Correct me if wrong)

austinabell:
So the light client verification test was moved to the new `light-client` package, but the verification can't happen in the `accounts` package because the light client utilities aren't exposed from it. Do you have an opinion on what I should do here? The test itself uses components from the `account` crate and the `light-client` utils to be able to verify that the data returned from RPC is a valid response.

Alternatively, I could decouple this and just write a new test that checks a few pre-defined execution proofs, so it depends on what types of tests are preferred here.

andy-haynes:
> So the light client verification test was moved to the new light-client package, but the execution verification can't happen in the accounts package because the light client utilities aren't exposed from it. Do you have an opinion on what I should do here?

The way I've been handling this is to add `@near-js/light-client` as a dev dependency to `@near-js/accounts`, making the verification method accessible in `accounts` tests without `light-client` becoming a build dependency.

andy-haynes:
> I don't have an opinion here, but I can move it into it's own package! Just didn't want to add any package bloat if it was against the goals of the refactoring that's going on :)

Initially I was thinking it made sense in `providers`, but if it's intended as a low-level API to be built-upon I'm not convinced it makes sense to couple these packages (or maybe `providers` would reference this as a new package 🤔). On the other hand it may be bloat in practice if these are intended only as methods for working with RPC responses. I imagine there are others who would want to weigh in before we commit to it though, I'm not entirely sure who will own this process in the future.

> Because there is only one overlapping type, it wasn't clear that having the larger map would be a clear win. This type of refactor sounds like something best split from this PR. Do you mind if it's handled separately and I open an issue for this?

I see, I missed that `@near-js/transactions` wasn't already a dependency here. I'm OK with the overlap in the interim, this is good motivation to get serialization decoupled from `@near-js/transactions`.

> After I'm done with it, we can revisit and see if we want?

Definitely! I was more curious whether you had anything in particular that couldn't be accommodated here.

packages/light-client/src/index.ts
```diff
@@ -0,0 +1,558 @@
...
+    NextLightClientBlockResponse,
+    ValidatorStakeView,
+} from '@near-js/types';
+import { Assignable } from '@near-js/types';
+import { PublicKey } from '@near-js/crypto';
+import BN from 'bn.js';
+import { serialize } from 'borsh';
+
+// TODO this abstract class exists in NAJ and seems unused. It is also copied in the transactions
+// TODO ..package. This should probably exist in utils and shared.
```
andy-haynes, line 20: Yup, now that borsh serialization is happening in multiple places it makes sense for this to be shared. Would you mind please moving this to `@near-js/types`?

packages/light-client/CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
+# @near-js/types
```
andy-haynes, line 1: This will be created automatically by `changeset`, it shouldn't be added manually 🙂

packages/light-client/src/index.ts
```diff
@@ -0,0 +1,558 @@
...
+            throw new Error('Enum can only take single value');
+        }
+        Object.keys(properties).map((key: string) => {
+            (this as any)[key] = properties[key];
+            this.enum = key;
+        });
+    }
+}
+
+// TODO: refactor this into separate files
```
andy-haynes, line 35: ❤️

packages/light-client/package.json
```diff
@@ -0,0 +1,37 @@
...
+    },
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
+    "dependencies": {
+        "bn.js": "5.2.1",
+        "@near-js/crypto": "workspace:*",
+        "@near-js/types": "workspace:*",
+        "@near-js/providers": "workspace:*",
+        "@near-js/utils": "workspace:*",
```
andy-haynes, line 23: This doesn't appear to be referenced

packages/light-client/package.json
```diff
@@ -0,0 +1,37 @@
...
+        "test": "jest test"
+    },
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
+    "dependencies": {
+        "bn.js": "5.2.1",
+        "@near-js/crypto": "workspace:*",
+        "@near-js/types": "workspace:*",
+        "@near-js/providers": "workspace:*",
```
andy-haynes, line 22: Since this is only referenced in a test it should be a dev dependency

austinabell:
Update on this: I added a runner in rainbow-bridge to validate the logic against the shared test suite here https://github.com/aurora-is-near/rainbow-bridge/pull/897. Hope that gives confidence in the implementation.

I can move the runner logic from https://github.com/austinabell/near-light-client-tests/blob/main/scripts/runExecutionTestVectors.ts, https://github.com/austinabell/near-light-client-tests/blob/main/scripts/runBlockTestVectors.ts into a test in this repo, similar to that PR, if preferred for more test coverage on this logic. If so, let me know how you'd like that set up :)

packages/light-client/src/block.ts
```diff
@@ -0,0 +1,165 @@
...
+            throw new Error('Next block producers hash doesn\'t match');
+        }
+    }
+}
+
+function hashBlockProducers(bps: ValidatorStakeView[]): Buffer {
+    const borshBps: BorshValidatorStakeView[] = bps.map((bp) => {
+        if (bp.validator_stake_struct_version) {
+            const version = parseInt(
+                bp.validator_stake_struct_version.slice(1)
```
robin-near, line 143: how about the more strict (and safer) check: `bp.validator_stake_struct_version === "V1"`

packages/light-client/src/borsh.ts
```diff
@@ -0,0 +1,173 @@
...
+        PublicKey,
+        {
+            kind: 'struct',
+            fields: [
+                ['keyType', 'u8'],
+                ['data', [32]],
+            ],
+        },
+    ],
+]);
```
robin-near, line 173: nit: consistency on newline after file

packages/light-client/test/config.js
```diff
@@ -0,0 +1,44 @@
+module.exports = function getConfig(env) {
```
robin-near, line 1: Is there a specific reason why we're writing tests in js and not ts?

packages/types/src/assignable.ts
```diff
@@ -5,3 +5,17 @@ export abstract class Assignable {
...
+
+export abstract class Enum {
+    enum: string;
+
+    constructor(properties: any) {
+        if (Object.keys(properties).length !== 1) {
+            throw new Error('Enum can only take single value');
+        }
+        Object.keys(properties).map((key: string) => {
+            (this as any)[key] = properties[key];
```
robin-near, line 17: nit: assert that key != "enum"?

packages/light-client/src/borsh.ts
```diff
@@ -0,0 +1,173 @@
...
+    v1?: BorshValidatorStakeViewV1;
+}
+
+export class BorshValidatorStakeViewWrapper extends Assignable {
+    bps: BorshValidatorStakeView[];
+}
+
+export class BorshEmpty extends Assignable {}
+
+export class BorshPartialExecutionStatus extends Enum {
```
robin-near, line 37: This is not a comment for this PR, but a general one for the borsh-js design. I think the design is rather inflexible as it requires that serializables are classes, and that makes it very unnatural that we have to use Enum and Assignable.

I'd imagine something like this is more reasonable:

```ts
export type BorshPartialExecutionStatus = (
  {unknown: {}} | {failure: {}} | {successValue: Uint8Array} | {successReceiptId: Uint8Array});

const SCHEMA_BorshEmpty: Schema<{}> = {
    kind: 'struct',
    fields: [],
};

const SCHEMA_BorshPartialExecutionStatus: Schema<BorshPartialExecutionStatus> = {
    kind: 'enum',
    field: 'enum',
    values: [
        ['unknown', SCHEMA_BorshEmpty],
        ['failure', SCHEMA_BorshEmpty],
        ['successValue', ['u8']],
        ['successReceiptId', [32]],
    ],
};

// to deserialize:
const executionStatus = deserialize(SCHEMA_BorshPartialExecutionStatus, buffer);
if ('successValue' in executionStatus) { useBuffer(executionStatus.successValue); }
// to serialize (no need to wrap):
const buffer = serialize(SCHEMA_BorshPartialExecutionStatus, {failure: {}});
```

But that requires changing borsh-js.

robin-near (approved):

packages/light-client/src/block.ts
```diff
@@ -0,0 +1,165 @@
...
+        newBlock.inner_lite.epoch_id !== lastKnownBlock.inner_lite.epoch_id &&
+        newBlock.inner_lite.epoch_id !== lastKnownBlock.inner_lite.next_epoch_id
+    ) {
+        throw new Error(
+            'New block must either be in the same epoch or the next epoch from the last known block'
+        );
+    }
+
+    const blockProducers: ValidatorStakeView[] = currentBlockProducers;
+    if (newBlock.approvals_after_next.length < blockProducers.length) {
```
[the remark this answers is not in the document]
austinabell: That implementation is wrong. The spec says, and in practice, there are blocks with more block producers than approvals because in the case of validator rotations, both the old and new bps must be included.

I can find and give a block where this is the case if needed, but I don't have an example off-hand.

packages/light-client/test/config.js
austinabell: no idea, was following the format of existing tests when I wrote

packages/light-client/src/borsh.ts
austinabell: Yeah, I agree. Was a large initiative and unrelated to my changes though, so I won't be addressing here.

packages/light-client/src/block.ts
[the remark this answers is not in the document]
austinabell: References, for posterity, to remove the trust aspect for others 😄 : https://github.com/near/NEPs/blob/master/specs/ChainSpec/LightClient.md#signature-verification. Also, line in rainbow bridge implementation, which is used in production https://github.com/aurora-is-near/rainbow-bridge/blob/33ca808b45cb5e9cf2e27f741b0f6e42d97c276b/contracts/eth/nearbridge/contracts/NearBridge.sol#L212 (python client just used for tests)

Also, anyone would be able to switch this to strict equal and see that it fails on recent blocks.

vikinatora:
Hey @austinabell, is this PR ready to be merged? As far as I understand most comments have been addressed and has went through the eyes of several people. If it's good I can resolve the merge conflicts and get it merged.

austinabell:
> Hey @austinabell, is this PR ready to be merged? As far as I understand most comments have been addressed and has went through the eyes of several people. If it's good I can resolve the merge conflicts and get it merged.

Hi @vikinatora, it's out of my control when and how this PR gets merged, but feel free to resolve merge conflicts if you'd like!

vikinatora:
@frol, I believe we can proceed with merging in this case. What do you think?

frol:
@vikinatora Let's get it merged and iterate from there 🙏

vikinatora:
@austinabell can I have write permission to your branch? I believe it's unnecessary to fork your fork and create a new PR just to resolve the conflicts. We'd make it harder to track the conversations here and context of the PR.

austinabell:
> @austinabell can I have write permission to your branch? I believe it's unnecessary to fork your fork and create a new PR just to resolve the conflicts. We'd make it harder to track the conversations here and context of the PR.

gave access! Let me know if you need anything else!

gagdiez:
This PR has been open for 2 years now, I doubt merging it would be trivial, if anyone still feels like working on it please reopen it

PR #1124, opened 2023-04-27
Restore deps key

andy-haynes:
## Pre-flight checklist

## Motivation

This PR enables the use of a `deps` argument to `NearConfig` instances. This restores legacy usage of passing in `deps.keyStore` instead of passing in `keyStore` as a top level argument to the configuration instance.

NB if both `keyStore` and `deps.keyStore` are passed, `keyStore` will continue to be used.

## Test Plan

The test plan involves updating to the latest `near-api-js` and verifying that signing still works when the `deps` parameter is used. Locally this was tested with `cookbook` examples, verifying that when `deps.keyStore` is used in place of `keyStore` there is no change in functionality.

### DIFF
--- a/.changeset/slimy-fans-tan.md
+++ b/.changeset/slimy-fans-tan.md
@@ -1,0 +1,6 @@
+---
+"near-api-js": patch
+"@near-js/wallet-account": patch
+---
+
+Allow use of legacy `deps.keyStore` path in NearConfig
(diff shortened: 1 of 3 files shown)


MaximusHaximus (approved):
:+1: Looks good @andy-haynes -- thanks for the speedy turnaround 💖

PR #1189, opened 2023-09-07
Update issue templates

trechriron:
Added two new templates as per the standard process.


trechriron:
I don't believe templates would impact any unit tests. This can merge once approved.

andy-haynes (approved):

Issue #1193, opened 2023-10-27
Global namespace collision for native node `fetch` when `near-api-js` is imported


andrew-scott-fischer:
### Prerequisites

- [X] I'm using the latest version of `near-api-js`.
- [X] I have tried to start with a fresh project and reproduce the defect with minimal code changes.
- [X] I have read the console error messages carefully (if applicable).

### Description

Attempting to use the node native `fetch` API in node >= 18 when the `near-api-js` is imported directly or by a transitive dependency results in the node global `fetch` property being overwritten by `near-api-js`.

These two lines of code [here](https://github.com/near/near-api-js/blob/53ba3f21c6da503e971a251b205bf50ea8b36dd0/packages/near-api-js/src/connect.ts#L31) and [here](https://github.com/near/near-api-js/blob/53ba3f21c6da503e971a251b205bf50ea8b36dd0/packages/providers/src/fetch_json.ts#L32) explicitly override the global `fetch` property.

They should not do this.

### Reproducible demo

This can be reporoduced in a node console:

```bash
$ node
Welcome to Node.js v18.17.1.
Type ".help" for more information.
> fetch
[AsyncFunction: fetch]
> require('near-api-js')
{
  keyStores: {
    KeyStore: [Getter],
    InMemoryKeyStore: [Getter],
    BrowserLocalStorageKeyStore: [Getter],
    UnencryptedFileSystemKeyStore: [Getter],
    MergeKeyStore: [Getter]
  },
  accountCreator: [Getter],
  providers: [Getter],
  utils: [Getter],
  transactions: [Getter],
  validators: [Getter],
  multisig: [Getter],
  Account: [Getter],
  Connection: [Getter],
  Contract: [Getter],
  InMemorySigner: [Getter],
  Signer: [Getter],
  KeyPair: [Getter],
  Near: [Getter],
  ConnectedWalletAccount: [Getter],
  WalletConnection: [Getter],
  connect: [Getter],
  DEFAULT_FUNCTION_CALL_GAS: [Getter]
}
> fetch
[Function: default_1]
```

### Steps to reproduce

1. open a project which has `near-api-js` somewhere in it's dependency tree
2. open a node console
3. type `fetch`
4. see output is `[AsyncFunction: fetch]`
5. type `require('near-api-js')`
6. type `fetch`
7. see output is `[Function: default_1]`

### Expected behavior

I would not expect a dependency to overwrite the global `fetch` property.

### Actual behavior

[...]

PR #1194, opened 2023-10-27
fix: remove global override of `fetch`

andrew-scott-fischer:
Feel free to help me get tests passing!

Ticket: #1193

## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

Attempting to use the node native fetch API in node >= 18 when the near-api-js is imported directly or by a transitive dependency results in the node global fetch property being overwritten by near-api-js.

These two lines of code [here](https://github.com/near/near-api-js/blob/53ba3f21c6da503e971a251b205bf50ea8b36dd0/packages/near-api-js/src/connect.ts#L31) and [here](https://github.com/near/near-api-js/blob/53ba3f21c6da503e971a251b205bf50ea8b36dd0/packages/providers/src/fetch_json.ts#L32) explicitly override the global fetch property.

They should not do this.

## Test Plan

I am not sure how to test this. Suggestions welcome

## Related issues/PRs

#1193

### DIFF
--- a/.changeset/wet-jars-attend.md
+++ b/.changeset/wet-jars-attend.md
@@ -1,0 +1,6 @@
+---
+"near-api-js": patch
+"@near-js/providers": patch
+---
+
+fixes override of global fetch property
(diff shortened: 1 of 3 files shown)


frol (approved):
This change makes total sense, thanks for addressing this cruel manipulation with the global object.

andrew-scott-fischer:
@frol I see the last release was ~5 months ago; any idea when the next one will be?

frol:
@andrew-scott-fischer I don't feel adventurous to do it before NEARCON, and then one week after NEARCON will be also a catch up game, so I would hope that we can have a release in the second half of November.

PR #1195, opened 2023-10-30
fix: add check for process global to exist

### LINKED ISSUE #737: process not defined in all environments
_**Describe the bug**_ I use NAJ in cloudflare workers and there is not `process` and of course no `process.env` so the following code throws errors:

[https://github.com/near/near-api-js/blob/689705926fe437b8d8671d72e77148228c895992/lib/utils/errors.js#L31](https://github.com/near/near-api-js/blob/689705926fe437b8d8671d72e77148228c895992/lib/utils/errors.js#L31)

I think this was introduced with the events that were recently added?

 _**To Reproduce**_ Steps to reproduce the behavior: 1\. Run signAndSendTransaction in any environment without process.env

 _**Expected behavior**_ It works

 _**Additional context**_ I recommend checking ALSO if `process` is defined first, then checking `process.env` so that envs without process defined can still use these methods.

ShaunSHamilton:
To prevent error 'process undefined' in environments where global process object does not exist

## Pre-flight checklist

## Motivation

Fixes issue whereby importing module in environments without a global `process` object would throw error `process undefined`.

## Test Plan

1) Wrote small Vitejs app with `"near-api-js": "."` dependency.
2) `vite` threw `process undefined` in browser console
3) Added change, `vite` no longer threw error in browser console

## Related issues/PRs

Closes #737

### DIFF
--- a/.changeset/funny-jars-camp.md
+++ b/.changeset/funny-jars-camp.md
@@ -1,0 +1,8 @@
+---
+"@near-js/accounts": patch
+"near-api-js": patch
+"@near-js/providers": patch
+"@near-js/utils": patch
+---
+
+add check for global 'process' object
--- a/packages/accounts/src/account.ts
+++ b/packages/accounts/src/account.ts
@@ -363,7 +363,7 @@ export class Account {
      * @param beneficiaryId The NEAR account that will receive the remaining Ⓝ balance from the account being deleted
      */
     async deleteAccount(beneficiaryId: string) {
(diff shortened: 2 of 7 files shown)


frol (approved):
Simple yet effective change. Thanks for contributing it!

frol:
I believe the CI error is not related to this change, and the root cause is the change in the protocol which does not allow top-level accounts creation anymore.

Issue #1200, opened 2023-11-09
Security: Supply chain protection is bad


paulmillr:
node_modules directory is 195.4 megabytes. All dependencies are using open-ended version ranges.

Any rogue dependency update (NPM acct of any of your dependency developer getting hacked) and near-api-js will spread malware and trojans, steal private keys. It's obvious 195 MBs of code has not been fully audited, so you don't even know what's there.

Steps required to fix:

1. Lock-down dependency versions, ensure the updates are rare
2. Reduce amount of dependencies. For example, bn.js, which is security-critical since it's used in crypto, can be completely removed.

These actions were done in ethereumjs libraries a long time ago.

gagdiez:
Hi! I automatically closed all issues pre-2025, will open this one again

paulmillr:
If it’s dep controlled by you it makes sense to keep the range. 

If it’s the dep by a well known org like react, it makes sense to keep the range.

If it’s the dep by an unknown person, it’s a trade off between ability to patch cves and supply chain attacks. Cves are 99% of time less severe than supply chain attacks which keep happening and draining user funds.

This particular nextjs vulnerability is one of its kind and was allowed to happen partially because of weird php-like nextjs/rsc architecture.

gagdiez:
ok, `near-api-js v7` has less dependencies, and they are all fixed versions:

https://github.com/near/near-api-js/blob/pre-release/v7/package.json#L43-L49

PR #1217, opened 2023-12-15
Pin exact dependency versions

gtsonevv:
## Pre-flight checklist

## Related issues/PRs

https://github.com/near/near-api-js/issues/1200


packages/biometric-ed25519/package.json
```diff
@@ -13,19 +13,19 @@
...
     "@near-js/crypto": "workspace:*",
     "@near-js/utils": "workspace:*",
-    "asn1-parser": "^1.1.8",
+    "asn1-parser": "1.1.8",
     "bn.js": "5.2.1",
     "borsh": "1.0.0",
-    "buffer": "^6.0.3",
-    "elliptic": "^6.5.4",
+    "buffer": "6.0.3",
+    "elliptic": "6.5.4",
```
paulmillr, line 24: why is elliptic used here in ed25519 context? elliptic's implementation of the curve is buggy and produces invalid outputs, which means in blockchain context "someone will lose money"

frol:
From the library usage perspective, it is not great to pin all the dependencies down to a single version as that may block deduplicating the dependencies if two libraries of slightly different versions are used (1.0.1 and 1.0.2). I am wondering if there is some middle ground that we can find here.

Also, this PR only resolves the first part of #1200, so it will be too early to close that issue without reviewing and removing buggy dependencies.

gtsonevv:
Hey guys, thank you for the replies. Based on them these are the next steps that we can take:
1. Our team will go through the dependency graph and resolve duplications.
2. We will try to reduce the bundle size. Looking at https://bundlephobia.com/package/near-api-js@2.1.4 we can see that 10% is bn.js which should be replaceable by native JS BigNumber. 7% is tweetnacl which is already replaced. We will try to find a comparable replacement for ajv because it’s biggest library by far(30%).

paulmillr:
others:

- js-sha256 => noble-hashes (already used)
- elliptic => noble-curves (already used)
- borsch deps
    - remove bn.js (native BigInt)
    - remove text-encoding-utf-8 (native TextEncoder)
    - bs58 should be upgraded upstream to remove safe-buffer, or replaced with scure-base
- re-evaluate need for node-fetch
- re-evaluate need for error-polyfill

frol (approved):
As suggested offline, let's merge this PR with the pinned dependencies, and then work on a separate PR that will slim down the dependencies.

PR #1223, opened 2023-12-21
Replace bn.js with BigInt

gtsonevv:
## Pre-flight checklist

## Motivation

Reduce amount of dependencies.

## Related issues/PRs

https://github.com/near/near-api-js/pull/1217
https://github.com/near/near-api-js/issues/1200


packages/accounts/src/account.ts
```diff
@@ -291,10 +290,10 @@ export class Account {
                 finality: 'optimistic'
             });
 
-            // store nonce as BN to preserve precision on big number
+            // store nonce as BigInt to preserve precision on big number
             const accessKey = {
                 ...rawAccessKey,
-                nonce: new BN(rawAccessKey.nonce),
+                nonce: BigInt(rawAccessKey.nonce || 0)
```
gagdiez, line 296: is this the intended behaviour? this will default to 0 if `nonce` is not defined, was that the case for `new BN(undefined)`? 

This is very relevant, specially since it is inside a `try/catch` statement

packages/utils/src/format.ts
```diff
@@ -9,12 +8,12 @@ export const NEAR_NOMINATION_EXP = 24;
 /**
  * Number of indivisible units in one NEAR. Derived from {@link NEAR_NOMINATION_EXP}.
  */
-export const NEAR_NOMINATION = new BN('10', 10).pow(new BN(NEAR_NOMINATION_EXP, 10));
+export const NEAR_NOMINATION = BigInt(10 ** NEAR_NOMINATION_EXP);
```
gagdiez, line 11: please check that this handle the right precision, before it was `Bigint(10)**BigInt(EXP)`

packages/utils/src/validators.ts
```diff
@@ -9,34 +8,34 @@ import depd from 'depd';
...
+    if (stakesSum < num) {
         throw new Error('Stakes are below seats');
     }
     // assert stakesSum >= numSeats
-    let left = new BN(1), right = stakesSum.add(new BN(1));
-    while (!left.eq(right.sub(new BN(1)))) {
-        const mid = left.add(right).div(new BN(2));
+    let left = BigInt(1), right = stakesSum + BigInt(1);
+    while (left !== right - BigInt(1)) {
+        const mid = left + right / BigInt(2);
```
gagdiez, line 33: is this right? I would have assumed the operation to be `(left+right)/2`

packages/utils/src/validators.ts
```diff
@@ -9,34 +8,34 @@ import depd from 'depd';
...
     return findSeatPriceForProtocolAfter49(validators, maxNumberOfSeats, minimumStakeRatio);
 }
 
-function findSeatPriceForProtocolBefore49(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], numSeats: number): BN {
-    const stakes = validators.map(v => new BN(v.stake, 10)).sort((a, b) => a.cmp(b));
-    const num = new BN(numSeats);
-    const stakesSum = stakes.reduce((a, b) => a.add(b));
-    if (stakesSum.lt(num)) {
+function findSeatPriceForProtocolBefore49(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], numSeats: number): bigint {
+    const stakes = validators.map(v => BigInt(v.stake)).sort((a, b) => (a < b ? -1 : a > b ? 1 : 0));
```
gagdiez, line 24: `BigInt` have order operators, is it necessary to provide a custom made one?

packages/utils/src/validators.ts
```diff
@@ -50,16 +49,16 @@ function findSeatPriceForProtocolBefore49(validators: (CurrentEpochValidatorInfo
...
 
 // nearcore reference: https://github.com/near/nearcore/blob/5a8ae263ec07930cd34d0dcf5bcee250c67c02aa/chain/epoch_manager/src/validator_selection.rs#L308;L315
-function findSeatPriceForProtocolAfter49(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], maxNumberOfSeats: number, minimumStakeRatio: number[]): BN {
+function findSeatPriceForProtocolAfter49(validators: (CurrentEpochValidatorInfo | NextEpochValidatorInfo)[], maxNumberOfSeats: number, minimumStakeRatio: number[]): bigint {
     if (minimumStakeRatio.length != 2) {
         throw Error('minimumStakeRatio should have 2 elements');
     }
-    const stakes = validators.map(v => new BN(v.stake, 10)).sort((a, b) => a.cmp(b));
-    const stakesSum = stakes.reduce((a, b) => a.add(b));
+    const stakes = validators.map(v => BigInt(v.stake)).sort((a, b) => (a < b ? -1 : a > b ? 1 : 0));
```
gagdiez, line 56: same as before

gagdiez (changes requested):
Please see all my comments above

packages/utils/src/validators.ts
gtsonevv: this is from the MDN docs.
```
mixed.sort(); // default sorting behavior
// [ -12n, 0, 0n, 10, 4n, 4, 6 ]

mixed.sort((a, b) => a - b);
// won't work since subtraction will not work with mixed types
// TypeError: can't convert BigInt value to Number value

// sort with an appropriate numeric comparator
mixed.sort((a, b) => (a < b ? -1 : a > b ? 1 : 0));
// [ -12n, 0, 0n, 4n, 4, 6, 10 ]
```

packages/utils/src/format.ts
gtsonevv: yes, should be `BigInt(10)**BigInt(EXP)` but we need to change the `target` option to `es2016` or later to make it work. Are you ok with that?

gagdiez: yes, since

`BigInt(10**24) == 999999999999999983222784n`

while

`BigInt(10)**BigInt(24) == 1000000000000000000000000n`

gagdiez:
How is that a test is not catching problems here? for example, the change of BN(10).exp(24) to BigInt(10**24) should have broke a bunch of stuff (since it is in the utils.js module, that every other module uses).

Wrongly changing it to `BigInt(10**24) == 999999999999999983222784n` should have broke a lot of tests... or is that we do not have tests for this?

packages/accounts/src/account.ts
gagdiez: I tested it myself, and indeed, `new BN(undefined) === BN(0)`

packages/utils/src/validators.ts
gagdiez: indeed, this is a bug

```js
const left = new BN(2)
const right = new BN(2)

const mid = left.add(right).div(new BN(2))
```

`mid === 2` while  `left + right / BigInt(2)  === 3n`

how is that this didn't break any test?

packages/accounts/src/account.ts
gtsonevv: will leave it `nonce: BigInt(rawAccessKey.nonce || 0)` because `BigInt` constructor doesn't work with `undefined`

packages/transactions/src/create_transaction.ts
```diff
@@ -1,9 +1,9 @@
...
-import BN from 'bn.js';
 
 import { Action } from './actions';
 import { Transaction } from './schema';
 
-export function createTransaction(signerId: string, publicKey: PublicKey, receiverId: string, nonce: BN | string | number, actions: Action[], blockHash: Uint8Array): Transaction {
-    return new Transaction({ signerId, publicKey, nonce: new BN(nonce), receiverId, actions, blockHash });
+export function createTransaction(signerId: string, publicKey: PublicKey, receiverId: string, nonce: bigint | string | number, actions: Action[], blockHash: Uint8Array): Transaction {
+    const bigIntNonce = typeof nonce === 'string' || typeof nonce === 'number' ? BigInt(nonce) : nonce;
+    return new Transaction({ signerId, publicKey, nonce: bigIntNonce, receiverId, actions, blockHash });
```
vikinatora, line 8: `bigIntNonce` is misleading since it can be of type `bigint | string | number`
```suggestion
    const txNonce = typeof nonce === 'string' || typeof nonce === 'number' ? BigInt(nonce) : nonce;
    return new Transaction({ signerId, publicKey, nonce: txNonce, receiverId, actions, blockHash });
```

vikinatora: Also I think it may be more readable and concise if you do the following
```suggestion
    const txNonce = typeof nonce === 'bigint' ? nonce :  BigInt(nonce);
    return new Transaction({ signerId, publicKey, nonce: txNonce, receiverId, actions, blockHash });
```

packages/utils/src/validators.ts
```diff
@@ -9,34 +8,34 @@ import depd from 'depd';
...
+    let left = BigInt(1), right = stakesSum + BigInt(1);
+    while (left !== right - BigInt(1)) {
+        const mid = (left + right) / BigInt(2);
         let found = false;
-        let currentSum = new BN(0);
+        let currentSum = BigInt(0);
         for (let i = 0; i < stakes.length; ++i) {
-            currentSum = currentSum.add(stakes[i].div(mid));
-            if (currentSum.gte(num)) {
+            currentSum = currentSum + stakes[i] / mid;
```
vikinatora, line 37: nitpick: better readablity
```suggestion
            currentSum = currentSum + (stakes[i] / mid);
```

packages/utils/src/validators.ts
vikinatora, line 56: Since ordering of bigints is used in two places already I think it's better to create a util file with a function that sorts it. It's will be much more understandable

vikinatora (approved):
Some nitpicks but looks very good

packages/utils/src/validators.ts
vikinatora, line 24: This is also a good candidate for a helper function.

vikinatora (dismissed):

gtsonevv:
Hey @gagdiez , this PR is ready for review. Merging is blocked until you approve it because you requested changes.

packages/utils/test/format.test.js
```diff
@@ -57,3 +57,7 @@ test('parseNearAmount fails when parsing values with ≥25 decimal places', () =
         'Cannot parse \'0.0000080990999998370878871\' as NEAR amount'
     );
 });
+
+test('NEAR_NOMINATION value', () => {
+    expect(NEAR_NOMINATION).toEqual(1000000000000000000000000n)
```
gagdiez, line 62: I see here we added the test for `NEAR_NOMINATION`, to make sure that the precision is right.

What puzzles me is that, just above this, there is a test checking that `formatNearAmount` is compatible with `parseNearAmount`. How is that it didn't fail when the bug of `BigInt(10**24)` was introduced?

They all need to be reviewed and, if necessary, more tests added. Otherwise, we cannot be sure we didn't added another bug to this cornerstone method.

gagdiez:
@gtsonevv 

I am still puzzled about the tests, and why they didn't fail. Specially on the `format` file.

We need to find an explanation, and add more tests if necessary.

If I am being picky with this PR is because a rounding error could lead to real people losing real money.

packages/utils/test/format.test.js
gagdiez: ok, that explains it

gagdiez (approved):

vikinatora:
Delaying the merge of this PR intentionally, because we want to batch these breaking changes with other oustanding breaking changes, in order to not increment major versions unnecessarily.

vikinatora (approved):

PR #1325, opened 2024-03-22
Extend Contract class to enable view function calls without Account

denbite:
## Pre-flight checklist

## Motivation

Many developers complained about the confusing implementation of the Contract class, which requires an Account parameter in the constructor even for view function calls

This PullRequest solves the issue by extending the interface of the first parameter in the constructor and additionally **enables full backward compatibility** with older versions

#### Old implementation
```js
// account is required
const contract = new Contract(account, options);

contract.view_method({});
contract.call_method({});
```

#### New implementation
```js
const contract = new Contract(connection, options);

// no account required
contract.view_method({});

// signerAccount must be specified to sign the transaction
contract.call_method({ signerAccount: account });
```

## Related issues/PRs

https://github.com/near/near-api-js/pull/1066
https://github.com/near/near-api-js/discussions/1064

### DIFF
--- a/packages/accounts/src/utils.ts
+++ b/packages/accounts/src/utils.ts
@@ -1,0 +1,105 @@
+import {
+    ViewStateResult,
+    BlockReference,
+    CodeResult,
+    PositionalArgsError,
+} from '@near-js/types';
+import { Connection } from './connection';
+import { printTxOutcomeLogs } from '@near-js/utils';
+import { ViewFunctionCallOptions } from './interface';
+
+function parseJsonFromRawResponse(response: Uint8Array): any {
+    return JSON.parse(Buffer.from(response).toString());
+}
+
+function bytesJsonStringify(input: any): Buffer {
+    return Buffer.from(JSON.stringify(input));
+}
+
+export function validateArgs(args: any) {
+    const isUint8Array = args.byteLength !== undefined && args.byteLength === args.length;
+    if (isUint8Array) {
+        return;
+    }
+
+    if (Array.isArray(args) || typeof args !== 'object') {
+        throw new PositionalArgsError();
+    }
+}
+
+export function encodeJSContractArgs(contractId: string, method: string, args) {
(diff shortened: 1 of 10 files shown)


packages/accounts/src/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (jsContract) {
+        encodedArgs = encodeJSContractArgs(contractId, methodName, Object.keys(args).length > 0 ? JSON.stringify(args) : '');
+    } else {
+        encodedArgs = stringify(args);
+    }
+
+    const result = await connection.provider.query<CodeResult>({
+        request_type: 'call_function',
+        ...blockQuery,
+        account_id: jsContract ? connection.jsvmAccountId : contractId,
```
vikinatora, line 95: Everything connected to `jsvmAccountId` is obsolete and will be removed very soon from near-api-js. Let's not add it to new functionality.

vikinatora (changes requested):
Looks very good, only one comment and I belive the biggest change I request is with the changeset, as there are quire a few breaking changes. Please adjust the changeset file acordingly. If you're not sure about how semantic versioning works, you can check the rules out [here](https://semver.org/#summary)

denbite:
> Looks very good, only one comment and I belive the biggest change I request is with the changeset, as there are quire a few breaking changes. Please adjust the changeset file accordingly(should be a major version increment most probably). If you're not sure about how semantic versioning works, you can check the rules out [here](https://semver.org/#summary)

These changes were made to be backward compatible, so releasing them to users shouldn't cause any issues (and it's not a breaking change from that perspective of view)

packages/accounts/src/utils.ts
denbite: The PR solves another issue and doesn't know anything about `jsvmAccountId`, this particular change has taken away function from the `Account` class and moved into another place
I wouldn't remove `jsvmAccountId` right here as it's a **breaking change** that may cause issues if you don't mind (and I assume you're preparing another PR out there that does remove it)

vikinatora (approved):

vikinatora (approved):

PR #1394, opened 2024-10-01
Switch to Yarn (~4.5.0) workspaces (with topological build sorting)

mikedotexe:
## Motivation

On on the LTS version of NodeJS, which is v20.17.0 at the time of this writing.

I had difficulty building the packages using the pnpm workspaces when trying to modify the `wallet-account` package. I resolved all the circular dependencies using a script located at the root. It's often difficult to build packages when they're interdependent, and modern Yarn workspaces has a "topological sort" that will run node commands in a wise way, for this very reason.

Screenshot 2024-09-30 at 10 07 24 PM

https://yarnpkg.com/cli/workspaces/foreach#details

 Wanted to get ESM and CommonJS working for all packages, where `near-api-js` is the only one that's CommonJS, and can use the CJS compilations from the other packages, which are modules.

It seems that "regular yarn" has workspaces that are too primitive. But after updating to 4.5.0 it seems more mature than the pnpm workspaces and whatnot. So I removed many things like turbo, husky, the pnpm mandate, and the iframe-rpc package, which use generics in a way that's not compatible with LTS and had a low amount of downloads/usage it seems prudent to move on. We can add back later if needed. I just need to be able to build all the packages so I can iterate on wallet selector problems.

A big part of the CJS issue was making sure file extensions are included, so you'll see a lot of adding `.js` to the end of imports. I consulted long and hard with AI and just did the tedious stuff until it got there.

## Test Plan

    nvm use --lts
    npm i -g yarn
    yarn set version berry
    yarn -v

You should see a version greater than 1, currently it's at 4.5.0.

    yarn
    yarn workspaces foreach --all -t run compile:cjs
    yarn workspaces foreach --all -t run compile:esm
    
Check for circular dependencies:

    ./check-circular-deps.sh
    
## Closing

I'm likely going to make this pull request a draft, and not be able to spend much energy shepherding it through. I mainly needed to get a way to build and iterate on the `wallet-account` package as I investigate something with attaching deposit to a function call. Might be a decent "reset" spot for folks who are interested in the potential streamlining Yarn workspaces might help with.

---

Note: you'll notice small inconsistencies that are temporary. like in the accounts package i was trying to see why the `dist` folder has multiple packages present, so the tsconfig setup is different and… kinda nonsensical, so pay no attn heheh


mpeterdev:
not a full PR review, but glanced at the description and thought I would chime in: topological build order should already be handled by turborepo as defined in the root `turbo.json`

https://turbo.build/repo/docs/crafting-your-repository/configuring-tasks#running-tasks-in-the-right-order

mikedotexe:
moved effort to https://github.com/mikedotexe/meer-api-js

PR #1421, opened 2024-10-30
chore: remove old codeowners

andy-haynes:
Out with the old, in with the new


r-near:
Added @frol and @r-near in #1460

Issue #1442, opened 2024-12-19
Wrong payload being signed


gagdiez:
The current implementation of `SignTransaction` signs the `transaction.encode()`, which is the borsh serialized version of the `transaction`:

https://github.com/near/near-api-js/blob/9cb7e89a688304fdd439411e2854235c358f4ab7/packages/client/src/transactions/sign_and_send.ts#L18-L25

In reality, we need to sign the `sha256` hash of the `borsh` serialized transaction:

```js
const encodedTx = new Uint8Array(sha256.sha256.array(transaction.encode()));
```

AlexKushnir1:
This means that signing is not working correctly. I don't fully understand how it worked beforehand and signed transactions were passed successfully and didn`t throw any errors?

gagdiez:
@AlexKushnir1 `@near-js/client` is a relatively new package, which still has no proper testing

My guess is that nobody used it and detected it, or simply nobody reported the error

AlexKushnir1:
@gagdiez I opine it has sense to clarify that we expect hash of `encodedTx` in the `signMessage` https://github.com/near/near-api-js/blob/c49fd67535baa5163b29ca7a43f585bfcfbe16e6/packages/client/src/interfaces/dependencies.ts#L10C1-L14C1 I see it like rewrite `encodedTx` into `encodedTransactionHash` and describe that signMessage expects a hash of encoded transaction.  WDYT?

gagdiez:
I see that actually `signMessage` already applies the `sha256`, so maybe we need to clarify that on the `signMessage` doc String, and add a test to try all this actually works

AlexKushnir1:
I just thought, if we specify on the `signMessage` that it should get an encoded hash it will cause a misleading naming.

I tested this func and related and there came to a conclusion that is"nt working, after adding hashing it was signed correctly and passed on the chain. Regarding the tests I would describe an issue and work for all client package

AlexKushnir1:
I"m not sure I understand where is `singMessage` applies the `sha256`. I would be really grateful for clarifying it to me!

gagdiez:
The `inMemorySigner` - I think this is the only one being implemented - hashes the message before signing it:

```js
const hash = new Uint8Array(sha256(message)); 
```

https://github.com/near/near-api-js/blob/c49fd67535baa5163b29ca7a43f585bfcfbe16e6/packages/signers/src/in_memory_signer.ts#L65-L80

AlexKushnir1:
[...]

PR #1444, opened 2024-12-28
Fix:  Changed encodedTx 

StarGachinum:
Resolved issue [#1442](https://github.com/near/near-api-js/issues/1442) .
Changed  `const encodedTx = transaction.encode();` 
to  `const encodedTx = new Uint8Array(sha256.sha256.array(transaction.encode()));`


gagdiez:
I see that actually signMessage already applies the sha256, so maybe instead of changing the payload being signed, we just need to clarify what signMessage does on its String, and add a test

simply changing this code without a proper test would lead to more errors down the line

gagdiez (changes requested):
I see that actually signMessage already applies the sha256, so maybe instead of changing the payload being signed, we just need to clarify what signMessage does on its String, and add a test

simply changing this code without a proper test would lead to more errors down the line

PR #1449, opened 2025-01-14
Extend wallet-account to support signAndSendTransactionAsync, signTransaction, sendTransaction & signDelegateAction

gtsonevv:
## Pre-flight checklist


packages/providers/src/json-rpc-provider.ts
```diff
@@ -131,7 +131,7 @@ export class JsonRpcProvider extends Provider {
      * @returns {Promise<FinalExecutionOutcome>}
      */
     async sendTransactionAsync(signedTransaction: SignedTransaction): Promise<FinalExecutionOutcome> {
-        return this.sendTransactionUntil(signedTransaction, 'NONE');
+        return this.sendTransactionUntil(signedTransaction, 'INCLUDED');
```
gagdiez, line 134: I think we can leave it as `NONE`, since anyway people will have to check its status, and that way we don't lose time on the overhead of waiting for the RPC to answer

packages/wallet-account/src/wallet_account.ts
```diff
@@ -451,4 +451,193 @@ export class ConnectedWalletAccount extends Account {
...
+            }
+        }
+
+        const block = await this.connection.provider.block({ finality: 'final' });
+        const blockHash = baseDecode(block.header.hash);
+        const publicKey = PublicKey.from(accessKey.public_key);
+        const nonce = Number(accessKey.access_key.nonce) + 1;
+        const transaction = createTransaction(this.accountId, publicKey, receiverId, nonce, actions, blockHash);
+        
+        this.walletConnection.requestSignTransactions({
```
gagdiez, line 488: don't we need to be async as well? otherwise I think you are only making it async if there is a function call key available

packages/wallet-account/src/wallet_account.ts
```diff
@@ -451,4 +451,193 @@ export class ConnectedWalletAccount extends Account {
...
+            }
+        }
+
+        const block = await this.connection.provider.block({ finality: 'final' });
+        const blockHash = baseDecode(block.header.hash);
+        const publicKey = PublicKey.from(accessKey.public_key);
+        const nonce = Number(accessKey.access_key.nonce) + 1;
+        const transaction = createTransaction(this.accountId, publicKey, receiverId, nonce, actions, blockHash);
+
+        this.walletConnection.requestSignTransactions({
```
gagdiez, line 631: is this command valid for signing delegate transactions?

gagdiez:
walletConnection is no longer existing, for which I will proceed to close this PR

PR #1492, opened 2025-03-13
fix: "Cannot mix BigInt and other types" for signing transaction with explicitly provided nonce

frol:
## Pre-flight checklist

### DIFF
--- a/.changeset/weak-adults-love.md
+++ b/.changeset/weak-adults-love.md
@@ -1,0 +1,5 @@
+---
+"@near-js/client": patch
+---
+
+Fixed "Cannot mix BigInt and other types" error when signing transactions with explicitly provided nonce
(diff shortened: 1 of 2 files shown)


BenKurrek (approved):

frol:
@gagdiez May I ask for your review? We have discovered this issue today with @BenKurrek since he was manually controlling nonce and NAJ was crashing his app since `undefined += 1n` did not work.

gagdiez:
why not just initializing the nonce as `0n`? I'm quite sure that leaving it as `undefined` will break something further down the line

What is the `issue` / `code` that originated this PR?

frol:
@gagdiez I considered that, but initializing it with 0 will create other bugs - if you first sign a transaction with explicit nonce and next time without explicit nonce, it will use nonce=1, which is wrong. Also, just on the conceptual level it is not a right move to initialize it with invalid value.

Regarding you concern that this fix might bite us elsewhere, I don’t believe so as this value is tied to this internal state and not used directly anywhere else.

gagdiez (approved):

gagdiez:
I have no better solution for now, so I will not block this, approved

PR #1589, opened 2025-06-19
refactor!: expose only `signBytes` for implementation 

AlexKushnir1:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation
Provide abstract class `Signer` that has only two methods to implement `signBytes` and `getPublickKey`, while implementing many signature methods.  That way, the flow stays consistent, and users only focus on the actual signing step

## Test Plan

Added unit tests to verify the correctness of the NEP-413 payload hash generation and signature logic:
- Hash Generation Test
- Signature Verification Test

### DIFF
--- a/.changeset/wild-adults-grow.md
+++ b/.changeset/wild-adults-grow.md
@@ -1,7 +1,5 @@
 ---
-"@near-js/signers": patch
+"near-api-js": major
 ---
 
-Refactored signer: move all logic into abstract signer, and expose only protected 'signBytes' that need to be implemented
-    
-Add unit tests to verify nep413MessageSignature 
+Refactor the abstract class `Signer` to implement every method, except for `getPublicKey` and `signBytes`, which are intended for users


denbite:
@AlexKushnir1 I agree with the idea that end users shouldn't handle signature verification themselves, yet the current approach still leaves room for mistakes, like forgetting to call `getPayloadHashForNEP413` inside `signNep413Message`

one alternative could be to move all the logic into the abstract `Signer` class and expose just a protected `signMessagePayload(bytes: Uint8Array)` method for end users to implement - that way, the flow stays consistent, and users only focus on the actual signing step

packages/signers/src/signer.ts
```diff
@@ -32,33 +37,101 @@ export const Nep413MessageSchema: Schema = {
...
+        params: SignMessageParams
+    ): Promise<SignedMessage> {
+        const hash = getPayloadHashForNEP413(params);
+
+        const signature = await this.signBytes(hash);
+        return {
+            accountId: accountId,
+            publicKey: await this.getPublicKey(),
+            signature: signature,
+            state: params.callbackUrl,
```
AlexKushnir1, line 77: @denbite I add state as param.callBackUrl, is that expected logic?

.changeset/wild-adults-grow.md
```diff
@@ -0,0 +1,10 @@
+---
+"@near-js/signers": patch
```
denbite, line 2: it's not a `patch` anymore as it's breaking the existing interface for users that already implemented their own `Signer` and adding a new `protected function signBytes(...)`, imo it's a `major` update

.changeset/wild-adults-grow.md
```diff
@@ -0,0 +1,10 @@
...
+---
+
+Separate NEP-413 hash logic and add related unit tests
```
denbite, line 5: let's update the description according to the latest changes, so it explains what's actually been done here

packages/signers/src/key_pair_signer.ts
```diff
@@ -1,22 +1,6 @@
...
-    Signature,
-    DelegateAction,
-    SignedDelegate,
-    encodeDelegateAction,
-} from '@near-js/transactions';
-import { serialize } from 'borsh';
+import { Signer } from './signer';
 
 /**
  * Signs using in memory key store.
```
denbite, line 6: let's fix this comment, just noticed that we must have been forgotten to change it

packages/signers/src/signer.ts
denbite: according to the [NEP doc](https://github.com/near/NEPs/blob/master/neps/nep-0413.md#returning-the-signature), a browser wallet should pass the `state` parameter as a query parameter in the `callbackUrl`, so we need to parse it out of `callbackUrl` and return it to a user if present

packages/signers/src/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param message String message to be signed or verified.
+ * @param recipient Account ID of the intended recipient of the message.
+ * @param nonce A nonce that uniquely identifies this instance (32 bytes).
+ * @param callbackUrl Optional URL that calls after signing.
+ *
+ * @returns A SHA-256 hash of the serialized payload as a `Uint8Array`.
+ */
+
+export function getPayloadHashForNEP413(params: SignMessageParams): Uint8Array {
```
denbite, line 21: do you think we still need this helper, considering it will be used only once?

denbite (changes requested):

packages/signers/src/signer.ts
AlexKushnir1: IMO it works as it should now.
[Previosly](https://github.com/near/near-api-js/blob/386a5c0715e445e1f4cddc840f24cf2a408b6cbc/packages/signers/src/key_pair_signer.ts#L42) the `state` was just ignored:

```ts
return {
    accountId: accountId,
    publicKey: pk,
    signature: signature,
};
```

PR #1605, opened 2025-07-15
Adds verifyNep413Message to _____ package [don't merge]

elliotBraem:
## Pre-flight checklist

## Motivation

Moved over some things, started an example, started putting it in utils but I don't think it should go here in order to avoid circular dependencies and to utilize some shared types w/ existing signNep413Message.

I'm thinking of moving it to a verify.ts in `signer`, or maybe to `crypto` -- but creating a new `@near-js/messages` package sounds pretty clean and would follow `transactions` pattern.  

## Test Plan

I'll write some tests and an example.

## Related issues/PRs

This was requested to accompany `signNep413Message`. I'm pulling logic from https://github.com/elliotBraem/near-sign-verify

### DIFF
--- a/packages/utils/src/verify-message.ts
+++ b/packages/utils/src/verify-message.ts
@@ -1,0 +1,68 @@
+import {
+    SignedMessage,
+} from '@near-js/signer'; // shared type maybe?
+
+type SignedMessage {
+    accountId: string; // The account name to which the publicKey corresponds as plain text (e.g. "alice.near")
+    publicKey: PublicKey; // The public counterpart of the key used to sign, expressed as a string with format "<key-type>:<base58-key-bytes>" (e.g. "ed25519:6TupyNrcHGTt5XRLmHTc2KGaiSbjhQi1KHtCXTgbcr4Y")
+    signature: Uint8Array; // The base64 representation of the signature.
+    state?: string;
+}
+
+export function verifyNep413Message({ accountId, publicKey, signature, state }: SignedMessage) {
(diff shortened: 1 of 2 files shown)


packages/utils/src/verify-message.ts
```diff
@@ -0,0 +1,68 @@
+import {
+    SignedMessage,
```
elliotBraem, line 2: I need this to be a shared type, could put it in `types`

`SignedMessage` + `Nep413Message` or just `Message`

packages/utils/src/verify-message.ts
```diff
@@ -0,0 +1,68 @@
...
+    try { // I do think we should throw, because there may be an unsupported keypair type
+        // return the verified payload on success, { success: true, payload: verifiedPayload }
+
+
+        verifySignature(payloadHash, signatureBytes, publicKey);
+    } catch (error) {
+        throw new Error(
+            `Cryptographic signature verification failed: ${error instanceof Error ? error.message : String(error)
+            }`,
+        );
```
elliotBraem, line 44: near-sign-verify throws on failed verify, I think we should do this here too rather than returning a boolean because it forces the developer to handle it

denbite: completely agree

packages/utils/src/verify-message.ts
gagdiez: +1 for putting shared types on the `types` package

Issue #1610, opened 2025-07-22
ClassicActions borsh schema tags are incorrect for global contract actions


pugachAG:
### Prerequisites

- [x] I'm using the latest version of `near-api-js`.
- [x] I have tried to start with a fresh project and reproduce the defect with minimal code changes.
- [x] I have read the console error messages carefully (if applicable).

### Description

`ClassicActions` borsh schema skips `Delegate` action which changes enum tag for `DeployGlobalContract` and `UseGlobalContract` to be 8 and 9 instead of 9 and 10.
Note that in nearcore `NonDelegateAction`  uses the same borsch schema as regular `Action` so I suggest removing separate schema for `ClassicActions` and use `Action` schema instead to be consistent with nearcore.

### Steps to reproduce

1. Create signed delegate with `UseGlobalContract` action and encode it to base64: 
```
const [, signedDelegate] = await userAccount.createSignedMetaTransaction(
    userAccount.accountId,
    [
      actionCreators.useGlobalContract(
        new GlobalContractIdentifier({
          AccountId: GLOBAL_CONTRACT_ACCOUNT_ID,
        })
      ),
    ]
  );

  const signedDelegateBase64 = Buffer.from(
    serialize(SCHEMA.SignedDelegate, signedDelegate)
  ).toString("base64");
```

2. Run the following test as part of nearcore:
```
    #[test]
    fn test_signed_delegate_base64() {
        let s = "FwAAAGxhcmdlZXN0YXRlMjM0NC50ZXN0bmV0FwAAAGxhcmdlZXN0YXRlMjM0NC50ZXN0bmV0AQAAAAkBKQAAADIwMjUtMDctMDFfMTMtNTEtMTJfbmVhcl9yZWNvdmVyeS50ZXN0bmV0w5zgweW6AADVTUwMAAAAAABg3MUvOJRprJUvTHbfGH1/kOJixF1WYdMVQHc7URblQABDQ5/e5vQJoiepB2ICZjdU+1tBOuAp2PpXr/MZK8UemImvvj8mgHLag5j0swW1LDJD3+X94eqAz1KWL1UQoSwM";
        let bytes = from_base64(s).unwrap();
        let mut buf = &bytes[..];
        let sender = AccountId::deserialize_reader(&mut buf).unwrap();
        dbg!(sender);
        let receiver = AccountId::deserialize_reader(&mut buf).unwrap();
        dbg!(receiver);
[...]

PR #1611, opened 2025-07-23
Fix issue where meta transactions with `deployGlobalContract` or `useGlobalContract` actions aren't passing validation

denbite:
## Pre-flight checklist

## Motivation

Currently, `ClassicActions` can't be deleted (at least with the current architecture), because of two-way dependencies - `DelegateAction` schema is dependent on `Action` and `Action` is, in turn, dependent on `SignedDelegate` schema, which contains `DelegateAction`

## Test Plan

Added e2e tests to make sure it never happens again

## Related issues/PRs

fixes https://github.com/near/near-api-js/issues/1610 
related to https://github.com/near/devrel/issues/574

### DIFF
--- a/packages/transactions/src/schema.ts
+++ b/packages/transactions/src/schema.ts
@@ -246,7 +246,7 @@ export const SCHEMA = new class BorshSchema {
             { struct: { addKey: this.AddKey } },
             { struct: { deleteKey: this.DeleteKey } },
             { struct: { deleteAccount: this.DeleteAccount } },
-            { struct: { signedDelegate: 'u8' } }, // temp fix: this should never be used
+            { struct: { signedDelegate: 'string' } }, // placeholder to keep the right enum order, should not be used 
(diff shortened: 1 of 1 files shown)


gagdiez (approved):

gagdiez:
I think it is fair to add a bit of context on this issue and fix.

Let's start with the fact that`Borsh` encodes enums as `index(enum_structure) + borsh(enum_structure)`

Here in `naj` we divided between `Actions` and `ClassicActions`, which are exactly the same with an important difference:

`Actions` has an entry for `delegateAction` which take a `ClassicActions` as input -> this allowed us to break the circular dependency of the `Actions` scheme with itself

Now that the new `globalContracts` actions were added, they also needed to be added to the `ClassicActions`, but in the right ordering. This is, they should come after `signedDelegate`. However, we never expect people to use the `classicActions` to define `signedDelegate`, so we can simply add a placeholder to keep the `enum` indexing right.

Issue #1615, opened 2025-08-08
Add `decimal` parameter to token's `.toDecimal` method


gagdiez:
### Description
Balances are expressed in whole units, for example, one NEAR is `10000...0` (24 zeros) 

Right now we have two methods to deal with balances format:
- `toUnits(decimal)` that converts from a decimal string (e.g. `1.23`) to units (`12300...000`)
- `toDecimal(units)` that converts from units (e.g. `100...00`) to decimals (`1,000000000`)

The main inconvenience is that `toDecimal` will create strings with as many decimals as possible, so `9976300876322311` units will become the `string` `"0.9976300876322311"`

We need to add a parameter to `toDecimal(units, decimals)` so we can ask for `toDecimal("9976300876322311", "2")` and returns `0.99` (always rounding down)

AlexKushnir1:
@gagdiez Is attached PR resolve your issue?

PR #1618, opened 2025-08-19
feat: add optional `precision` param to control digits after decimal point

AlexKushnir1:
## Pre-flight checklist

## Test Plan
Default `toDemical` should keep the same logic. If user provide value he is responsible for result, like if value smaller that decimals after point it will return 0.

## Related issues/PRs
Close https://github.com/near/near-api-js/issues/1615

### DIFF
--- a/packages/tokens/src/ft/format.ts
+++ b/packages/tokens/src/ft/format.ts
@@ -16,7 +16,7 @@ export function formatAmount(
         .substring(0, fracDigits)
         .padStart(fracDigits, '0');
 
-    fractionStr = fractionStr.substring(0, Number(precision));
+    fractionStr = fractionStr.substring(0, precision);
 
     return trimTrailingZeroes(`${wholeStr}.${fractionStr}`);
 }
--- a/packages/tokens/src/ft/format.ts
+++ b/packages/tokens/src/ft/format.ts
@@ -6,7 +6,7 @@ export const MAX_NOMINATION_EXP = 24;
 export function formatAmount(
     units: string | number | bigint,
     fracDigits: number,
-    precision: bigint | string | number = fracDigits
+    precision: number = fracDigits
 ): string {
     units = cleanupUnits(units);
 
--- a/packages/tokens/src/ft/format.ts
+++ b/packages/tokens/src/ft/format.ts
@@ -5,15 +5,18 @@ export const MAX_NOMINATION_EXP = 24;
  */
 export function formatAmount(
     units: string | number | bigint,
-    fracDigits: number
+    fracDigits: number,
+    precision: bigint | string | number = fracDigits
 ): string {
     units = cleanupUnits(units);
 
     const wholeStr = units.substring(0, units.length - fracDigits) || '0';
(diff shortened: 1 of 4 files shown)


gagdiez:
nice! I would like to see a test with numbers bellow 0

packages/tokens/test/fungible_token.test.ts
```diff
@@ -67,18 +67,33 @@ test('test toDecimal formats units', () => {
     expect(FT.toDecimal('1000000')).toBe('1');
     expect(FT.toDecimal(1_000_000)).toBe('1');
     expect(FT.toDecimal(BigInt(1_000_000))).toBe('1');
+    expect(FT.toDecimal('1000000', 2)).toBe('1');
```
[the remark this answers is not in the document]
AlexKushnir1: That is because the function `formatAmount()` calls [trimTrailingZeroes()](https://github.com/near/near-api-js/blob/386a5c0715e445e1f4cddc840f24cf2a408b6cbc/packages/tokens/src/ft/format.ts#L18) on the result.
If the intention is to always keep the trailing zeros (e.g. `'1.00'`), then trimTrailingZeroes() should be skipped.
WDYT?

packages/tokens/test/fungible_token.test.ts
```diff
@@ -67,18 +67,34 @@ test('test toDecimal formats units', () => {
...
+    expect(FT.toDecimal(12_345_678, 2)).toBe('12.34');
+    expect(FT.toDecimal(BigInt(12_345_678), 2)).toBe('12.34');
+
+    expect(FT.toDecimal('710', '2')).toBe('0');
+    expect(FT.toDecimal(710, 0)).toBe('0');
+    expect(FT.toDecimal(BigInt(710), 2)).toBe('0');
+
+    expect(FT.toDecimal('0', 2)).toBe('0');
+    expect(FT.toDecimal(0, 2)).toBe('0');
+    expect(FT.toDecimal(BigInt(0), 2)).toBe('0');
```
gagdiez, line 97: we need tests to see if we can convert numbers to only decimals, e.g. 
- FT.toDecimal(123_456, '2') should be '0.12',
- FT.toDecimal(123_456, '4') should be '0.1234'
- FT.toDecimal(1999, '4') should be '0.001'
- FT.toDecimal(1239, '6') should be '0.00123'

packages/tokens/src/ft/format.ts
```diff
@@ -5,16 +5,19 @@ export const MAX_NOMINATION_EXP = 24;
...
     units = cleanupUnits(units);
 
     const wholeStr = units.substring(0, units.length - fracDigits) || '0';
-    const fractionStr = units
+    let fractionStr = units
         .substring(units.length - fracDigits)
         .substring(0, fracDigits)
         .padStart(fracDigits, '0');
 
+    fractionStr = fractionStr.substring(0, Number(precision));
```
gagdiez, line 19:
```suggestion
    fractionStr = fractionStr.substring(0, precision);
```

PR #1620, opened 2025-08-25
fix: use `optimistic` as default finality for RPC queries

denbite:
## Pre-flight checklist

## Motivation

This PR fixes a race condition that could happen when sending multiple consecutive transactions from the same account, signed by a single key, it addresses the problem by aligning the default `finality` for RPC queries and `waitUntil` status for transactions to `optimistic` execution, ensuring consistent behaviour

## Test Plan

It adds a test case to ensure this issue doesn't come back

### DIFF
--- a/packages/accounts/src/account.ts
+++ b/packages/accounts/src/account.ts
@@ -66,7 +66,7 @@ const {
 } = actionCreators;
 
 const DEFAULT_FINALITY: Finality =  "optimistic";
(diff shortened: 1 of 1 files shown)


gagdiez (changes requested):
I thought we were fixing "inconsistencies of different waiting on different functions", not that we were changing the default values

gagdiez (approved):

PR #1355, opened 2024-06-26
Add Secp256k1 support

gtsonevv:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/near/near-api-js/blob/master/CONTRIBUTING.md).
- [x] Commit messages follow the [conventional commits](https://www.conventionalcommits.org/) spec
- [x] **If this is a code change**: I have written unit tests.
- [x] **If this changes code in a published package**: I have run `pnpm changeset` to create a `changeset` JSON document appropriate for this change.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

Adds support for secp256k1 curve

## Test Plan

All tests pass

## Related issues/PRs

https://github.com/near/near-api-js/pull/985
https://github.com/near/near-api-js/issues/1039
https://github.com/near/near-api-js/issues/440

PR #1625, opened 2025-08-29
Fix: Ensure passkey-derived keys are prefixed with ed25519: and add regression test

SmolinPavel:
This PR addresses a bug in `@near-js/biometric-ed25519` where passkey-derived key strings were not prefixed with `ed25519:`, causing `KeyPair.fromString` to throw errors:

```
Invalid encoded key format, must be <curve>:<encoded key>
```

The change ensures that all passkey-derived keys are properly prefixed, and adds a regression test to prevent future regressions.

## Pre-flight checklist

## Motivation

The bug was caused by legacy behavior in `@near-js/biometric-ed25519` where KeyPair.fromString was sometimes called with a raw base-encoded secret+public key string.

Legacy behavior (pre-fix) allowed KeyPairEd25519 to be constructed from a single-part string:

```
if (parts.length === 1) {
    return new KeyPairEd25519(parts[0]);
}
```
See [pre-fix state](https://github.com/near/near-api-js/blame/b61b66d271146f0a1b5b88e75da06dffc354860e/packages/crypto/src/key_pair.ts#L24) for reference.

The update introduced in https://github.com/near/near-api-js/pull/1355 made param to be `KeyPairString`, assuming there is always a prefix.

The update introduced in this PR enforces the prefix explicitly, ensuring compatibility with `KeyPair.fromString` and making the behavior consistent and predictable.

## Test Plan

The regression tests were added, making sure that the call to `KeyPair.fromString` will fail without the ed25519 prefix, while the one with the prefix will work.

## Related issues/PRs

https://github.com/near/near-api-js/pull/1355

### DIFF
--- a/packages/biometric-ed25519/src/index.ts
+++ b/packages/biometric-ed25519/src/index.ts
@@ -87,7 +87,7 @@ export const createKey = async (username: string): Promise<KeyPair> => {
             const publicKeyBytes = get64BytePublicKeyFromPEM(publicKey);
             const secretKey = sha256.create().update(Buffer.from(publicKeyBytes)).digest();
             const pubKey = ed25519.getPublicKey(secretKey);
-            return KeyPair.fromString(baseEncode(new Uint8Array(Buffer.concat([Buffer.from(secretKey), Buffer.from(pubKey)]))) as KeyPairString);
+            return KeyPair.fromString(makeEd25519KeyString(secretKey, pubKey));
         });
 };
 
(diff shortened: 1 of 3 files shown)


packages/biometric-ed25519/src/index.ts
```diff
@@ -87,7 +87,7 @@ export const createKey = async (username: string): Promise<KeyPair> => {
             const publicKeyBytes = get64BytePublicKeyFromPEM(publicKey);
             const secretKey = sha256.create().update(Buffer.from(publicKeyBytes)).digest();
             const pubKey = ed25519.getPublicKey(secretKey);
-            return KeyPair.fromString(baseEncode(new Uint8Array(Buffer.concat([Buffer.from(secretKey), Buffer.from(pubKey)]))) as KeyPairString);
+            return KeyPair.fromString(makeEd25519KeyString(secretKey, pubKey));
```
denbite, line 90: let's change that to `new KeyPairEd25519`

packages/biometric-ed25519/src/index.ts
```diff
@@ -130,8 +130,8 @@ export const getKeys = async (username: string): Promise<[KeyPair, KeyPair]> =>
             const firstEDPublic = ed25519.getPublicKey(firstEDSecret);
             const secondEDSecret = sha256.create().update(Buffer.from(correctPKs[1])).digest();
             const secondEDPublic = ed25519.getPublicKey(secondEDSecret);
-            const firstKeyPair = KeyPair.fromString(baseEncode(new Uint8Array(Buffer.concat([Buffer.from(firstEDSecret), Buffer.from(firstEDPublic)]))) as KeyPairString);
-            const secondKeyPair = KeyPair.fromString(baseEncode(new Uint8Array(Buffer.concat([Buffer.from(secondEDSecret), Buffer.from(secondEDPublic)]))) as KeyPairString);
+            const firstKeyPair = KeyPair.fromString(makeEd25519KeyString(firstEDSecret, firstEDPublic));
+            const secondKeyPair = KeyPair.fromString(makeEd25519KeyString(secondEDSecret, secondEDPublic));
```
denbite, line 134: the same as above

denbite (changes requested):

PR #1648, opened 2025-10-26
feat(providers): allow `Uint8Array` as args input for `callFunctionRaw` and `callFunction`

denbite:
## Pre-flight checklist

## Motivation

Adds support for binary arguments in `callFunctionRaw`, allowing you to pass a `Uint8Array` directly instead of only JSON-serializable objects.

## Related issues/PRs

https://t.me/neardev/74289

### DIFF
--- a/packages/providers/src/json-rpc-provider.ts
+++ b/packages/providers/src/json-rpc-provider.ts
@@ -226,7 +226,7 @@ export class JsonRpcProvider implements Provider {
     public async callFunction<T extends SerializedReturnValue>(
         contractId: string,
         method: string,
-        args: Record<string, unknown>,
+        args: Record<string, unknown> | Uint8Array,
         blockQuery: BlockReference = { finality: DEFAULT_FINALITY }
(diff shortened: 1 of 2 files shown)


packages/providers/src/json-rpc-provider.ts
```diff
@@ -226,7 +226,7 @@ export class JsonRpcProvider implements Provider {
     public async callFunction<T extends SerializedReturnValue>(
         contractId: string,
         method: string,
-        args: Record<string, unknown>,
+        args: Record<string, unknown> | Buffer,
```
eclipseeer: Possible issue with `Record<string, unknown> `- it allows you to pass `args = { a: 1n } // bigint` which will fail during JSON.stringify

denbite: good catch, we've made some progress on this before but never fully implemented it
if you've got an experience here, would you be willing to open another PR to fix types?

gagdiez (approved):

PR #1894, opened 2026-03-11
Inline borsh, make is-my-json-valid optional

mikedotexe:
Toward the goal described in #1876 
I've tried to create a couple PRs for ease of review, and this is the first one. It's making some efficiency gains that will be used and leveraged in the two PRs following this. (those have linked to these with decent explanation)

This is the first one that's best to review and consider.

Sidenote, it's interesting that the npm package for borsh is so good it can be serialization for just about anything. What we found here was that it can be much smaller if it's just for the NEAR Protocol side, so that's the idea.

## Summary

- **Inline borsh serializer** — Replace the `borsh` npm dependency with a lean 289-line zero-dep serializer/deserializer tailored to NEAR's schema needs. Eliminates a transitive dependency tree.
- **Make is-my-json-valid optional** — Move `is-my-json-valid` to `optionalDependencies` and dynamically import it with try/catch fallback. Environments without it still work; ABI schema validation is skipped gracefully.
- **Add `.claude/` to `.gitignore`**

## Changes

| File | Change |
|------|--------|
| `src/borsh.ts` | New — 289-line zero-dep borsh serializer/deserializer |
| `src/transactions/schema.ts` | `import from 'borsh'` → `import from '../borsh.js'` |
| `src/nep413/schema.ts` | `import from 'borsh'` → `import from '../borsh.js'` |
| `test/unit/signers/key_pair_signer.test.ts` | `import from 'borsh'` → `import from '../../../src/borsh.js'` |
| `test/unit/transactions/serialize.test.ts` | `import from 'borsh'` → `import from '../../../src/borsh.js'` |
| `src/accounts/errors.ts` | Inline `ValidationError` type (was imported from is-my-json-valid) |
| `src/accounts/typed_contract.ts` | Dynamic `import('is-my-json-valid')` with try/catch; `validateArguments` → async |
| `package.json` | Remove `borsh` from deps; move `is-my-json-valid` to `optionalDependencies` |

## Test plan

- [x] `pnpm compile` (tsc) — no errors
- [x] Borsh round-trip coverage: serialize/deserialize for all NEAR transaction types

### DIFF
--- a/.changeset/inline-borsh-serializer.md
+++ b/.changeset/inline-borsh-serializer.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
+Inline borsh serializer: replace the `borsh` npm dependency with a lean 289-line zero-dep serializer/deserializer tailored to NEAR's schema needs. The `serialize` and `deserialize` functions maintain the same API. This eliminates a transitive dependency tree and reduces install footprint.
--- a/.changeset/optional-json-validator.md
+++ b/.changeset/optional-json-validator.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
+Make `is-my-json-valid` an optional dependency: ABI schema validation via `validateArguments` is now async and dynamically imports `is-my-json-valid` with a try/catch fallback. Environments without it installed will skip validation silently. The `ValidationError` type is inlined in `src/accounts/errors.ts`. Consumers using typed contracts with ABI validation should add `is-my-json-valid` to their own dependencies.
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 #IDE
 .idea
 .vscode
+.claude/
 
 storage
 node_modules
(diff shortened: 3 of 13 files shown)


mikedotexe:
will address these failed CI tests 🫡

mikedotexe:
> will address these failed CI tests 🫡

actually, looks like `master` has CI issues. Thought i'd throw this one-liner PR up, but no worries if it's not the path the team would like to go on: #1897 my thought was just to exclude the `e2e` side or punt it until later so CI is working on current and incoming PRs

gagdiez:
## Respect to borsh.js

How much does it impact removing borsh?

[`borsh`](https://github.com/near/borsh-js/tree/master) is already quite small, dependency-free, and thoroughly tested. By inlining it here we would lose not only the tests, but also create redundancy just to save a couple bytes. I would rather not inline it to be honest.

## Respect to `my-json-valid` optional

We might want to consider how useful it actually is, if we can make it optional then maybe we can just delete it? Users that want to use it could simply validate the input by themselves. Because right now we are "silently skipping it", which to me sounds like unexpected behavior more than a feature. cc @denbite

mikedotexe:
the borsh just includes the blockchain side, not the "borsh can be used for anything" so it makes it a good amount smaller. don't have cycles to keep up, closing

PR #1895, opened 2026-03-11
Add standalone crypto utils and plain transaction builder

mikedotexe:
Addressing #1876 and utilizing the efficiencies of the flat utilities we created in FastNearJS and bringing it in as a new package. The results of compilation sizes is what's driving the reason behind this. So it's effectively a duplicate workspace package, but would be great if I sunset FastNearJS and we just have it here, in the OG repo

## Summary

New public API surface enabling downstream consumers (near-connect wallets) to import crypto + transaction helpers directly from near-api-js instead of @fastnear/utils.

- **Crypto utils** — `sha256`, `curveFromKey`, `keyFromString`, `keyToString`, `publicKeyFromPrivate`, `privateKeyFromRandom`, `signHash`, `signBytes`
- **Plain transaction types** — `PlainTransaction`, `PlainAction`, `PlainAccessKey` + `mapTransaction`, `mapAction`, `mapSignedTransaction` for converting JSON-friendly objects to borsh-serializable format
- **Re-exports** — `parseRpcError` from providers, `actionCreators` alias for backwards compat with @near-js/transactions

**Depends on:** #1894 (inline borsh). Merge #1894 first, then rebase this.

## Changes

| File | Change |
|------|--------|
| `src/crypto/utils.ts` | New — standalone crypto utilities |
| `src/crypto/index.ts` | Re-export new utils |
| `src/transactions/plain.ts` | New — PlainTransaction types + mappers |
| `src/transactions/index.ts` | Add `export * from './plain.js'`; `export type *` → `export *` for actions |
| `src/transactions/action_creators.ts` | Add `actionCreators` alias |
| `src/providers/index.ts` | Add `export { parseRpcError }` |
| `src/index.ts` | Re-export crypto utils + KeyCurve type |

## Test plan

- [x] `pnpm compile` (tsc) — no errors

### DIFF
--- a/.changeset/inline-borsh-serializer.md
+++ b/.changeset/inline-borsh-serializer.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
+Inline borsh serializer: replace the `borsh` npm dependency with a lean 289-line zero-dep serializer/deserializer tailored to NEAR's schema needs. The `serialize` and `deserialize` functions maintain the same API. This eliminates a transitive dependency tree and reduces install footprint.
--- a/.changeset/optional-json-validator.md
+++ b/.changeset/optional-json-validator.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
+Make `is-my-json-valid` an optional dependency: ABI schema validation via `validateArguments` is now async and dynamically imports `is-my-json-valid` with a try/catch fallback. Environments without it installed will skip validation silently. The `ValidationError` type is inlined in `src/accounts/errors.ts`. Consumers using typed contracts with ABI validation should add `is-my-json-valid` to their own dependencies.
--- a/.changeset/plain-transaction-builder.md
+++ b/.changeset/plain-transaction-builder.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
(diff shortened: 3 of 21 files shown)


gagdiez:
## Crypto Utils

I see this utils mostly export functionality that already exists in `KeyPair`, is there any reason to have them all as public utils instead of asking people to just handle their keys as `KeyPair`s?

Because right now you can do:
- `const key = KeyPair.fromRandom()`
- `const key = KeyPair.fromString(privateKey)`
- `key.toString()`
- `key.getPublicKey()`

## Plain Transactions
I'm not super sure what the `plain transactions` help to achieve, in `v7` we actually hide the internal actions representations because people was always getting confused on how to use them and thus using them wrong.

That's why now all actions are hidden behind the `actions` creator (known before as the `actionsCreator` which this PR correctly brings back and marks as deprecated).

mikedotexe:
> is there any reason to have them all as public utils instead of asking people to just handle their keys as `KeyPair`s

moving away from object-oriented programming into pure functions, since the core issue with the old near-api-js was that class instances can come from dependencies that use near-api-js where the types will mismatch

gagdiez:
indeed, that was mainly caused by two things:

1. using `isinstanceof` in the codebase, which I think we remove completely, so that problem should not be happening anymore

2. mixture of "@near-js" packages, which now should be no longer happening as `near-api-js` is a single library again

but I might be wrong here, are you still finding this error somewhere? if yes, then we can implement the standalone crypto, or maybe just migrate PublicKey to be an interface instead of a class

mikedotexe:
yeah there are still errors, I asked AI to come up with one and it did in 15 minutes

PR #1896, opened 2026-03-11
Switch build from tsc to tsup (ESM + CJS + IIFE)

mikedotexe:
the big kahuna. closes #1876 
first, I know it's kind of rude or perhaps a bigger deal to suggestion replacing `tsc` with `tsup`, but it came after long and grueling hours. Originally with FastNearJS I had a combo of `tsc` using `no-emit` and `esbuild`, but found `tsup` is a superset of `esbuild` and just did everything in what felt like a bulletproof way.

What's not as obvious from the summary below, is that the IIFE bundle means `near-api-js` can and will work on a static HTML site. No `yarn start` it can just use a `<script>` tag and be pulled from a CDN if desired. Quite portable while still allowing for regular NodeJS apps to use the EcmaScript build.

This change allows these three target ouputs EcmaScript, CommonJS, and browser bundle IIFE to come from the same code, and leverage ESM's safety/efficiencies. So someone can build a React app like normal and another person can (literally) run `near-api-js` on a tiny IoT sensor talking to the blockchain.

I say this not to sway, but to celebrate as we look into this: we'll have the most efficient web3 library in the world, I'm willing to debate. the most efficient web3 lib for the most efficient blockchain in the world 💪🏻

## Summary

Replace the tsc-only build with tsup producing three output formats, matching the FastNear JS Monorepo build pattern:

- **ESM** (unbundled) — tree-shakeable, preserves module structure
- **CJS** (unbundled) — Node.js CommonJS with `.cjs` extensions
- **IIFE** (bundled) — 252KB browser bundle for `<script>` tags, with lightweight shims

Adds browser shims to keep the IIFE bundle small (~116KB saved):
- `error-parse-lite.ts` — lightweight error parser replacing ~62KB error class hierarchy
- `json-validator-noop.ts` — no-op ABI validator replacing ~44KB is-my-json-valid
- `util.ts` — browser shim for node:util

**Depends on:** #1894 (inline borsh) and #1895 (crypto utils). Merge those first, then rebase this.

## Changes

| File | Change |
|------|--------|
| `tsup.config.ts` | New — build config for ESM, CJS, IIFE |
| `scripts/fix-cjs-imports.js` | New — post-build CJS .js→.cjs path fixer |
| `src/shims/error-parse-lite.ts` | New — lightweight IIFE error parser |
| `src/shims/json-validator-noop.ts` | New — no-op IIFE validator shim |
| `src/shims/util.ts` | New — browser shim for node:util |
| `package.json` | `exports` (lib/→dist/), `main`/`module`/`browser`/`types`, `build` script (tsup), `tsup` devDep, `files` (dist), `keywords`, updated `clean`/`autoclave`/`check-exports` |

## Build output

| Format | Size | Notes |
|--------|------|-------|
| ESM | ~200KB | Unbundled, tree-shakeable |
| CJS | ~300KB | Unbundled, .cjs extensions |
| IIFE | 252KB | Single bundled file, browser-ready |

## Test plan

- [x] `pnpm build` (tsup) — no errors, all 3 formats produced
- [x] `pnpm check-exports` (attw) — all 7 subpath exports green across node10/node16/bundler

### DIFF
--- a/.changeset/iife-browser-shims.md
+++ b/.changeset/iife-browser-shims.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
+Add lightweight browser shims for the IIFE bundle: `error-parse-lite.ts` replaces the ~62KB error class hierarchy with a generic `RpcError` parser (preserving `InvalidNonceError` for nonce retry logic), `json-validator-noop.ts` stubs out the ~44KB `is-my-json-valid` dependency, and `util.ts` provides a minimal `util.format` shim. These save ~116KB in the browser bundle while maintaining full functionality in ESM/CJS builds.
--- a/.changeset/inline-borsh-serializer.md
+++ b/.changeset/inline-borsh-serializer.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
+Inline borsh serializer: replace the `borsh` npm dependency with a lean 289-line zero-dep serializer/deserializer tailored to NEAR's schema needs. The `serialize` and `deserialize` functions maintain the same API. This eliminates a transitive dependency tree and reduces install footprint.
--- a/.changeset/optional-json-validator.md
+++ b/.changeset/optional-json-validator.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": minor
+---
+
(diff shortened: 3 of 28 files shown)


gagdiez:
I like the idea of bringing ESM + CJS back, we used to have it and some users reported that they need it back, would be nice if we could have a PR that only does that (i.e. it only changes `tsc` for `tscup` - though some community members suggested that maybe we should look for `tsdown` which is the one being maintained)

mikedotexe:
the good part about CJS is that an agent can `curl` the javascript library and immediately use it without `pnpm install` and stuff

PR #1897, opened 2026-03-11
fix: exclude e2e from root tsconfig typecheck

mikedotexe:
It seems current state is having CI issues on the end-to-end testing code. So this just gets rid of the inclusion of `e2e` which could also be seen as a total hack just to get CI to work (happy to close) or a way to exclude this until there's a better time to address the e2e stuff and re-add it. I'm loosely held on this, just thought I'd pusn a quickie since I saw CI issues on my recent tranche of 3 PRs

## Summary
- Remove `e2e` from the root `tsconfig.json` `include` array
- The `e2e/` directory has its own `tsconfig.json` and its own dependency install step in CI (`pnpm it --ignore-workspace --dir e2e`), so including it in the root typecheck causes failures when e2e deps aren't installed
- This makes `pnpm typecheck` work locally without needing to run e2e install first

## Test plan
- [ ] `pnpm typecheck` passes locally without running e2e install
- [ ] CI pipeline passes (e2e tests still run via their own step)

### DIFF
--- a/.changeset/exclude-e2e-from-typecheck.md
+++ b/.changeset/exclude-e2e-from-typecheck.md
@@ -1,0 +1,5 @@
+---
+"near-api-js": patch
+---
+
+Exclude `e2e` directory from root `tsconfig.json` typecheck. The e2e directory has its own tsconfig and its own dependency install step, so including it in the root typecheck caused spurious failures when e2e deps weren't installed.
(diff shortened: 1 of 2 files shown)


mikedotexe:
i'll keep this pull request open because when I made it, it fixed existing problems with CI, but I can't keep up, just a simple fix

gagdiez (dismissed):

gagdiez (approved):

